PR #6, opened 2020-08-14
Added gCurrentScriptList and some more matching funcs

JoshW-7:


nanaian:
Probably worth giving gCurrentScriptList the type `script_context* [128]` and using the ARRAY_LENGTH macro when iterating over it.

JoshW-7:
I modified the code to reflect the suggestion above. Instead of using ARRAY_COUNT though, I'm using a global #define of MAX_SCRIPT_LIST_LENGTH 128, since there are two lists of script_context* that it can point to.

PR #7, opened 2020-08-14
match load_map_bg

nanaian:
Also added three new symbols to `undefined_syms.h`. Two of these are almost certainly string literals in `load_map_bg`, but - as discussed - we don't have rodata capability yet so they have to be declared in a different way.


include/variables.h
```diff
@@ -15,6 +15,9 @@ extern s16* D_80151328;
 extern s16 D_8010CD10;
 extern s16 D_8010CD12;
 extern s32 D_801595A0;
+extern u8 gCloudyFlowerFieldsBg; // "fla_bg"
```
ethteck, line 18: Does it work to make these `u8*`? plain `u8`s is kind of odd.

src/code_dc470_len_14c0.c
```diff
@@ -1,25 +1,26 @@
...
-            if (strcmp(tempvar0, 0x8014F120) == 0) {
-                tempvar0 = flowerfields_bg1;
-            }
-            load_asset_by_name (tempvar0, 0);
+void load_map_bg(u8* optAssetName) {
+    UNK_PTR compressedData;
+    u32 assetSize;
+    u8* assetName;
+
+    if (!optAssetName) return;
```
ethteck, line 8: I think `if (optAssetName == NULL)` is better here

include/variables.h
nanaian: `u8*` causes type warnings to appear when giving `strcmp(_, &gCloudyFlowerFieldsBg)` since it means we're actually giving it `u8**`.

nanaian: Dropping the `&` and using `u8*` compiles, but doesn't match. Using `&` or a string literal changes up the instruction order somewhat.

pixel-stuck: imo, the declaration for this should be: ``extern char gCloudyFlowerFieldsBg[];`` If that works, then I'd prefer that.

nanaian: `char` over `u8`?

ethteck: this won't be a thing soon anyway, since it'll be a string literal, so I wouldn't worry too much

nanaian: Updated the code to use `char*` for string pointers and `char[]` for those externs as @pixel-stuck suggested - it matches.

RE: string literal, it caused the loading of the string address to happen earlier, which matches. Without a literal or `&` a few instructions were swapped. With the `char[]` typing I can omit `&` and it continues to work fine, however :sparkles:

PR #8, opened 2020-08-14
update game_status struct

nanaian:
Updated game_status to match the Star Rod 0.3.3 release candidate I have access to's database  
~~Non-matching right now, I probably made an alignment/sizing error somewhere - proofreading would be greatly appreciated~~ edit: good to merge :shipit:

### DIFF
--- a/include/common_structs.h
+++ b/include/common_structs.h
@@ -7,15 +7,15 @@
 typedef struct {
     /* 0x000 */ u32 currentButtons;
     /* 0x004 */ u32 altCurrentButtons; /* used for battle when flag 80000 set */
-    /* 0x008 */ char unk_08[4];
+    /* 0x008 */ char unk_08[8];
     /* 0x010 */ u32 pressedButtons; /* 1 for the frame button begins to be held */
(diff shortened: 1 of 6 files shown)


include/common_structs.h
```diff
@@ -5,62 +5,92 @@
...
     /* 0x058 */ char unk_58[0x10];
     /* 0x068 */ u16 demoButtonInput;
     /* 0x06A */ s8 demoStickX;
     /* 0x06B */ s8 demoStickY;
-    /* 0x06C */ char unk_6C[0x4];
-    /* 0x070 */ s8 unk_70;
-    /* 0x071 */ s8 demoState;
-    /* 0x072 */ s8 unk_72;
+    /* 0x06C */ u32* mainScriptID;
+    /* 0x070 */ u8 isBattle;
```
pixel-stuck, line 32: most likely the source of the nonmatching is something like one of these which you've changed from signed to unsigned. Be careful with this, as this can be the difference between, for example, ``lb`` and ``lbu``

nanaian: Yep, it was a combination of signedness differences are misalignment. It matches now :tada:

PR #9, opened 2020-08-14
Work on script interpreter

ethteck:


src/code_e92d0_len_5da0.c
```diff
@@ -120,79 +235,173 @@ INCLUDE_ASM(code_e92d0_len_5da0, si_handle_get_4_float);
...
+s32 func_802C73B0(script_context* script) {
+    return 2;
+}
 
-INCLUDE_ASM(code_e92d0_len_5da0, func_802C73B0);
+s32 func_802C73B8(script_context* script) {
+    s32 i;
 
-INCLUDE_ASM(code_e92d0_len_5da0, func_802C73B8);
+    for (i = 0; i < 0x80; i++) {
```
pixel-stuck, line 398: do we know what this is the size of? If not, all good to leave as-is, though maybe a todo should be left.

nanaian: `ARRAY_COUNT(gCurrentScriptListPtr)`, given that each iteration accesses `(*gCurrentScriptListPtr)[index]`

PR #15, opened 2020-08-15
near-match si_handle_call

nanaian:
Note: based on the types in the asm for si_handle_call, I think the `initialCall` bool passed to script api functions, which is currently typed s32, is actually a u8. See `callFunction`'s type.  
It might be the case that the return type for api functions should also be u8.


ethteck:
Interesting discovery, but I'd rather just fully match this before we merge it. I can give it a shot soon

ethteck:
@nanaian heads up that this currently is failing on Jenkins. Not sure why exactly but yeah

nanaian:
I think it's failing because of [this type change](https://github.com/ethteck/papermario/pull/15/commits/d6986180a7f6e3b001b1ece6963b3b3af66dc039#diff-7501190fb5fb5a0e0491b96eed69b5abL181), but si_handle_call it definitely seems to be a `u8`...

ethteck:
closed since we can't match this, but yee

PR #16, opened 2020-08-15
match UpdateLerp and clean up MakeLerp

nanaian:


src/code_f8f60_len_1560.c
```diff
@@ -2,23 +2,34 @@
...
+    script->varTable[0xE] = 0; // elapsed
+
     return 2;
 }
 
-INCLUDE_ASM(code_f8f60_len_1560, UpdateLerp);
+s32 UpdateLerp(script_context* script, s32 initialCall) {
+    script->varTable[0x0] = (s32) update_lerp(
+        script->varTable[0xB],
+        (s32) script->varTable[0xC],
```
ethteck, line 18: I'm not sure these casts are necessary since these variables are already s32s

nanaian: They're necessary! `update_lerp` takes `f32` in these two slots, and without the cast the compiler attempts to convert the s32 ints to float rather than just moving them to f registers (not converting).

nanaian: wait no

PR #18, opened 2020-08-15
split and match most functions at f8f60

nanaian:
Split and matched all functions in this file, other than:
- Path-related functions - `setup_path_data`, for example, is _huge_!
- `func_802D4BDC`, `func_802D4C4C` - has a redundant branch instruction that probably skipped a debug printf at one point. I don't know how to replicate this - the compiler optimises away the empty if block.
- `func_802D4CC4` - loads an f32 from `0xBF800000` - which is a reference to some DMA'd ROM data, I think.


include/si_vars.h
```diff
@@ -0,0 +1,8 @@
+#ifndef _SI_VARS_H_
+#define _SI_VARS_H
+
+#define SI_VAR_0 0xFE363C80
+
+// TODO
+
+#endif
```
ethteck, line 8: Newline pls

src/code_f8f60_len_1560.c
```diff
@@ -55,18 +260,103 @@ INCLUDE_ASM(code_f8f60_len_1560, LoadPath);
...
+    return 2;
+}
+
+s32 ClampAngleFloat(script_context* script, s32 initialCall) {
+    bytecode* ptrReadPos = script->ptrReadPos;
+
+    set_float_variable(script, *ptrReadPos, clamp_angle(get_float_variable(script, *ptrReadPos)));
+
+    return 2;
+}
```
ethteck, line 362: newlin

include/si_vars.h
ethteck: I'm not sure how I feel about this file. What do you imagine going here?

nanaian: I expect we'll add constants for each of the special values that `get_variable` and friends dereference. i.e. `SI_VAR_0` is equivalent to Star Rod's `*Var[0]`.

> we maybe should just shorten it to SI_ instead of SI_STATUS_
:+1: to that. They could go in this file. Might be worth renaming SI_VAR_ (to maybe just VAR_?) since the vars aren't part of that SI return status "enum".

include/common_structs.h
```diff
@@ -4,6 +4,12 @@
 #include "ultra64.h"
 #include "types.h"
 
+typedef struct vec3f {
+    /* 0x00 */ f32 x;
+    /* 0x04 */ f32 y;
+    /* 0x08 */ f32 z;
+} vec3f; // size = 0x0C
```
pixel-stuck, line 11: I'd personally prefer Vec3f; will see how Ethan feels about that.

include/common_structs.h
```diff
@@ -18,7 +24,7 @@ typedef struct cam_pos_settings {
     /* 0x04 */ f32 boomLength;
     /* 0x08 */ f32 boomPitch;
     /* 0x0C */ f32 viewPitch;
-    /* 0x10 */ f32 position[3];
+    /* 0x10 */ struct vec3f position;
```
pixel-stuck, line 27: if we do end up using Vec3f, it needs to be changed here. Also, you can leave the struct off, and just use Vec3f.

include/variables.h
```diff
@@ -50,4 +50,6 @@ extern u16 D_800A0940;
 extern u16 D_800A0942;
 extern s16 D_800A0944;
 
+extern f32 D_BF800000;
```
pixel-stuck, line 53: as mentioned on discord, 0xBF800000 isn't an address, it's just the raw hex representation of -1.0f

src/code_f8f60_len_1560.c
```diff
@@ -31,21 +31,226 @@ s32 UpdateLerp(script_context* script, s32 initialCall) {
...
+        return 0;
+    }
+}
+
+s32 IsPlayerWithin(script_context* script, s32 initialCall) {
+    bytecode* ptrReadPos = script->ptrReadPos;
+    player_status* playerStatus = &gPlayerStatus;
 
-INCLUDE_ASM(code_f8f60_len_1560, AwaitPlayerLeave);
+    /* Function temporaries are redundant, stack should be used instead */
```
pixel-stuck, line 117: Not sure all of these redundant comments are really needed

include/common_structs.h
ethteck: The capital version is what oot and co use, but we have a tendency in this project to use lowercase type names already (which I'm not a huge fan of, tbh, but I think that's a bigger conversation). I'd say we should use the lowercase version and then have another conversation about uppercase types later

src/code_f8f60_len_1560.c
ethteck: Yeah, I was iffy on that as well..

include/common_structs.h
nanaian: > which I'm not a huge fan of, tbh, but I think that's a bigger conversation
New issue incoming? I'm not a big fan of it either - Star Rod is even internally inconsistent.

For now, yeah, let's stay with lowercase.

include/common_structs.h
nanaian: I use `struct` everywhere in this file because it's what the `starrod_structs_to_h.py` generates, but we can (should?) change that.

include/functions.h
```diff
@@ -9,24 +9,37 @@ void osCleanupThread(void);
...
 void func_80137E10(s32, u8, u8, u8);
+void func_80137E4C(s32, s32, s32, s32);
 s32 rand_int(s32);
 void sort_items(void);
 s32 is_ability_active(s32 arg0);
 f32 update_lerp(EASING easing, f32 start, f32 end, s32 elapsed, s32 duration);
 
+npc* get_npc_safe(NPC npcID);
+npc* get_npc_unsafe(NPC npcID);
+npc* resolve_npc(script_context* script, NPC npcID);
```
nanaian, line 37: It might be worth declaring some kind of union for `resolve_npc`'s second arg - e.g. `union { npc* ptr; NPC special; u8 id }`. AFAIK the `get_npc` don't check for `NPC special` and should just take a `u8` id. We might even want to rename them - and suggest the same to Clover - to `get_npc_by_id`.

nanaian: Update: changing the `get_npc_` signatures to take a `u8` rather than an `s32` doesn't match.

include/si_vars.h
ethteck: I think this should just be called si.h then, and yeah that makes sense I think, but I'm also not sure what these vars actually represent

nanaian: > I'm also not sure what these vars actually represent

`get_variable` and related functions resolve `bytecode`s. Take the following script, for example:
```py
Set *Var[0] 0             # SET 00000002 SI_VAR_0 0
Set *Var[0] *Var[0]       # SET 00000002 SI_VAR_0 SI_VAR_0
```
Notice that the second operand to `Set` is a union between an immediate int and some special constant values (these are very very low, negative values and are never needed as immediates).
So in this case `get_variable(script, 0)` will return `0` whilst `get_variable(script, SI_VAR_0)` will return whatever the value of `script->varTable[0]` is.

nanaian: From Clover's docs:
Screenshot_20200815_213514

include/common_structs.h
nanaian: I tried changing to avoid `struct` where possible, but it forces us to declare types before they are used meaning reordering everything in the file with zero-dependency structs at the top. I reckon we should leave `struct` there until we split `common_structs.h` into a bunch of header files closer to the c files which use them.

asm/nonmatchings/code_f8f60_len_1560/func_802D4BDC.s
```diff
@@ -0,0 +1,38 @@
...
+/* 0F95D8 802D4C28 C4600074 */  lwc1  $f0, 0x74($v1)
+/* 0F95DC 802D4C2C 46800020 */  cvt.s.w $f0, $f0
+/* 0F95E0 802D4C30 44050000 */  mfc1  $a1, $f0
+/* 0F95E4 802D4C34 0C04DF69 */  jal   func_80137DA4
+/* 0F95E8 802D4C38 2404000A */   addiu $a0, $zero, 0xa
+/* 0F95EC 802D4C3C 0000102D */  daddu $v0, $zero, $zero
+.L802D4C40:
+/* 0F95F0 802D4C40 8FBF0010 */  lw    $ra, 0x10($sp)
+/* 0F95F4 802D4C44 03E00008 */  jr    $ra
+/* 0F95F8 802D4C48 27BD0018 */   addiu $sp, $sp, 0x18
```
ethteck, line 38: I hate to sound annoying about the newline stuff, but it actually matters specifically in the .s case because the assembler emits warnings for files without newlines.

include/types.h
```diff
@@ -8,6 +8,6 @@
 #define UNK_FUN_PTR(name) void(*name)(void)
 #define UNK_ARGS
 
-typedef s32* bytecode;
+typedef s32 bytecode;
 
 #endif
```
ethteck, line 13: I don't think it's your fault, but while you're at it could you add a newline here?

nanaian: Will resolve these newline issues, I like 'em at the end of files too. This hopefully won't be an issue for most in the future thanks to #19 :package:

PR #19, opened 2020-08-15
add editor dotfiles

nanaian:
Added dotfiles for vscode and a generic `.editorconfig` for other editors that support it. AFAIK including complex editor-specific (but _not_ user-specific) configuration like this is pretty normal for repos.

On vscode, you can use ctrl+shift+b (run default build task) to compile the game, which reports errors and warnings from the compiler in-editor (see image). Additionally I made a test task for running `diff.py` - it prompts you for a function name before running.

image

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+[*]
+indent_style = space
+indent_size = 4
+end_of_line = lf
+charset = utf-8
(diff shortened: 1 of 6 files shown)


ethteck:
@pixel-stuck can I go ahead and merge? I forgot if you use vscode

PR #23, opened 2020-08-17
add ASSERT macro

nanaian:
Also matched si_handle_switch_const and si_handle_switch, which use ASSERT.


src/code_e92d0_len_5da0.c
```diff
@@ -94,20 +92,27 @@ s32 si_handle_end_if(script_context* script) {
...
+
+    script->switchBlockValue[switchDepth] = value;
+    script->switchBlockState[switchDepth] = 1;
     return 2;
-}*/
+}
+
+s32 si_handle_switch_const(script_context* script) {
+    bytecode value = *script->ptrReadPos;
+    s32 switchDepth = ++script->switchDepth;
```
ethteck, line 108: can we do post-increment unless otherwise necessary?

src/code_e92d0_len_5da0.c
```diff
@@ -134,24 +139,19 @@ INCLUDE_ASM(code_e92d0_len_5da0, si_handle_case_equal_AND);
...
     s32 switchDepth = script->switchDepth;
-    
-    if (switchDepth < 0) {
-        inf_loop: goto inf_loop; // todo macro? how to do without label
-    }
 
-    script->switchBlockState[script->switchDepth] = 0;
+    ASSERT(switchDepth >= 0);
+
+    script->switchBlockState[switchDepth] = 0; // here
```
ethteck, line 152: `// here` can be removed. thanks for matching this with the macro!

nanaian: Oh, whoops, will remove this.

src/code_e92d0_len_5da0.c
nanaian: It's necessary! Alternatively we could do
```c
script->switchDepth++;
s32 switchDepth = switchDepth;
```
but I don't like that as much.

ethteck: Oh I see why that's different - yeah. Okay, carry on :)

PR #24, opened 2020-08-17
Fix GitHub's language detection

nanaian:
GitHub current thinks some of the header files in this repo are written in Objective-C or C++:
image

This minor PR [tells github/linguist that *.h files are written in C](https://github.com/github/linguist#overrides).

(FTR, [zeldaret/oot](https://github.com/zeldaret/oot) has the same issue - it might be worth doing something similar there too. It even has a `.gitattributes` file already!)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.h linguist-language=C


nanaian:
It also might be worth excluding `tools/*` from the language stats, but I haven't done that here.

PR #25, opened 2020-08-17
update readme and add install script

nanaian:
Install script has been tested under Arch. I'd appreciate someone on Ubuntu testing the script there, although it should work fine.


README.md
```diff
@@ -0,0 +1,110 @@
...
+
+Open up the relevant `.c` file and replace the function's `INCLUDE_ASM` macro with the output from mips_to_c. Run the following command to attempt to compile, replacing `function_name` with the name of the function you're working with:
+
+```sh
+./diff.py -mwo function_name
+```
+
+Fix any errors and rerun `diff.py`. This will involve typing the function signature correctly, which you will probably find in [Star Rod's library database](https://github.com/nanaian/star-rod/blob/master/database/common_func_library.lib). See also [common_structs.h](include/common_structs.h).
+
+Once a successful build is made, `difflpy` will show you the difference between the original game's assembly (on the left) and what your C code generated (on the right).
```
ethteck, line 89: diffl

PR #26, opened 2020-08-17
add astyle, clang-tidy

nanaian:
- Fixed the `format.sh` inherited from the SM64 decomp.
- Updated the checks clang-tidy uses to those which probably won't produce non-matching functions.
- Disabled clang-tidy `--fix`.
- Added `clang-tidy`'s Visual Studio Code extension to suggested extensions
- Added [a GitHub action workflow](https://github.com/nanaian/papermario/runs/995983424?check_suite_focus=true) which runs on push and on pull requests which modify `*.c` or `*.h`, which fails if `format.sh` changed any files or `clang-tidy` emits any errors. Where possible, it'll only format/check files changed in that commit or PR.
- Updated existing code to not emit clang-tidy errors.


include/common_structs.h
```diff
@@ -542,7 +543,8 @@ typedef struct BattleStatus {
     /* 0x17E */ s16 currentAttackDamage;
     /* 0x180 */ s16 lastAttackDamage;
     /* 0x182 */ char unk_182[2];
-    /* 0x184 */ s32 currentTargetListFlags; /* set when creating a target list, also obtain from the flags field of moves */
+    /* 0x184 */ s32
+        currentTargetListFlags; /* set when creating a target list, also obtain from the flags field of moves */
```
nanaian, line 547: This is awkward, we could set the clang-format line length limit to `999`, or manually place the documentation comment above.

include/ultra64.h
```diff
@@ -1,40 +1,40 @@
...
+#include <PR/mbi.h>
+#include <PR/libaudio.h>
+#include <PR/gu.h>
+#include <PR/ramrom.h>
+#include <PR/sp.h>
+#include <PR/ucode.h>
+#include <PR/ultraerror.h>
+#include <PR/ultralog.h>
+
+#endif
```
nanaian, line 40: A line-ending change. `include/PR/*.h` are all CRLF also, but formatting/linting doesn't run on them.

src/code_f8f60_len_1560.c
```diff
@@ -263,12 +253,14 @@ ApiStatus GetDist2D(ScriptInstance* script, s32 isInitialCall) {
...
-        get_float_variable(script, *ptrReadPos++)
-    ));
+    set_float_variable(
+        script,
+        outVar,
+        dist2D(
+            get_float_variable(script, *ptrReadPos++),
+            get_float_variable(script, *ptrReadPos++),
+            get_float_variable(script, *ptrReadPos++),
+            get_float_variable(script, *ptrReadPos++)));
```
nanaian, line 263: Maybe worth extracting the distance to a variable above to make this more readable?

nanaian: Ideally clang-format would format multiline function calls like the following, but I'm not sure how to get this behaviour.
```diff
  foo(
    arg0,
    arg1,
-   arg2)
+   arg2
+ )
```

include/common_structs.h
ethteck: Yeah this isn't ideal. We disable the formatting of headers in oot for this kind of reason.

include/ultra64.h
ethteck: Ah yeah, makes sense.

src/code_f8f60_len_1560.c
ethteck: Yeah...I think this is definitely a bummer but not a huge deal

src/os/code_3d330_len_990.c
```diff
@@ -1,13 +1,44 @@
 #include "common.h"
 
-void INCLUDE_ASM(os/code_3d330_len_990, guFrustumF, float mf[4][4], float l, float r, float b, float t, float n,
-                 float f, float scale);
+void INCLUDE_ASM(
```
ethteck, line 3: These are yucky

include/common_structs.h
```diff
@@ -247,12 +252,12 @@ typedef struct ScriptInstance {
     /* 0x0F0 */ s32 loopCounterTable[8];
     /* 0x110 */ u8 switchBlockState[8];
     /* 0x118 */ s32 switchBlockValue[8];
-    /* 0x138 */ s32* buffer;
-    /* 0x13C */ UNK_PTR array;
-    /* 0x140 */ UNK_PTR flagArray;
+    /* 0x138 */ ScriptBufferValue* buffer;
```
ethteck, line 255: I'm not sure I really like naming a struct Thing**Value** but I don't have a better name atm..hm

nanaian: I wonder if I could declare _buffer_ as a union between a pointer to an int or a pointer to a float, rather than the type of what it points to. That is:
```diff
- buffer[0].s
+ buffer.s[0]
```

src/os/code_3d330_len_990.c
nanaian: Personally I think they're.....fine. We could change the function param breaking style to something more compact if you'd prefer? Or selectively disable clang-format for this line.

include/common_structs.h
nanaian: If it helps, Star Rod uses the types `IntBuffer` and `FloatBuffer` for the structs you pass to the `SetBuffer` and `SetFloatBuffer` script commands.

ethteck: okay, yeah I can't think of a better name so moop

.clang-format
```diff
@@ -2,9 +2,8 @@ IndentWidth: 4
 ContinuationIndentWidth: 4
 Language: Cpp
 UseTab: Never
-ColumnLimit: 120
+ColumnLimit: 0
```
ethteck, line 5: Can't we get away with a longer limit like 160 or something? I do like having not infinite line length.

ethteck: After the discussion with @pixel-stuck in discord, I think we should try something like 160 and see if we can get away with not having to put struct comments on their own line. Also we're both in agreement about not spacing struct fields apart and having one line per field with no gaps.

nanaian:
Should be good now :package:

PR #27, opened 2020-08-18
remove matched asm, add NON_MATCHING

nanaian:


tools/coverage.py
```diff
@@ -0,0 +1,49 @@
...
+for filename in C_FILES:
+    with open(filename, "r") as file:
+        matched += funcs_in_c(file.read())
+
+non_matched = [path.splitext(path.basename(filename))[0] for filename in ASM_FILES]
+
+total = len(matched) + len(non_matched)
+print(f"{len(matched)}/{total} ({(len(matched) / total) * 100:.2f}%)")
+
+matched_but_undeleted_asm = [f for f in matched if f in non_matched]
```
ethteck, line 43: I like that you added a mechanism for removing unused asm files, but this seems like a kind of random place to do it. Let's abstract it into another file so we can call it directly and from here, maybe?

pixel-stuck: I agree here; it should be its own script we can run, as well as use with jenkins, to make sure unused asm is deleted/figure out if unused asm is present

nanaian: > as well as use with jenkins

Believe it was discussed that we'd add a GitHub action on pull request to check for undeleted asm

PR #29, opened 2020-08-18
match commented functions which use floats

nanaian:


src/code_e79b0_len_1920.c
```diff
@@ -19,25 +19,24 @@ INCLUDE_ASM(code_e79b0_len_1920, start_child_script);
...
 
 INCLUDE_ASM(code_e79b0_len_1920, restart_script);
-// TODO: Find out why things break when script->timeScale = 1 goes after the previous lines
 /*
 ScriptInstance* restart_script(ScriptInstance* script) {
     script->loopDepth = -1;
     script->switchDepth = -1;
     script->currentOpcode = 0;
-    script->frameCounter = 0;
+    script->frameCounter = 0; // TODO: force compiler to not optimise away this
```
ethteck, line 27: What's the issue here?

nanaian: In the asm the same instruction is repeated twice for seemingly no reason.
```
sw zero, frameCounter(script)
sw zero, frameCounter(script)
```

PR #32, opened 2020-08-19
More NPC functions

JoshW-7:


src/code_e20eb0.c
```diff
@@ -1,6 +1,20 @@
 #include "common.h"
 
-INCLUDE_ASM(code_e20eb0, func_802B7000);
+// TODO: Figure out what this function is doing
+s32 func_802B7000(f32 angle) {
+    s32 angleDiff;
+
+    if (270.0f <= angle) {
```
ethteck, line 7: can the `angle` be on the left side of these conditionals? I hope so...

src/code_f2470_len_27f0.c
```diff
@@ -191,26 +191,150 @@ INCLUDE_API_ASM(code_f2470_len_27f0, SetNpcFlagBits);
...
+    return ApiStatus_DONE2;
+}
+
+// TODO: Figure out what this function does
+ApiStatus func_802CF54C(ScriptInstance* script, s32 isInitialCall) {
+    func_800EF43C();
+    return ApiStatus_DONE2;
+}
+
+// TODO: Figure out what this function does
```
ethteck, line 283: I think unnamed / undocumented functions are clear in that they're not documented or named, so this comment isn't necessary

src/code_f2470_len_27f0.c
```diff
@@ -191,26 +191,150 @@ INCLUDE_API_ASM(code_f2470_len_27f0, SetNpcFlagBits);
...
+ApiStatus func_802CF56C(ScriptInstance* script, s32 isInitialCall) {
+    Bytecode* ptrReadPos = script->ptrReadPos;
+    s32 value = get_variable(script, *ptrReadPos++); 
+
+    if(value == 2) {
```
ethteck, line 288: `if (`

src/code_f2470_len_27f0.c
```diff
@@ -191,26 +191,150 @@ INCLUDE_API_ASM(code_f2470_len_27f0, SetNpcFlagBits);
 
 INCLUDE_API_ASM(code_f2470_len_27f0, GetNpcPos);
 
-INCLUDE_API_ASM(code_f2470_len_27f0, EnableNpcShadow);
+ApiStatus EnableNpcShadow(ScriptInstance* script, s32 isInitialCall) {
+    Bytecode* ptrReadPos = script->ptrReadPos;
+    NpcId npcID = get_variable(script, *ptrReadPos++);
+    s32 enableShadow = get_variable(script, *ptrReadPos++);
+    Npc* npcPtr = resolve_npc(script, npcID);
```
ethteck, line 198: I don't think we should name temps thingPtr generally, this and all below can just be `npc`

src/code_f2470_len_27f0.c
```diff
@@ -191,26 +191,150 @@ INCLUDE_API_ASM(code_f2470_len_27f0, SetNpcFlagBits);
...
+    NpcId npcID = get_variable(script, *ptrReadPos++);
+    s32 enableBlur = get_variable(script, *ptrReadPos++);
+    Npc* npcPtr = resolve_npc(script, npcID);
+
+    if (npcPtr != NULL) {
+        if (enableBlur) {
+            enable_npc_blur(npcPtr);
+            return ApiStatus_DONE2;
+        }
+        else {
```
ethteck, line 223: The control flow of these are weird...but if we need the early return, we don't need an else probably

src/code_e20eb0.c
JoshW-7: I tried and it cannot

src/code_f2470_len_27f0.c
JoshW-7: I like having TODO because it allows for easy searching of things that need to be done.

src/code_f2470_len_27f0.c
JoshW-7: Yeah, the returns are required. I removed the `else`.

src/code_e20eb0.c
pixel-stuck: it can be; ``angle >= 270.0f``in my matching copy locally. Same pattern for the rest

src/code_f2470_len_27f0.c
pixel-stuck: there will be multiple eventual documentation passes over all files in an attempt to name things once the codebase is done/close to done/as we discover new things. TODOs aren't really necessary imo.

src/code_e20eb0.c
JoshW-7: Ah, I was doing `angle > 270.f` like a dummy. Thank you. Sending up the revised version...

src/code_ef070_len_3400.c
```diff
@@ -1,6 +1,44 @@
 #include "common.h"
 
 INCLUDE_API_ASM(code_ef070_len_3400, SetCamEnabled);
+/*
```
pixel-stuck: can you leave a single line comment above this specifying what's wrong with the commented func?

src/code_f2470_len_27f0.c
```diff
@@ -191,26 +191,145 @@ INCLUDE_API_ASM(code_f2470_len_27f0, SetNpcFlagBits);
...
+    NpcId npcID = get_variable(script, *ptrReadPos++);
+    s32 enableShadow = get_variable(script, *ptrReadPos++);
+    Npc* npc = resolve_npc(script, npcID);
+
+    if (npc != NULL) {
+        if (enableShadow) {
+            enable_npc_shadow(npc);
+            return ApiStatus_DONE2;
+        }
+        else {
```
pixel-stuck, line 205: styling here, and maybe removal of the else entirely (as below with EnableNpcBlur)

src/code_ef070_len_3400.c
JoshW-7: I just went ahead and removed this one as it's in WIP by me.

PR #34, opened 2020-08-20
add dependencies for n64splat

nanaian:
Might also be good to add a `requirements.txt` to n64splat's repo and install that.
The change from `apt install` to `apt-get install` should support WSL's Ubuntu better.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,14 +1,15 @@
 #!/bin/bash
 
 # Ubuntu
-if command -v apt-install &> /dev/null; then
+if command -v apt-get &> /dev/null; then
     echo "Installing packages for Ubuntu"
 
-    sudo apt install -y git build-essential binutils-mips-linux-gnu zlib1g-dev libcapstone-dev libyaml-dev gcc-multilib || exit 1
+    sudo apt-get install -y git python3 python3-pip build-essential binutils-mips-linux-gnu zlib1g-dev libyaml-dev gcc-multilib || exit 1
+    python3 -m pip install capstone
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -1,14 +1,15 @@
 #!/bin/bash
 
 # Ubuntu
-if command -v apt-install &> /dev/null; then
+if command -v apt-get &> /dev/null; then
```
ethteck, line 4: I missed this at the first PR, but you can just use `apt` here. `apt install`

install.sh
```diff
@@ -1,14 +1,15 @@
...
     echo "Installing packages for Ubuntu"
 
-    sudo apt install -y git build-essential binutils-mips-linux-gnu zlib1g-dev libcapstone-dev libyaml-dev gcc-multilib || exit 1
+    sudo apt-get install -y git python3 python3-pip build-essential binutils-mips-linux-gnu zlib1g-dev libyaml-dev gcc-multilib || exit 1
```
ethteck, line 7: `sudo apt install`

install.sh
```diff
@@ -1,14 +1,15 @@
...
+    python3 -m pip install capstone
 
     if [[ $1 == "--extra" ]]; then
         echo "Installing extra"
-        sudo apt install -y python3 python3-pip clang-tidy astyle || exit 1
+        sudo apt-get install -y clang-tidy astyle || exit 1
```
ethteck, line 12: `sudo apt install`

PR #38, opened 2020-08-20
Lots of Cam funcs & updates to Camera struct.

JoshW-7:
Notes:

- Line placement of `Camera* cameras = &gCameras;` and `Camera* camera = &cameras[id];` for each func is necessary for matching. Some funcs work by using `(&gCameras[id])->` instead of using `camera->` whereas others do not, so I defaulted to using `cameras->` where possible.
- Added "func_802CA988.s" due to a splitting issue. The func is WIP and commented out.


src/code_ef070_len_3400.c
```diff
@@ -1,10 +1,100 @@
...
+    (&gCameras[id])->mode = 2;
+    (&gCameras[id])->unk_1C = -1 * round((&gCameras[id])->currentPitch);
+    (&gCameras[id])->unk_1E = -1 * (&gCameras[id])->currentBlendedYawNegated;
+    (&gCameras[id])->unk_06 = 0;
+
+    temp1 = (&gCameras[id])->lookAt_obj[0] - (&gCameras[id])->lookAt_eye[0];
+    temp2 = (&gCameras[id])->lookAt_obj[1] - (&gCameras[id])->lookAt_eye[1];
+    temp3 = (&gCameras[id])->lookAt_obj[2] - (&gCameras[id])->lookAt_eye[2];
+
+    tempSqrt = sqrtf((temp1 * temp1) + (temp2 * temp2) + (temp3 * temp3));
```
pixel-stuck, line 81: you can probably remove the tempSqrt var and just put the assignment in the body of the if statement here to get this to match. Not required for the PR, but just a suggestion

PR #42, opened 2020-08-24
Some more si funcs

ethteck:


src/code_e92d0_len_5da0.c
```diff
@@ -47,21 +47,102 @@ ApiStatus si_handle_break_loop(ScriptInstance* script) {
     return ApiStatus_DONE2;
 }
 
-INCLUDE_ASM("code_e92d0_len_5da0", si_handle_wait);
+ApiStatus si_handle_wait(ScriptInstance* script) {
+    Bytecode* ptrReadPos = script->ptrReadPos;
+    s32 todo = 1; // val can be anything
```
pixel-stuck, line 52: Can this be moved into the scope of the if (just above the if(todo))?

I had Josh do this as well, so if we aren't going to do that then we'd need to change those too.

pixel-stuck: (similar for the other ones)

PR #45, opened 2020-08-30
Decomp of func_8014A52C and WSL script

sonicspiral:
Decomp of `func_8014A52C` from `code_e0b30_len_b80`.
* updates to undefined_syms
* updates to variables.h
* added helper window_terminal.bat for windows users

Original mips to c output:
```
s16 func_8014A52C(void) {
    *(void *)0x80159AE2 = (u16)8;
    *(void *)0x80159AE4 = (u16)8;
    return (u16)8;
}
```


src/code_e0b30_len_b80.c
```diff
@@ -4,7 +4,10 @@ INCLUDE_ASM("code_e0b30_len_b80", get_default_variation_for_song);
 
 INCLUDE_ASM("code_e0b30_len_b80", func_8014A498);
 
-INCLUDE_ASM("code_e0b30_len_b80", func_8014A52C);
+void func_8014A52C(void) {
+    D_80159AE2 = (s16)0x08;
```
ethteck, line 8: These casts shouldn't be needed since these variables are already s16s.

tools/windows_terminal.bat
```diff
@@ -0,0 +1,6 @@
+:: replace with your distro
```
ethteck, line 1: Could you explain what this does exactly? It sets up a few tabs in Windows Terminal? Seems kinda useful, but maybe it should take the distro as a flag or use the main distro by default.

sonicspiral:
Thanks for the feedback! I'll start looking at the comments.

src/code_e0b30_len_b80.c
sonicspiral: Sounds good, made the change under [this commit](https://github.com/ethteck/papermario/pull/45/commits/dedd132adcdb5683dbff926209d3e87d36e87354).

tools/windows_terminal.bat
sonicspiral: That's correct, it sets up tabs in the working directory that one could use mips_to_c or diff.py in. I double click on the .bat file and it opens the tabs for me. I like that idea, I'll look into setting the default distro automatically.

ethteck: awesome! thanks

sonicspiral:
@ethteck I'll propose the terminal with the default distro in another PR., wasn't able to find a quick solution to it. Feel free to merge  when ready 👍.

ethteck:
sounds good! kk

PR #46, opened 2020-09-02
Finished code_6000.c

ethteck:


src/code_6000.c
```diff
@@ -50,16 +48,25 @@ s32 heap_free(s32 size) {
...
+s32 collision_heap_malloc(s32 size) {
+    if (!(*gGameStatusPtr)->isBattle) {
+        return _heap_malloc(&D_80268000, size);
+    } else {
+        return _heap_malloc(&D_803DA800, size);
+    }
+}
 
-#ifdef NON_MATCHING
 s32 collision_heap_free(s32 size) {
```
pixel-stuck, line 66: Is this really size? Looking up a heap entry just via size seems uhhh, well, suspect to say the least. Usually free takes in a  pointer and not a size. (Also seems like _heap_free can handle multiple heaps so you need to tell it which one, hence the first ptr). I'd make the arg to this func take in a ``void *`` to a var called `buffer` or `buf` or so.

ethteck: Oh yeah oops it's a pointer, not a size. clover calls it "data" which makes sense to me. idk

pixel-stuck: Data sounds like a fine name to me

src/code_42e0_len_1f60.c
```diff
@@ -10,7 +10,7 @@ INCLUDE_ASM("code_42e0_len_1f60", func_80028FE0);
 
 INCLUDE_API_ASM("code_42e0_len_1f60", length2D);
 
-HeapNode* INCLUDE_ASM("code_42e0_len_1f60", _heap_create, s32 addr, s32 size);
+HeapNode* INCLUDE_ASM("code_42e0_len_1f60", _heap_create, u32* addr, s32 size);
```
nanaian, line 13: Perhaps this should be `void*`

ethteck: yee fixed.

src/code_6000.c
```diff
@@ -23,7 +21,7 @@ s32 general_heap_free(s32 size) {
 }
 
 s32 battle_heap_create(void) {
-    if (_heap_create(&D_803DA800, 0x25800) == -1) {
+    if ((s32)_heap_create(&D_803DA800, 0x25800) == -1) {
```
pixel-stuck, line 24: why the s32 cast?

src/code_6000.c
```diff
@@ -50,16 +48,25 @@ s32 heap_free(s32 size) {
...
+s32 collision_heap_malloc(s32 size) {
+    if (!(*gGameStatusPtr)->isBattle) {
+        return _heap_malloc(&D_80268000, size);
+    } else {
+        return _heap_malloc(&D_803DA800, size);
+    }
+}
 
-    return _heap_free((isBattle == 0) ? (&D_80268000) : (&D_803DA800), size);
+s32 collision_heap_free(UNK_PTR data) {
```
pixel-stuck, line 66: imo, this shouldn't be UNK_PTR, it should be a generic pointer type (void*), which in C can take in any kind of pointer and thus is used for free, memcpy, etc.

PR #60, opened 2020-10-03
Random stuff

ethteck:
* Updated diff.py
* Modified a libultra header so mips_to_c's context feature worked with the repo
* Lots of new functions decompiled
* m2ctx.py


src/code_3bf0_len_6f0.c
```diff
@@ -17,8 +17,8 @@ void func_800287F0(void) {
     {
         GameStatus* gameStatus = *gGameStatusPtr;
```
pixel-stuck, line 18: maybe out of scope for this, but this should really be GAME_STATUS

src/code_7bb60_len_41b0.c
```diff
@@ -121,71 +140,117 @@ void set_action_state(s32 actionState) {
...
 }
 
 INCLUDE_ASM(s32, "code_7bb60_len_41b0", update_locomotion_state);
-
-INCLUDE_ASM(s32, "code_7bb60_len_41b0", start_falling);
-
-INCLUDE_ASM(s32, "code_7bb60_len_41b0", start_bounce_a);
-
-INCLUDE_ASM(s32, "code_7bb60_len_41b0", start_bounce_b);
+/*void update_locomotion_state(void) {
```
pixel-stuck, line 193: What's up with this? At least a comment should be left, but if its functionally equivalent, ifdef NON_MATCHING, preferably.

src/code_e940_len_290.c
```diff
@@ -1,11 +1,89 @@
...
+    }
+
+    D_8009A650[0] = 0;
+    gameStatus->areaID = 0;
+    gameStatus->isBattle = 0;
+    gameStatus->prevArea = -1;
+    gameStatus->mapID = 0;
+    gameStatus->entryID = 0;
+
+    (*gGameStatusPtr)->unk_76 = 0;
```
pixel-stuck, line 25: GAME_STATUS on these

pixel-stuck (changes requested):

src/code_7bb60_len_41b0.c
ethteck: Oops, that was supposed to be removed from the PR.

src/code_e940_len_290.c
ethteck: I don't think it makes sense to do that, since I'm really not sure if these are even correct in the first place. Using the macro feels more like a statement despite it not making a technical difference. In general I don't really like the idea of cleaning up NON_MATCHINGs in this way. I see why you said it though. Honestly, I'd prefer something with a little less commitment like "IN_PROGRESS"

pixel-stuck (approved):

PR #71, opened 2020-10-15
Update install.sh to include the dependencies of diff.py for --extra

Gorialis:
`install.sh` doesn't install certain dependencies like `colorama`, even though it does install other ones:
https://github.com/Gorialis/papermario/blob/1f54aaefa1605837375e8f91ef792294b88a18a1/install.sh#L8
https://github.com/Gorialis/papermario/blob/1f54aaefa1605837375e8f91ef792294b88a18a1/install.sh#L13

However, `colorama` and others are required by `diff.py` and it will fail without them:
https://github.com/Gorialis/papermario/blob/1f54aaefa1605837375e8f91ef792294b88a18a1/diff.py#L218-L228

This adjusts install.sh so `install.sh --extra` will install the extra `diff.py` dependencies.
This also pulls the dependencies out of `install.sh` into their own `requirements.txt` and `requirements_extra.txt` so separate distros will not have to maintain their own Python dependencies in the future.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -5,12 +5,12 @@ if command -v apt &> /dev/null; then
     echo "Installing packages for Ubuntu"
 
     sudo apt install -y git python3 python3-pip build-essential binutils-mips-linux-gnu zlib1g-dev libyaml-dev gcc-multilib || exit 1
-    python3 -m pip install capstone
+    python3 -m pip install -U -r requirements.txt
 
     if [[ $1 == "--extra" ]]; then
         echo "Installing extra"
         sudo apt install -y clang-tidy astyle || exit 1
-        python3 -m pip install stringcase || exit 1
+        python3 -m pip install -U -r requirements_extra.txt || exit 1
     fi
 
     echo "Done"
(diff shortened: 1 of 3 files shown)


pixel-stuck (commented):
This looks good to me. Holding off on approving, as I'm curious if we should keep install.sh and the associated dependency files in the root directory of the repo. install.sh is usually only used once when bootstrapping a new repo, and tool-wise, we usually reserve the root for tools which are used often. Additionally, the dependency files add to the clutter in the root. I'll wait for Ethan to discuss.

Gorialis:
I've added `pycparser` and `PyYAML` to the main dependencies, as these are required by n64splat in `make setup`.

nanaian (approved):
LGTM; using `requirements.txt` was a good call.

@pixel-stuck I'm fine with moving `install.sh` and the requirements files to `tools/` (even though it isn't really a _tool_ per se). I personally see it as out-of-scope for this PR but if @Gorialis wants to move the files, make sure to update the relevant lines in the README.

nanaian:
Relevant discussion:
image

PR #73, opened 2020-10-16
Cornucopia 2

ethteck:


include/common_structs.h
```diff
@@ -1282,7 +1285,7 @@ typedef struct Actor {
     /* 0x211 */ u8 debuffDuration;
     /* 0x212 */ u8 staticStatus; /* 0B = yes */
     /* 0x213 */ u8 staticDuration;
-    /* 0x214 */ u8 stoneStatus; /* 0C = yes */
+    /* 0x214 */ s8 stoneStatus; /* 0C = yes */
     /* 0x215 */ u8 stoneDuration;
     /* 0x216 */ u8 koStatus; /* 0D = yes */
```
nanaian, line 1290: Very likely all of these are s8 given that stoneStatus is s8.

nanaian (approved):

include/variables.h
```diff
@@ -34,6 +34,8 @@ extern s32 gScriptListCount;
 extern s32 gScriptIdList[MAX_SCRIPTS];
 extern s32 gScriptIndexList[MAX_SCRIPTS];
 
+extern s32 gMoveScriptTable[10][4]; // not 10 - how many?
```
nanaian, line 37: ?

include/common_structs.h
ethteck: yeah, but we'll find out in due time when a function that uses them is decomped

nanaian:
WIP now, do not merge

src/code_190B20.c
```diff
@@ -130,23 +284,152 @@ INCLUDE_ASM(s32, "code_190B20", func_8026709C);
...
+}
+
+void remove_part_shadow(s32 actorId, s32 partIndex) {
+    ActorPart* part = get_actor_part(get_actor(actorId), partIndex);
+
+    part->flags |= 4;
+    func_80112328(part->shadow);
+}
+
+void create_part_shadow_by_ptr(UNK_TYPE arg0, ActorPart* part) {
```
pixel-stuck, line 360: should we use "by reference" rather than "by pointer" for this?

src/code_190B20.c
```diff
@@ -42,9 +46,13 @@ INCLUDE_ASM(s32, "code_190B20", func_80263E08);
 
 INCLUDE_ASM(s32, "code_190B20", set_animation_rate);
 
-INCLUDE_ASM(s32, "code_190B20", set_actor_yaw);
+void set_actor_yaw(s32 actorId, s32 yaw) {
+    get_actor(actorId)->yaw = yaw;
```
pixel-stuck, line 50: we should really decide what we're doing with these weird accesses. I hate them, but if everyone else wants them, then we should go for it

pixel-stuck (changes requested):
Obviously I don't think this PR is done yet, so I won't comment on some things, but this is what I see so far that isn't "obvious"

src/code_190B20.c
nanaian: Yes. Maybe "by_ref"?

format.sh
```diff
@@ -21,7 +21,7 @@ astyle ${FILES} \
     --lineend=linux \
     --convert-tabs \
     --max-code-length=120 \
-    --min-conditional-indent=1 \
+    --min-conditional-indent=0 \
```
pixel-stuck, line 24: What's up with this format.sh change?

pixel-stuck (approved):
Other than the format.sh thing, lgtm

PR #82, opened 2020-10-19
New distro support: openSUSE, Alpine Linux

Gorialis:
This adds dependency installation support for openSUSE Leap, openSUSE Tumbleweed and Alpine Linux to `./install.sh`.

It also replaces `tools/cc1` and `tools/mips-nintendo-nu64-as` with statically built versions. This means they do not depend on the `multilib`/`lib32` packages, and can run natively on a 64-bit host.

The 32-bit package dependencies have been therefore removed from the install script for the other distros (which is advantageous as distros such as Alpine Linux do not have them).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,10 +1,10 @@
 #!/bin/bash
 
-# Ubuntu
+# Ubuntu (apt)
 if command -v apt &> /dev/null; then
-    echo "Installing packages for Ubuntu"
+    echo "Installing packages for Ubuntu (apt)"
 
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -58,6 +58,103 @@ if command -v pacman &> /dev/null; then
...
+        python3 -m pip install -U -r requirements_extra.txt || exit 1
+    fi
+
+    echo "Done"
+    exit
+
+fi
+
+
+echo "Only the following distros are supported by install.sh:"
```
nanaian, line 154: I would drop "only," but other than that :+1:

nanaian (approved):

pixel-stuck (approved):

PR #89, opened 2020-10-20
Cornucopia: The Squeakquel

ethteck:


nanaian:
Can we freeze this (i.e. @ethteck develops on another branch) so it can be reviewed and merged

nanaian (approved):
Looks good! Some function names (e.g. "SomeXYZFuncTodoRename") are...questionable, but we can easily change them in the future once we understand in what context they are used.

After #91 is merged, remove `static` from the API functions and `#define FuncName M(FuncName)` in all these `.inc.c` files.

nanaian (changes requested):
> remove static from the API functions and #define FuncName M(FuncName) in all these .inc.c files.

Good to merge once this is done

PR #94, opened 2020-10-23
New Makefile

nanaian:
As discussed, the existing Makefile is somewhat inflexible (especially for modding) and in fact has a bunch of recursion issues in master right now.
So I kinda rewrote it to take a config defined by `sources.mk` and be less slow.


pixel-stuck (approved):
Thanks for doing this. It all looks great to me :)

nanaian:
@ethteck it would be cool to see Jenkins attempt to build `make COMPARE=0 NON_MATCHING=1` builds for PRs - it doesn't compile right now, but it really should.

PR #98, opened 2020-10-26
Cornucopia 3D

ethteck:
some si stuff and assorted core engine functions, a few map functions


pixel-stuck (approved):
lgtm

include/common_structs.h
```diff
@@ -104,7 +104,7 @@ typedef struct Npc {
     /* 0x08A */ s16 yawCamOffset;
     /* 0x08C */ char unk_8C[2];
     /* 0x08E */ s16 duration; /* formerly interp_counter */
-    /* 0x090 */ s16 homePos[3];
+    /* 0x090 */ Vec3s homePos;
```
nanaian, line 107: Maybe drop "pos"? Position is kinda implied by the `Vec` type and hungarian notation bad. Same with other fields but that's out-of-scope for this PR.

src/code_18F340.c
```diff
@@ -17,18 +17,19 @@ ApiStatus func_80260DB8(ScriptInstance* script, s32 isInitialCall) {
...
+        return ApiStatus_DONE2;
+    }
 
-        if (var >= rand_int(100)) {
-            script->varTable[0] = 1;
-        } else {
-            script->varTable[0] = 0;
-        }
+    var = player->varTable[0];
+    if (var >= rand_int(100)) {
```
nanaian, line 27: Why not `if (player->varTable[0] >= rand_int(100))`?

src/si.c
```diff
@@ -4,7 +4,7 @@ s32 si_find_label(ScriptInstance* script, s32 arg1);
 s32 si_skip_if(ScriptInstance* script);
 s32 si_skip_else(ScriptInstance* script);
 s32 si_goto_end_loop(ScriptInstance* script);
-s32 si_goto_end_case(ScriptInstance* script);
+Bytecode* si_goto_end_case(ScriptInstance* script);
```
nanaian, line 7: :+1:

The other `si_goto` functions (should really be `si_find` IMO) will return `Bytecode*` too I expect.

src/world/area_trd/trd_03/9A4650.c
```diff
@@ -1,5 +1,22 @@
 #include "trd_03.h"
 
-INCLUDE_ASM(s32, "world/area_trd/trd_03/9A4650", func_80240000_9A4650);
+static ApiStatus func_80240000_9A4650(ScriptInstance* script, s32 isInitialCall) {
+    Bytecode* args = script->ptrReadPos;
+    u32 stickX;
 
-INCLUDE_ASM(s32, "world/area_trd/trd_03/9A4650", func_8024006C_9A46BC);
+    stickX = 0;
+    if (gCollisionStatus.pushingAgainstWall == 35) {
```
nanaian, line 8: Map shape and hit data have string names in them for all models and colliders.

Probably worth adding a comment that this collider is called `o84`, or even a `#define` in the header: once we have `trd_03_hit` data dumped we can generate these defines but it'd be good to avoid magic numbers even before that point.

src/world/area_trd/trd_03/9A4650.c
```diff
@@ -1,5 +1,22 @@
...
+        stickX = ((u32)GAME_STATUS->stickX) >> 0x1F;
```
nanaian, line 9: Is this the simplest this cast can get? Maybe `stickX` is a `s32` given that it is an `s32` globally...

src/world/area_trd/trd_03/9A4650.c
```diff
@@ -1,5 +1,22 @@
...
+    }
+    if ((gCollisionStatus.pushingAgainstWall == 37) && (GAME_STATUS->stickX > 0)) {
```
nanaian, line 11: This collider is called `o85`

src/world/common/SomeItemEntityFunc.inc.c
```diff
@@ -0,0 +1,31 @@
...
+    if (isInitialCall) {
+        script->functionTemp[2].f = itemEntity->position.y;
+        script->functionTemp[1].s = 0;
+        script->functionTemp[3].s = 0;
+    }
+
+    switch (script->functionTemp[1].s) {
+        case 0:
+            itemEntity->position.y = script->functionTemp[2].f + ((1.0f - cos_rad((script->functionTemp[3].s *
+                                     (1.570796f)) / 30.0f)) * 20.0f);
```
nanaian, line 15: `1.570796f` should be `PI / 2.0f`.

src/world/area_trd/trd_03/9A4650.c
nanaian: e.g.
```c
// trd_03.h

#define COLLIDER_o84 35
#define COLLIDER_o85 37
```

nanaian: (Note not all colliders/models are named `oXX`; some have descriptive romanji names; see `kmr_12/events.c` for reference. I need to set up similar `#define`s for kmr_12 too...)

src/world/area_trd/trd_03/9A4650.c
pixel-stuck: missed this... it may be `stickX = GAME_STATUS->stickX < 0;` or so

include/common_structs.h
ethteck: I feel like simply "home" is a poor name for what is presumably some sort of default position.

src/code_18F340.c
ethteck: becasue it's needed for matching

src/world/area_trd/trd_03/9A4650.c
ethteck: Nice, thanks @pixel-stuck - I knew I was missing something there..

nanaian (approved):

PR #99, opened 2020-10-27
Cornucopia 4nucopia

ethteck:


src/code_1A5830.c
```diff
@@ -94,41 +231,164 @@ INCLUDE_ASM(s32, "code_1A5830", EnemyTestTarget);
...
 
 INCLUDE_ASM(s32, "code_1A5830", func_8027D434);
 
 INCLUDE_ASM(s32, "code_1A5830", func_8027D4C8);
 
-INCLUDE_ASM(s32, "code_1A5830", EnableActorBlur);
+ApiStatus EnableActorBlur(ScriptInstance* script, s32 isInitialCall) {
+    Bytecode* args = script->ptrReadPos;
+    ActorId actorID = get_variable(script, *args++);
+    s32 var1 = get_variable(script, *args++);
```
nanaian, line 277: Can call this `enable`

src/code_fa4c0_len_3bf0.c
```diff
@@ -22,23 +22,60 @@ ApiStatus SetMusicTrack(ScriptInstance* script, s32 isInitialCall) {
...
+ApiStatus FadeInMusic(ScriptInstance* script, s32 isInitialCall) {
+    Bytecode* args = script->ptrReadPos;
+    s32 var0 = get_variable(script, *args++);
+    s32 var1 = get_variable(script, *args++);
+    s32 var2 = get_variable(script, *args++);
+    s32 var3 = get_variable(script, *args++);
+    s16 var4 = get_variable(script, *args++);
+    s16 var5 = get_variable(script, *args++);
+
+    return (func_8014A964(var0, var1, var2, var3, var4, var5) != 0) * 2;
```
nanaian, line 34: Can this be expressed with an `if` or a ternary in a matching way? The boolean arithmetic here isn't very fun

src/code_fa4c0_len_3bf0.c
```diff
@@ -22,23 +22,60 @@ ApiStatus SetMusicTrack(ScriptInstance* script, s32 isInitialCall) {
...
-INCLUDE_ASM(s32, "code_fa4c0_len_3bf0", func_802D5F28);
+ApiStatus func_802D5F28(ScriptInstance* script, s32 isInitialCall) {
+    Bytecode* args = script->ptrReadPos;
+
+    func_8014AA54(get_variable(script, *args++), get_variable(script, *args++), (s16)get_variable(script, *args++));
+    return ApiStatus_DONE2;
+}
+
+ApiStatus func_802D5FA4(ScriptInstance* script, s32 isInitialCall) {
+    func_8014AB0C(0, (s16)get_variable(script, *script->ptrReadPos));
```
nanaian, line 47: Probably worth typing this function so the explicit cast isn't required. Same with func_8014AA54 above.

nanaian (approved):
Cool and good

src/code_fa4c0_len_3bf0.c
ethteck: I would normally agree, but given this is our only call to this function, it's not clear yet if it takes an s16 or it doesn't and the cast is actually not needed. I'd rather not assume yet

src/code_fa4c0_len_3bf0.c
ethteck: We have other functions in the codebase that do this. I imagine it's required, as it is a concise way to return either BLOCK or DONE2. I agree it looks kinda weird, but this is the only way I've been able to write these

src/code_1A5830.c
```diff
@@ -18,19 +18,77 @@ INCLUDE_ASM(s32, "code_1A5830", dispatch_damage_event_actor_0);
...
+    Bytecode* args = script->ptrReadPos;
+    ActorId actorID = get_variable(script, *args++);
+    s32 var1;
+
+    if (actorID == ActorId_SELF) {
+        actorID = script->owner1.actorID;
+    }
+
+    var1 = get_variable(script, *args++);
+    get_actor(actorID)->onHitCode = var1;
```
pixel-stuck, line 75: if we're really going full out, assuming it matches, we might even be able to do:
`get_actor(actorID)->onHitCode = get_variable(script, *args++);`

src/code_fa4c0_len_3bf0.c
pixel-stuck: for clarity, could we potentially multiply it by ApiStatus_DONE2 here, so it's clear what's going on? Otherwise, I think it is indeed what the original devs probably wrote.

pixel-stuck (changes requested):

src/code_fa4c0_len_3bf0.c
ethteck: Totally agree, looks way better that way. dun

src/code_1A5830.c
ethteck: I tried just to see earlier, but we need the temp for matching

pixel-stuck (approved):

PR #100, opened 2020-10-28
100th PR Bonanza: Corn kernels and cornucopiae for all!

ethteck:


src/code_F5750.c
```diff
@@ -238,7 +238,9 @@ INCLUDE_ASM(s32, "code_F5750", func_802D2484);
 
 INCLUDE_ASM(s32, "code_F5750", func_802D249C);
 
-INCLUDE_ASM(s32, "code_F5750", func_802D24F4);
+ApiStatus func_802D24F4(ScriptInstance* script, s32 isInitialCall) {
+    return (gPlayerStatus.moveFrames == 0) * ApiStatus_DONE2;
```
nanaian, line 242: Sigh. Clover thought that they added DONE1 specifically to allow returning a bool, but I guess that's not the case. Maybe we'll find out what DONE1 does soon!

src/code_f2470_len_27f0.c
```diff
@@ -35,11 +35,35 @@ ApiStatus GetNpcPointer(ScriptInstance* script, s32 isInitialCall) {
     return ApiStatus_DONE2;
 }
 
+#ifdef NON_MATCHING
+ApiStatus SetNpcPos(ScriptInstance* script, s32 isInitialCall) {
```
nanaian, line 39: What's non-matching about this? Would be good to add a comment.

src/code_f2470_len_27f0.c
```diff
@@ -35,11 +35,35 @@ ApiStatus GetNpcPointer(ScriptInstance* script, s32 isInitialCall) {
...
+    if (npc == NULL) {
+        return ApiStatus_DONE2;
+    }
+
+    npc->pos.x = x;
+    npc->pos.y = y;
+    npc->pos.z = z;
+    npc->colliderPos.x = x;
+    npc->colliderPos.y = y;
+    npc->colliderPos.z = z;
```
nanaian, line 56: Can maybe do `npc->pos = { x, y, z }` if that's cleaner (I think so) for these if it matches.

src/world/area_mac/mac_00/7EAD10.c
```diff
@@ -4,7 +4,33 @@
...
+        stickX = -stickX;
+    }
+
+    if ((stickX != 0) || (stickY != 0)) {
+        if (atan2(0.0f, 0.0f, stickX, stickY) < 60.0f) {
+            script->varTable[0] = 1;
+            return ApiStatus_DONE2;
+        }
+    }
+    return ApiStatus_BLOCK;
```
nanaian, line 29: Maybe a newline before this

nanaian (approved):
Looks good

src/code_f2470_len_27f0.c
ethteck: I've never done something like this before. Is `{ x, y, z } a struct literal? If so, this is going to generate a struct copy which should be different codegen..but maybe that's why this doesn't match. I'll give it and other things a shot..

nanaian: Yep, it's a struct literal.

nanaian:
Needs SetNpcPos.s deleted and ./format.sh ran, but other than LGTM

src/code_F5750.c
pixel-stuck: I don't see how this negates done1 from being a bool. In this case, they're deciding if they should block or not based *on* a bool value, but aren't attempting to return a bool in particular, so I think that's still on the table?

pixel-stuck (approved):

PR #101, opened 2020-10-28
Makefile go brrrrr

nanaian:


tools/splat.yaml
```diff
@@ -909,6 +829,7 @@ segments:
     - [0x7E3700, "c"]
     - [0x7e4d00, "bin"]
   - name: world/area_mac/machi/
+    ld_name: machi
```
ethteck, line 832: I know this is technically a splat thing and I should have caught it there, but I prefer the LD stuff to all be handled behind the scenes as much as possible. What's the reason for this option? What would happen without it?

sources.mk
```diff
@@ -1,35 +1,57 @@
...
-
-OBJECTS = \
-	$(foreach file, $(C_FILES) $(S_FILES) $(BIN_FILES), $(BUILD_DIR)/$(file).o) \
-	$(foreach file, $(COMPRESSED_BIN_FILES), $(BUILD_DIR)/$(file:.bin=.Yay0.o))
+OBJECTS := $(subst BUILD_DIR, $(BUILD_DIR), $(shell grep -E 'BUILD_DIR.+\.o' papermario.ld -o))
+
+DSL_C_FILES := $(shell grep -lrF "SCRIPT" src)
+
+MAPS := \
+	dro_01 dro_02 \
```
ethteck, line 6: It's a bummer to hard-code these. I assume it's for performance reasons? I wish there was a better way to dynamically support areas.

ethteck (changes requested):

sources.mk
nanaian: They have a specific order that I haven't found a pattern to

tools/splat.yaml
nanaian: It's used for the asset table, which isn't in this PR. It allows us to refer to this as `ROM_machi` etc. (see the generated include/ld_addrs.h).

sources.mk
nanaian: This was in `fs.json` before.

nanaian: Any mods can just use `find` or something, but we have to adhere to the same order used originally.

tools/splat.yaml
ethteck: yeah I feel like we should probably make the ld names different, but again my bad for not really catching that in the splat pr. I kinda knew I'd need to see it in practice to have an opinion.

PR #105, opened 2020-10-30
Audio stuff

nanaian:


src/code_e0b30_len_b80.c
```diff
@@ -1,38 +1,217 @@
...
+/// @returns -1: no override; 0: override to variation 0; 1 override to variation 1
+s32 get_song_variation_override_for_cur_map(SongID songID) {
+    u32 i = 0;
+    Area* areas = gAreas;
+    SongID* allowed = gSongsUsingVariationFlag;
+    GameStatus** gameStatusPtr = gGameStatusPtr;
+
+    for (i = 0; i < ARRAY_COUNT(gSongsUsingVariationFlag); i++) {
+        if (allowed[i] == songID) {
+            GameStatus* gameStatus = *gameStatusPtr;
```
ethteck, line 19: You can't do `= GAME_STATUS` here?

ethteck: or above and reference it here? or try `GAME_STATUS->`

ethteck (dismissed):

src/code_28910_len_5090.c
```diff
@@ -1,16 +1,46 @@
...
+#include "code_28910_len_5090.h"
 
 INCLUDE_ASM(s32, "code_28910_len_5090", func_8004D510);
 
 INCLUDE_ASM(s32, "code_28910_len_5090", func_8004D794);
 
 INCLUDE_ASM(s32, "code_28910_len_5090", func_8004D7E0);
 
-INCLUDE_ASM(s32, "code_28910_len_5090", func_8004DA0C);
+s32 func_8004DA0C(UNK_TYPE arg0) {
```
pixel-stuck, line 9: this function may be able to be rewritten (potentially early return on the initial check of arg0)

src/code_e0b30_len_b80.c
```diff
@@ -1,38 +1,215 @@
...
 
-INCLUDE_ASM(s32, "code_e0b30_len_b80", _set_music_track);
+s32 _set_music_track(s32 playerIndex, SongID songID, s32 variation, s32 fadeOutTime, s16 volume) {
+    GameStatus* gameStatus = GAME_STATUS;
+
+    if (gameStatus->demoState != 0) {
+        return 1;
+    } else {
+        // TODO: these two lines likely a MUSIC_PLAYER(idx) macro
+        MusicPlayer* musicPlayers = gMusicPlayers;
```
pixel-stuck, line 56: my guess is that this pattern might produce the correct codegen but isn't quite correct. Probably they have two instances of this particular line, one in each scope or so, then just always use musicPlayers[idx] or so. Maybe I'm wrong, though

src/code_28910_len_5090.c
nanaian: I wasn't able to get this to match without a `ret` var- the asm stores in s0 then jumps to `move v0, s0; jr ra` which is weird

pixel-stuck (approved):

PR #106, opened 2020-10-30
Match gAreas

nanaian:
(Extends #105. Review and merge that first!)


src/world/world.c
```diff
@@ -0,0 +1,663 @@
...
+    { MAP_UNSPLIT(sam_08, 0x802417D0), .bgName = "yki_bg", .songVariation = 1, .flags = 1 },
+    { MAP_UNSPLIT(sam_09, 0x80240420), .bgName = "yki_bg", .songVariation = 1, .flags = 1 },
+    { MAP_UNSPLIT(sam_10, 0x80240CE0), .bgName = "yki_bg", .songVariation = 1, .flags = 1 },
+    { MAP_UNSPLIT(sam_11, 0x802421A0), .bgName = "yki_bg" },
+    { MAP_UNSPLIT(sam_12, 0x802403B0), .bgName = "yki_bg", .songVariation = 1, .flags = 1 },
+};
+
+/// Crystal Palace
+static Map pra_maps[] = {
+    { MAP_UNSPLIT(pra_01, 0x80241400), .bgName = "yki_bg", .flags = 1, .init = (MapInitialiser*)0x80240000 },
```
ethteck, line 546: I'm just seeing this (`MapInitialiser`) now, not sure if it's new, but we should probably have a conversation about whether we want to use British English or American English for the code itself. I don't think it's as important that we be consistent in the comments, but we should be in the code.

tools/splat.yaml
```diff
@@ -4878,11 +4884,11 @@ segments:
       - [0xC047F0, "bin"]
       - [0xC20C70, "bin"] # rodata
       - [0xC20CF8, "bin"] # rodata
-      - [0xC20F40, "bin"] # rodata
+  - [0xC20F40, "bin"]
   - name: world/area_dgb/dgb_02/
     type: code
     overlay: True
-    start: 0xC2EBA0
+    start: 0xC2EBA0 # this is right
```
ethteck, line 4891: If this is right, do we need a comment saying so? :p

nanaian: lol oops

ethteck (approved):

pixel-stuck (approved):

PR #112, opened 2020-11-02
kmr + more symbol offsets

ethteck:


src/world/area_kmr/kmr_06/8D2A10.c
```diff
@@ -14,4 +14,21 @@ INCLUDE_ASM(s32, "world/area_kmr/kmr_06/8D2A10", func_80241A68_8D2F78);
...
+    s32 mapVar11 = get_variable(script, SI_MAP_VAR(11));
+    s32 mapVar12 = get_variable(script, SI_MAP_VAR(12));
+
+    func_801235C0(mapVar10[1]);
+
+    heap_free(mapVar10);
+    heap_free(mapVar11);
+    heap_free(mapVar12);
+
+    set_variable(script, SI_MAP_VAR(10), 0);
```
pixel-stuck, line 29: I kinda feel like these should be NULL (considering they were pointers), but idk, not sure how I feel.

ethteck: yeah idk because it's unclear what they are and if they're all the same, etc. I feel it should just be revisited and cleaned up later at which point that can be fixed

tools/symbol_addrs.txt
```diff
@@ -241,7 +241,7 @@ render_effects = 0x80059EF4;
 remove_effect = 0x8005A450;
 remove_all_effects = 0x8005A4E0;
 play_effect = 0x8005A564;
-load_map_script_lib = 0x8005A6C0;
+load_world_script_api = 0x8005A6C0;
```
nanaian, line 244: :+1:

src/world/area_kmr/kmr_10/8D84D0.c
```diff
@@ -8,7 +8,22 @@ INCLUDE_ASM(s32, "world/area_kmr/kmr_10/8D84D0", func_80240000_8D84D0);
 
 #include "world/common/CheckItemFlags40.inc.c"
 
+#ifdef NON_MATCHING
+ApiStatus func_802402BC_8D878C(ScriptInstance* script, s32 isInitialCall) {
```
nanaian, line 12: why the non-match?

src/world/area_kmr/kmr_06/8D2A10.c
nanaian: I agree with @pixel-stuck. Once we work out exactly what these are we can add `#define FOO SI_MAP_VAR(10)` etc in the header also

nanaian (approved):

PR #118, opened 2020-11-08
Script DSL updates

nanaian:
This PR:
- Updates `disasm_script.py`'s `symbol_addrs.txt` parser to work with the new format
- Adds `match` syntax (see below)
- Adds `if a ? b` syntax (runs block if `(a & b) != 0`)
- Adds `matchc`, `=c`, and `=f` (also `+=f` etc) for explicitly using variables as constants/floats
- Changes some si.h function and macro names to be more in-line with DSL syntax

---

I'll write proper script DSL documentation at some point, but here's how `match` works:

```c
// Match statements are similar to C's switch statement, but lack fallthrough and support more checks than just equal-to.
match A {
    B {
        // runs if A == B (produces SI_CASE_OR_EQ)
    }
    == C {
        // runs if A == C (produces SI_CASE_EQ)
    }
    // other operators such as >, <, >=, <= are supported
    ? D {
        // runs if (A & D) != 0, i.e. flag D is set
    }
    E, F {
        // runs if A == E or A == F
    }
    G..H {
        // runs if A is in the range G to H
    }
    > I, J, K {
        // runs if A > I or A == J or A == K (produces SI_CASE_GT, SI_CASE_OR_EQ, SI_CASE_OR_EQ)
    }
    else {
        // runs if no other case matched
    }
}
```


ethteck (approved):

src/world/area_kmr/kmr_03/8C8140.c
```diff
@@ -15,60 +15,38 @@ ApiStatus func_802401B0_8C8140(ScriptInstance* script, s32 isInitialCall) {
...
-Script M(script_ExitWalk_802406F0) = {
-    SI_GROUP(27),
-    SI_CALL(0x802D216C, 60, 0),
-    SI_EXEC(0x80285CF4),
-    SI_CALL(0x802CA400, "kmr_04", 0),
-    SI_WAIT_FRAMES(100),
-    SI_RETURN(),
-    SI_END(),
-};
+Script M(ExitWalk_802406F0) = EXIT_WALK_SCRIPT(60, 0, "kmr_04", 0);
```
ethteck, line 18: does disasm_map produce these now?

ethteck (approved):

src/world/area_kmr/kmr_03/8C8140.c
nanaian: No, we've had this macro for a while and I just noticed it

pixel-stuck (approved):

PR #120, opened 2020-11-10
Birth of the Pause Menu

ethteck:


src/os/osEPiWriteIo.c
```diff
@@ -0,0 +1,15 @@
+#include "common.h"
+
+#ifdef MOVE_ISSUE
```
nanaian, line 3: MOVE_ISSUE?

src/os/osEPiWriteIo.c
```diff
@@ -0,0 +1,15 @@
...
+s32 osEPiWriteIo(OSPiHandle* handle, u32 devAddr, u32 data) {
+    register s32 ret;
```
nanaian, line 5: :o

tools/assist.py
```diff
@@ -171,6 +175,18 @@ def do_query(query):
...
+
+def do_cross_query():
+    clusters = []
+
+    for sym_name in map_syms:
+        sym = map_syms[sym_name]
+        if get_symbol_length(sym_name) > 8:
+            cluster_match = False
+            for cluster in clusters:
+                pass # todo do
```
nanaian, line 187: Left a todo here, is it meant to be done?

src/os/osEPiWriteIo.c
ethteck: Yeah, I've added this before. It's for libultra, where the move pseudoinstruction gets assembled into the wrong thing. addu/addiu/or all can work, and it's doing the wrong one

tools/assist.py
ethteck: yeah, I'll do it over time but it's just an enhancement

nanaian (approved):

tools/assist.py
nanaian: IMO unfinished stuff that doesn't somewhat have use (since this seems like it won't output anything?) shouldn't be merged in, but thats your call

ethteck: I generally agree, but since I'm the only one who uses this tool, I don't think this affects anyone, similar to how substitute.py exists and I'm the only one who uses that. But I'll revert it for now

pixel-stuck (approved):

PR #121, opened 2020-11-11
NPC Sprites

nanaian:
A header file for each sprite is created in `include/sprite`. I originally intended to make an aggregate file with all animation ids etc (similar to `enums.h`), but this forced every single map C file to have to rebuild whenever you changed a sprite palette name or something. Instead, discrete headers for each sprite are generated and maps include what they use (e.g. `kmr_12/events.c` includes `sprite/npc/goomba.h` directly).


ethteck (approved):

tools/splat.yaml
```diff
@@ -7028,241 +7028,338 @@ segments:
...
-  - [0x1B79F08, "Yay0"]
-  - [0x1B7AA08, "Yay0"]
-  - [0x1B7B008, "Yay0"]
-  - [0x1B7BB50, "Yay0"]
-  - [0x1B7EC68, "Yay0"]
-  - [0x1B7FF48, "Yay0"]
-  - [0x1B81E88, "Yay0"]
-  - [0x1B82058, "Yay0"]
-  - [0x1B82202, "bin"]
+  - start: 0x19E67B8
```
pixel-stuck, line 7031: does n64splat properly pad this? Given the address used to be 0x19E67B2 and is now 0x19E67B8; even if it does, maybe we should have some kind of ALIGN thing in splat? Not sure. Second part is a bit out of scope for this PR, though.

pixel-stuck (approved):
what's up with tools/sort_symbol_addrs.py? Github is showing it as an empty file. After merge conflicts are resolved, lgtm.

tools/splat.yaml
ethteck: oh yipes. I actually am not sure... I mean there's no code to manually handle alignment of sections or anything, but there probably should be..

nanaian: Well, the build matches... this will be resolved once I get around to implementing player sprites and the sprite table header (at this address IIRC) in a future PR.

nanaian:
@pixel-stuck I made `tools/sort_symbol_addrs.py` executable

ethteck:
will merge once conflict is resolved. (sorry for the ugly conflict)

PR #123, opened 2020-11-12
Fix a couple functions

Kelebek1:


pixel-stuck:
run ./coverage.py --delete-matches

src/code_190B20.c
```diff
@@ -284,23 +281,19 @@ INCLUDE_ASM(s32, "code_190B20", func_8026709C);
...
-// Should be easy to clean up once DecorationTable is more understood
 void add_part_decoration(ActorPart* part, s32 decorationIndex, DecorationId decorationType) {
-    if (part->idleAnimations && (part->flags & 2) == 0) {
-        DecorationTable* decoration = &part->decorationTable->unk_00[decorationIndex];
+    if ((part->idleAnimations) && !(part->flags & 2)) {
+        char* decorationTable = part->decorationTable->unk_00;
+        DecorationTable* decoration;
+
         _remove_part_decoration(part, decorationIndex);
+        decoration = decorationTable + decorationIndex;
```
ethteck, line 290: It'd be nice if there's some way to do this by using proper array index accessing syntax

src/code_190B20.c
```diff
@@ -103,22 +102,20 @@ s32 lookup_defense(DefenseTableEntry* defenseTable, Element elementKey) {
     DefenseTableEntry* row;
     s32 normalDefense = 0;
 
-    for (row = defenseTable; row->element != Element_END; row++) {
+    for (row = defenseTable; row->element != Element_END; row++, defenseTable++) {
         if (row->element == Element_NORMAL) {
-            normalDefense = row->defense;
+            normalDefense = defenseTable->defense;
```
pixel-stuck, line 107: Is reusing defense table here necessary? Seems like the main change is the assignment and break below this. At the very least, can you try without the reuse?

Kelebek1: Yes it's required: https://i.imgur.com/MWxR6ql.png -- the problem with the function is that there are 2 iterators, both a0 (defenseTable) and v1 (row) and they increment differently and are used differently. a0 is used for the 2 loads inside the ifs, whereas v1 is used for the if conditions themselves.

pixel-stuck: :+1:

src/code_190B20.c
nanaian: Agreed; it looks like this is what I tried to do originally, which made this not match. Perhaps we'll work it out later when we understand the decorationTable better.

src/code_190B20.c
nanaian: Very awkward that they did this; good job matching it!

nanaian (approved):
Looks good! Thanks for cleaning up my functions :shipit:

src/code_190B20.c
nanaian: Agreed; it looks like this is what I tried to do originally, which made this not match. Perhaps we'll work it out later when we understand the decorationTable better.

pixel-stuck: This, in particular, seems just wrong. Likely, there's a bunch of arrays in the struct (ones which we haven't defined yet) and those arrays are most likely all some byte-sized arrays. So the compiler can optimize accessing each array with this index into just adding the index to the base offset of the struct and then accessing everything else from a fixed offset from there. I think this should most likely be explored before we just merge this.

ethteck (approved):

src/code_190B20.c
pixel-stuck: Yeah, my above suspicions were correct. I modified the struct a bit, though it may still not be correct, and managed to get a match: https://hastebin.com/ogocunazos.cpp

Still not sure if this struct is correct, but the optimization the compiler does is confirmed with this. Not sure how I feel about having this struct and func in here, since we're still not sure that it's correct.

nanaian (approved):

pixel-stuck (approved):
lgtm

PR #124, opened 2020-11-13
More Pause Stuff + Extras

ethteck:


src/code_ddaf0_len_c50.c
```diff
@@ -1,22 +1,12 @@
...
-    /* 0x00 */ u8 unk_00;
-    /* 0x04 */ s32 unk_04;
-    /* 0x08 */ s32 unk_08;
-    /* 0x0C */ char unk_01[0x10];
-    /* 0x1C */ u8 unk_1C;
-} struct_80159D50; // size = 0x20
-
-extern struct_80159D50 D_80159D50[0x40];
-
+// this is definitely wrong but I had issues trying to use an array accessor
```
pixel-stuck, line 3: Nah, I came to the conclusion that they tracked it separately as well. It's kinda neat that gcc can make that distinction.

src/code_138CC0.c
```diff
@@ -94,6 +117,31 @@ INCLUDE_ASM(s32, "code_138CC0", pause_badges_init);
...
+    if ((temp < 2) || currentMenuPage->numRows < 9) {
+        D_80270394 = 0;
+    } else if (temp >= currentMenuPage->numRows - 2) {
+        D_80270394 = currentMenuPage->numRows - 8;
+    } else {
+        s32* unkSym = &D_80270394;
+
+        if (temp - *unkSym >= 7) {
+            *unkSym = temp - 6;
+        } else {
```
pixel-stuck, line 135: can this be `else if(condition) {` ?

src/code_ddaf0_len_c50.c
ethteck: Even if they're incrementing the struct pointer and using that, I still think something about this is wrong. What do you mean about gcc making a distinction?

PR #127, opened 2020-11-22
Oft-Duplicated Map Funcs

ethteck:
+ some extra goodies


src/world/common/UnkNpcDurationFlagFunc.inc.c
```diff
@@ -0,0 +1,16 @@
+#include "common.h"
+#include "map.h"
+
+static void UnkNpcDurationFlagFunc(ScriptInstance* script) {
+    Enemy* enemy = script->owner1.enemy;
+    Npc* npc = get_npc_unsafe(enemy->npcID);
+
+    npc->duration--;
+    if ((npc->duration << 0x10) == 0) {
```
pixel-stuck, line 9: iirc, this is actually something like & 0xFFFF or so? Not sure, but this at the very least seems incorrect. (also, if possible shift amounts in decimal. Makes it way easier to recognize patterns)

src/world/common/UnkNpcAIFunc8.inc.c
```diff
@@ -0,0 +1,20 @@
+#include "common.h"
+#include "map.h"
+
+static void UnkNpcAIFunc8(ScriptInstance* script) {
+    Enemy* enemy = script->owner1.enemy;
+    Npc* npc = get_npc_unsafe(enemy->npcID);
+
+    npc->duration--;
+    if ((npc->duration << 0x10) <= 0) {
```
pixel-stuck, line 9: I think this (and the other similar ones) can just be npc->duration <= 0

src/world/common/SetGameStatusUnk84_1.inc.c
```diff
@@ -0,0 +1,6 @@
+#include "common.h"
+
+static ApiStatus SetGameStatusUnk84_1(ScriptInstance* script, s32 isInitialCall) {
+    GAME_STATUS->unk_84 = 1;
+    return ApiStatus_BLOCK;
```
pixel-stuck, line 5: hm, interesting that this always block; won't this cause an infloop if called?

pixel-stuck (changes requested):

src/world/common/SetGameStatusUnk84_1.inc.c
ethteck: This either may not be an ApiStatus or the context in which it's used prevents it from being an infinite loop. @nanaian may have more input

nanaian: They definitely do this sometimes. It won't create an infinite loop, just make the script never continue.

pixel-stuck (approved):

PR #131, opened 2020-12-02
remove local variable case camelBack requirement

nanaian:
As discussed with @ethteck, I'm disabling this warning because mips_to_c outputs identifiers like `temp_v0` by default, and seeing the warnings during early function decompilation is unhelpful and annoying (plus we'd likely pick up issues like these in code review).

(are we even sure we want to use camelCase for parameters/struct fields? personally I think I'd prefer snake_case for everything but types... worth discussing this I think)


ethteck:
> (are we even sure we want to use camelCase for parameters/struct fields? personally I think I'd prefer snake_case for everything but types... worth discussing this I think)
I personally am not into snake case at all for c code, and the only reason I think it works in this project is because we have different "types" of functions we're dealing with. But happy to hear other thoughts as always

ethteck (approved):

PR #134, opened 2020-12-04
decomp initialize_status_menu

Rainchus:


src/code_80850_len_3060.c
```diff
@@ -268,7 +268,108 @@ void enforce_hpfp_limits(void) {
...
+    var0 = create_icon(&D_801086AC);
+    uiStatus->starpointsIconIndex = var0;
+    set_icon_flags(var0, 0x80);
+    clear_icon_flags(var0, 0x8000);
+    var0 = create_icon(&D_80104BEC);
+    uiStatus->starpointsIconIndex2 = var0;
+    set_icon_flags(var0, 0x80);
+    clear_icon_flags(var0, 0x8000);
+    var1 = &D_80108068;
+    new_var = create_icon(&D_80108068);
```
ethteck, line 350: are you sure new_var is needed here? should be able to just do var0 = like the rest of these ideally

src/code_80850_len_3060.c
```diff
@@ -268,7 +268,108 @@ void enforce_hpfp_limits(void) {
     }
 }
 
-INCLUDE_ASM(s32, "code_80850_len_3060", initialize_status_menu);
+void initialize_status_menu() {
+    UiStatus* uiStatus = &gUIStatus;
+    PlayerData* playerData = &gPlayerData;
+    s32 var0;
```
ethteck, line 274: var0 can be renamed to iconIndex since it represents an icon index

src/code_80850_len_3060.c
```diff
@@ -268,7 +268,108 @@ void enforce_hpfp_limits(void) {
     }
 }
 
-INCLUDE_ASM(s32, "code_80850_len_3060", initialize_status_menu);
+void initialize_status_menu() {
+    UiStatus* uiStatus = &gUIStatus;
+    PlayerData* playerData = &gPlayerData;
```
ethteck, line 273: You should be able to use UI_STATUS and PLAYER_DATA macros here. If the UI_STATUS one doesn't exist, feel free to create it in macros.h like the player data one

src/code_80850_len_3060.c
```diff
@@ -268,7 +268,108 @@ void enforce_hpfp_limits(void) {
...
+    s32 var0;
+    s32 var1;
+    s32 new_var;
+
+    uiStatus->drawPosX = 0x0C;
```
ethteck, line 278: could you convert drawPosX, drawPosY, and showTimer to decimal?

src/code_80850_len_3060.c
```diff
@@ -268,7 +268,108 @@ void enforce_hpfp_limits(void) {
...
+    uiStatus->unk_3B[1] = 0;
+    uiStatus->unk_57[0] = 0;
+    uiStatus->unk_57[1] = 0;
+    uiStatus->unk_57[2] = 0;
+    uiStatus->unk_57[3] = 0;
+    uiStatus->unk_6C[0] = 0;
+    uiStatus->unk_6C[1] = 0;
+    uiStatus->iconIndex12 = -1;
+    func_800E97B8();
+    var0 = create_icon(&D_80108248);
```
ethteck, line 317: I think each of these blocks that start with var0 and end with clear_icon_flags should be separated by an empty line, just to make it clear that they're separate little things

ethteck (changes requested):
Nice job! Just some style/tidying things

src/code_80850_len_3060.c
nanaian: classic rain

src/code_80850_len_3060.c
Rainchus: just doing
`iconIndex = create_icon(&D_80108068);`
makes s0 and s1 swap in the code

nanaian (approved):

ethteck (approved):

src/code_80850_len_3060.c
```diff
@@ -268,7 +268,121 @@ void enforce_hpfp_limits(void) {
     }
 }
 
-INCLUDE_ASM(s32, "code_80850_len_3060", initialize_status_menu);
+void initialize_status_menu() {
```
pixel-stuck, line 271: void in the arguments for this func

pixel-stuck (changes requested):

src/code_80850_len_3060.c
ethteck: ahh yee, nice catch

if you weren't aware Rain, `()` vs `(void)` are actually kind of different. the latter specifically means 0 args, and the former means anything

PR #135, opened 2020-12-05
Mac support

ethteck:


Makefile
```diff
@@ -58,6 +58,8 @@ CPP := cpp
 LD := $(CROSS)ld
 OBJCOPY := $(CROSS)objcopy
 
+-include cpp.mk # Used to specify alternate CPP if `cpp` is the clang one on MacOS
```
nanaian, line 61: may be better to detect this in the makefile instead or using settings.mk

pixel-stuck (approved):

PR #136, opened 2020-12-22
add Star Rod

nanaian:
This adds a new phony make target which runs a modified version of Star Rod. This allows GUI sprite animation viewing/editing.
```
$ make editor
```

image

The `tools/star-rod` submodule is **not initialised with `make setup`**, but rather lazily cloned when you first run `make editor`. This should hopefully address concerns expressed back when we had Star Rod as a submodule before: it's not a hard dependency at all.

(In the future I'll patch in support for Star Rod's map editor; this will require quite a hefty n64splat extension for dumping map data.)


Makefile
```diff
@@ -115,14 +122,16 @@ tools:
 
 setup: clean submodules tools split $(LD_SCRIPT)
 
+# tools/star-rod submodule intentionally omitted
 submodules:
+	git submodule init tools/n64splat
 	git submodule update --init --recursive
```
ethteck, line 128: --recursive will get star rod too, no?

Makefile
```diff
@@ -68,6 +68,13 @@ ifeq ($(UNAME_S),Darwin)
 	ICONV := tools/iconv.py UTF-8 SHIFT-JIS
 endif
 
+WSL := 0
+JAVA := java
+ifeq ($(findstring microsoft,$(shell cat /proc/sys/kernel/osrelease)),microsoft)
```
ethteck, line 73: I'd prefer we use this method to detect whether we're on windows: https://stackoverflow.com/questions/714100/os-detecting-makefile (the above if chain is from here too). The WSL and JAVA vars can be declared above too I think

ethteck (changes requested):

Makefile
nanaian: Oh, I forgot to drop `--init` here

Makefile
nanaian: We're not detecting Windows, we're detecting _WSL_

ethteck (approved):

pixel-stuck (approved):

PR #138, opened 2020-12-22
match world partner table

nanaian:


Makefile
```diff
@@ -257,10 +258,12 @@ include/sprite/npc/%.h: sprite/npc/%/SpriteSheet.xml tools/gen_sprite_animations
...
 	@echo "building $@"
 	@$(PYTHON) tools/gen_sprite_animations_h.py $@ sprite/npc/$* $(NPC_DIRS)
-include/sprite: $(foreach dir, $(NPC_DIRS), include/$(dir).h)
+
+
+### Linker ###
 
 $(LD_SCRIPT): $(SPLAT_YAML)
-	$(SPLAT) --modes ld
+	$(SPLAT) --modes ld > /dev/null
```
ethteck, line 266: I think we should have a quieter splat output that doesn't show all the stuff instead of this

nanaian:
useAbility and update may be the wrong way around

asm/nonmatchings/code_317020/world_goombario_pre_battle.s
```diff
@@ -1,7 +1,7 @@
 .set noat      # allow manual use of $at
 .set noreorder # don't insert nops after branches
 
-glabel func_802BDBBC
+glabel world_goombario_pre_battle
```
ethteck, line 4: symbol_addrs - I think this is missing?

Makefile
```diff
@@ -257,10 +260,13 @@ include/sprite/npc/%.h: sprite/npc/%/SpriteSheet.xml tools/gen_sprite_animations
...
 	@echo "building $@"
 	@$(PYTHON) tools/gen_sprite_animations_h.py $@ sprite/npc/$* $(NPC_DIRS)
-include/sprite: $(foreach dir, $(NPC_DIRS), include/$(dir).h)
+
+
+### Linker ###
 
 $(LD_SCRIPT): $(SPLAT_YAML)
-	$(SPLAT) --modes ld
+	$(SPLAT) --modes ld bin Yay0 PaperMarioMapFS PaperMarioMessages img PaperMarioNpcSprites --new
```
ethteck, line 268: since this is now splitting out everything required, I think it shouldn't be the LD_SCRIPT target but instead like...split-req or something? open to other names

asm/nonmatchings/code_317020/world_goombario_pre_battle.s
ethteck: oh maybe we don't want the world funcs there? not sure..

Makefile
nanaian: Perhaps. LD_SCRIPT is the primary output - assets only get resplit when needed (with `--new`).

asm/nonmatchings/code_317020/world_goombario_pre_battle.s
nanaian: I'm not certain on what symbol_addrs.txt is for. It's probably relevant that world partner files are overlays

Makefile
nanaian: Making it a phony target would mean we couldn't depend on it as we do now AFAIK

nanaian: For people who want to manually split you can still do `make split` and it'll route to this rule unconditionally

tools/splat.yaml
```diff
@@ -401,71 +407,84 @@ segments:
...
   - type: code
+    ld_name: world_partner_goombario
     start: 0x317020
     vram: 0x802BD100
     files:
-      - [0x317020, c]
-      - [0x317b60, bin]
+      - [0x317020, c, world/partner/goombario]
+      - [0x317b60, .data, world/partner/goombario]
+      - [0x317DA0, bin, world/partner/goombario] # rodata containing a double and a jumptable
```
pixel-stuck, line 416: since we know this is rodata, use the ``rodata`` type in n64splat; this will auto-disassemble the rodata based on the asm in the code associated with the file

asm/nonmatchings/code_317020/world_goombario_pre_battle.s
ethteck: yee nm

Makefile
ethteck: allroyght, we can see how it goes

tools/splat.yaml
ethteck: I don't think we should require the use of this necessarily..the data disassembly is still in the very early stages, so I don't think it makes sense to use it unless you're going to leverage it (migrate the rodata, move it into c, etc)

ethteck (approved):
I approve once the lint check passes

tools/splat.yaml
nanaian: iirc I tried it but it didn't work

src/world/partners.c
```diff
@@ -0,0 +1,216 @@
...
+#include "partner/twink.h"
+#include "sprite/npc/twink.h"
+
+s32 world_partner_can_use_ability_default(Npc* partner);
+s32 world_partner_can_player_pause_default(Npc* partner);
+
+static s32 _pad[] = { 0x00, 0x00, 0x00 };
+
+WorldPartner gWorldPartners[] = {
+    // XXX: it's possible that's there's a "none" entry here to match up with enum PartnerID
```
ethteck, line 38: There's almost definitely a none entry. At least I found a loop that goes through partners and starts at 1. Maybe unrelated actually.

nanaian: Yeah, this pops up a bunch

nanaian: It'd be nice to figure out what RAM address this data is at, then we'd know

PR #139, opened 2020-12-24
Fixup Dockerfile

mkst:
Very similar to my conker [Dockerfile](https://github.com/mkst/conker/blob/master/Dockerfile), but with the addition of `git` and `lark`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,64 +1,25 @@
-FROM ubuntu:18.04 as build
+FROM ubuntu:20.04 as build
+
+ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,64 +1,25 @@
-FROM ubuntu:18.04 as build
+FROM ubuntu:20.04 as build
+
+ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && \
```
nanaian, line 5: It would probably be better to run `./install.sh --extra` rather than hardcode libraries here

nanaian (changes requested):

ethteck (approved):

PR #140, opened 2020-12-24
WIP - Matching code_39cb0_len_100.c

mkst:
Does not yet produce a matching ROM, but the functions themselves match original ASM.

image

**Notes:**
 - `osInitialize` is actually a macro, and needs `_FINALROM` to be defined such that it only calls `__osInitialize_common`.


ethteck:
I looked into this, and the issue is because the D_8009A630 symbol isn't being put into the correct place. However, I'm kind of lost, as it seems to be outside the bounds of this overlay, and I'm not sure where it would belong..math below:

symbol addr: 8009A630
vram start: 80025C00
the rom offset of this symbol (if it exists in this overlay) would be 74A30 into the overlay
the overlay starts at 0x1000 so that'd be rom position 0x75A30

ethteck:
@mkst brought up very smartly that this could be bss, which I am very un-knowledgeable about. we may need to wait until later to figure this out, since we need to map out the rest of the bin space first probably

ethteck:
I cleaned this up a bit and made the function with the .bss NON_MATCHING for now.

ethteck (approved):

src/os/code_39cb0_len_100.c
```diff
@@ -3,31 +3,34 @@
...
-        do { /* nothing */ } while ((temp_v0 = D_8009A630) == NULL);
-        temp_v0();
+        void (*func)(void);
+
+        do {
+            func = D_8009A630;
+        } while (func == NULL);
+
+        func();
     } while (1);
```
nanaian: Probably worth making this `while (TRUE)`, looks good though

nanaian (approved):

pixel-stuck:
> @mkst brought up very smartly that this could be bss, which I am very un-knowledgeable about. we may need to wait until later to figure this out, since we need to map out the rest of the bin space first probably

on this:
Screenshot from 2020-12-29 14-17-15
this is from the spreadsheet on google sheets, so this is indeed bss. I think this should be a fairly simple change, though might require some new splat configuration option, I suppose. Maybe out of scope for this PR, but I did want to bring it up.

pixel-stuck (approved):

PR #141, opened 2020-12-26
Battles

nanaian:


ethteck:
wooooo! After the conflcits are resolved and it passes I approve

ethteck (approved):

tools/splat.yaml
```diff
@@ -925,280 +926,333 @@ segments:
...
-  - name: battle/section_10/
+  - name: battle/area_sbk/
     type: code
     start: 0x4DA5E0
     vram: 0x80218000
+    overlay: True
     files:
-    - [0x4DA5E0, bin]
-  - name: battle/section_11/
+    - [0x4DA5E0, bin] # 1 func
```
pixel-stuck, line 1009: any particular reason not to mark these as .text do splat can split them out?

src/world/world.c
```diff
@@ -657,5 +657,5 @@ Area gAreas[] = {
     AREA(mgm, "ミニゲーム"),  // minige-mu [Minigames]
     AREA(gv, "ゲームオーバー"),  // ge-mu o-ba- [Game Over]
     AREA(tst, "テストマップ"),  // tesuto mappu [Test map]
-    0, 0, 0, 0,
+    {},
```
pixel-stuck, line 660: imo, these should remain as properly zero'd (C doesn't guarantee what these get initialized to; even if they do in this case, could cause problems later with e.g. more modern compilers)

src/world/area_kmr/kmr_12/events.c
```diff
@@ -187,8 +187,8 @@ StaticNpc M(goombaNpc) = {
 };
 
 NpcGroupList M(npcGroupList) = {
-    NPC_GROUP(M(goombaNpc), FORMATION_ID(1, 0, 3)),
-    NPC_GROUP_LIST_END(),
+    NPC_GROUP(M(goombaNpc), BATTLE_ID(1, 0, 3)),
+    {},
```
pixel-stuck, line 191: similar here with the zero'd struct

src/battle/battle.c
```diff
@@ -0,0 +1,366 @@
...
+        .name = "エリア ＯＭＯ２＿５",
+    },
+    {
+        AREA(area_omo2_6),
+        .name = "エリア ＯＭＯ２＿６",
+    },
+    {},
+};
+
+const s32 rodata_75978[] = {
```
pixel-stuck, line 361: hm. Is it possible to put these in a rodata.s file, since this is most likely a jump table, and the data would never be defined in C?

pixel-stuck (changes requested):

src/battle/battle.c
nanaian: Possibly, I have no idea how to do that though

src/world/world.c
nanaian: @ethteck thoughts?

tools/splat.yaml
nanaian: Yeah, @ethteck is going to do it after this is merged

src/world/world.c
ethteck: https://stackoverflow.com/questions/11152160/initializing-a-struct-to-0 From this discussion, it seems this isn't really a big issue, at least with GCC, which is the only compiler I see us caring about in the near future. If it becomes an issue later on, we can zero-initialize everything, but until then this is easy to search for if we need to change it and looks a lot cleaner.

src/battle/battle.c
ethteck: It'd help to have the vram addresses of these rodata_ vars and a comment saying it's most likely a jump table. Beyond that, I think we can merge this as-is and migrate the rodata later on, in the interest of keeping things moving

ethteck: I added the comment

PR #142, opened 2020-12-27
Change #included map functions to use M()

ethteck:
The project no longer has any static functions, which will be an ongoing goal during the decomp phase so we can better track progress.

Also, made a change to n64splat that improves symbol detection and reduces the vast majority of asm warnings we were getting


undefined_syms.txt
```diff
@@ -337,3 +337,19 @@ gSongsUsingVariationFlag = 0x8014F720;
 D_8014F738 = 0x8014F738;
 
 D_802BDF64 = 0x802BDF64;
+D_802BDD88 = 0x802BDD88;
+
+D_802BD58C = 0x802BD58C;
+D_802BDC40 = 0x802BDC40;
+
+
+D_C2BE35B8 = 0xC2BE35B8;
```
pixel-stuck, line 346: are these real addresses?

pixel-stuck: it didn't get the lines I wanted, lol.

ethteck: No, but they're issues with the asm in the area func asm files that I haven't resolved because if I take the new splat output for them, I lose the function name changes. I need to update rename.py's input file but I'll do that later

PR #145, opened 2021-01-01
Battle Code + More

ethteck:
Split out battle code + decomped some of it.

Also includes some other misc decomp


src/code_1A5830.c
```diff
@@ -274,20 +285,37 @@ ApiStatus SetStatusTable(ScriptInstance* script, s32 isInitialCall) {
...
+    ActorID actorID = get_variable(script, *args++);
+    s32 partIndex;
+
+    if (actorID == ActorID_SELF) {
+        actorID = script->owner1.actorID;
+    }
+
+    partIndex = get_variable(script, *args++);
+
+    // weirdly unused
```
pixel-stuck, line 312: if they had more vars that went unused, they still have to advance past the bytecode so that things don't get out of sync

src/code_759b0_len_61b0.c
```diff
@@ -54,27 +71,88 @@ INCLUDE_ASM(s32, "code_759b0_len_61b0", check_input_use_partner);
...
 
 void clear_player_status(void) {
     mem_clear(&gPlayerStatus, sizeof(gPlayerStatus));
 }
 
-INCLUDE_ASM(s32, "code_759b0_len_61b0", func_800DFC74);
+void func_800DFC74(void) {
+    PlayerStatus* playerStatus = PLAYER_STATUS;
+
+    mem_clear(playerStatus, sizeof(PlayerStatus));
```
pixel-stuck, line 95: tangent, but should this mem_clear function maybe be renamed to bzero?

src/code_fa4c0_len_3bf0.c
```diff
@@ -34,7 +34,12 @@ ApiStatus FadeInMusic(ScriptInstance* script, s32 isInitialCall) {
     return (func_8014A964(var0, var1, var2, var3, var4, var5) != 0) * ApiStatus_DONE2;
 }
 
-INCLUDE_ASM(s32, "code_fa4c0_len_3bf0", func_802D5EE0);
+ApiStatus func_802D5EE0(ScriptInstance* script, s32 isInitialCall) {
+    Bytecode* args = script->ptrReadPos;
+
+    (&gMusicPlayers[get_variable(script, *args++)])->flags |= 0x2;
```
pixel-stuck, line 40: does a temp for gMusicPlayers work here? Would be a bit less ugly, imo, and thus preferred

pixel-stuck (changes requested):

src/code_1A5830.c
ethteck: They could just do *args++ for that, but maybe they used get_variable since it's a macro

src/code_759b0_len_61b0.c
ethteck: I'll add a comment but would rather refrain from renaming things until we know more stuff

src/code_fa4c0_len_3bf0.c
ethteck: do you mean a temp for `(&gMusicPlayers[get_variable(script, *args++)])` or just for gMusicPlayers itself?

pixel-stuck: The latter

PR #154, opened 2021-01-15
Ninja

nanaian:
The setup process is now:
1. `./configure.py` - runs n64splat, then generates `build.ninja` based on objects
2. `ninja` - compiles everything, targeting papermario.z64, then checks `OK`

(Please also test this on macOS @ethteck)

`make clean` is now `ninja -t clean`. `ninja -t cleandead` is similar but only removes outputs from previous builds that are no longer in the build file, which is nice. Builds are a lot more deterministic now so cleaning "just to check" will be less required.

`make split`,  `make setup`, etc. are now simply `./configure.py`.

`.c` files now depend on the `.s` files they include with `INCLUDE_ASM`.

`.msg` files get `.h` files generated next to them similar to sprites. This can eventually replace `include/messages.h`'s constants once we allow naming messages in splat.yaml or similar.

Ninja doesn't track changes to `splat.yaml`, so if it changes - or there are otherwise changes to the number of objects needed to build - you need to run `./configure.py` again yourself. This currently takes a lil while because we have to `cpp` all the c files to discover SCRIPT/INCLUDE_ASM usage, but this process could be cached in the future if it becomes an issue.

Issues:
- [x] no pipefail
- [x] `diff.py`, permuter don't work (would appreciate some assistance PRing to these, @ethteck)
- [x] ~~lacks `make editor` equivalent (should probably make this a `./starrod.sh` instead; it's unrelated to the build system)~~ (out-of-scope for this PR)
- [x] ~~no `make submodules` (do we need this? I hear we're switching to subtree soon? what about Star Rod?)~~ (#156)
- [x] no way to specify extra cflags (e.g. `-DNON_MATCHING`) from configure


ethteck:
Initial notes:
Some dependency wasn't added to requirements.txt, forgot which
The Jenkinsfile isn't pulling in 64splat so this can wait on my subrepo change I guess

ethteck:
we'll need to add a configure "target" or something, or a ninja target, or idk, to split code.
also I think the makefiles can be totally removed before this is merged, ideally

nanaian:
From-scratch speed comparison according to Jenkins.

Ninja:
image

Make:
image

CONTRIBUTING.md
```diff
@@ -78,3 +78,5 @@ $ ./format.sh
 ```
 
 If `format.sh` has any problems with your code, go and fix the issues. If you can't fix a warning without making the function not match anymore, append `// NOLINT` to the offending line.
+
+Then, please [create a pull request](https://github.com/pmret/nananian/pulls)!
```
pixel-stuck, line 82: this seems to have your username hardcoded?

INSTALL.md
```diff
@@ -0,0 +1,62 @@
...
+git clone https://github.com/pmret/papermario
+cd papermario
+```
+
+Install build dependencies:
+```sh
+./install.sh
+```
+
+Copy a Paper Mario (USA) ROM into the root directory of the repository with the name `baserom.z64`. If you're using WSL, you can enter the Linux filesystem by opening `\\wsl$` in File Explorer; e.g. `\\wsl$\Ubuntu\home\<your username>\papermario`.
```
pixel-stuck, line 26: I think we should add the sha1 hash here.

pixel-stuck (changes requested):

CONTRIBUTING.md
nanaian: oooooooooooooooooops

nanaian: good spot!

INSTALL.md
nanaian: `./configure.py` actually checks that its valid now, so I think it would be redundant, Additionally the README now links to the database (hashes, size, etc.) for the ROM we expect

pixel-stuck: at least make mention of it, imo

nanaian: should be resolved now

pixel-stuck (approved):

nanaian:
RTM once https://github.com/simonlindholm/decomp-permuter/pull/94 is merged

PR #165, opened 2021-01-28
Matched all stub functions

Mr-Wiseguy:
Automated this by writing a quick python script that finds all files of a given maximum length, replaces the INCLUDE_ASM call with a stub function definition and deletes the asm file.


pixel-stuck:
First, thanks for the PR :)

A few small things: 
 - the way you've done these functions doesn't comply with our current styling; please run format.sh found in the root of the repo. 
 - Most likely, these stub functions do have different arguments for different situations, even if unused... This is probably something to worry about at some other point, though.

Mr-Wiseguy:
I ran format.sh but it didn't change anything so I assumed that they followed the repo's styling. Any idea why it's not doing anything?

pixel-stuck:
potentially format.sh doesn't handle this case. Either way, per other stubs in the repo, the closing curly brace should be on a newline, like so:
```C
void nullsub(void){
}```

Mr-Wiseguy:
I've fixed the formatting of the stubs. As for the arguments, with ido at least even with -O2 the compiler will leave traces behind that indicates the presence of arguments. Do you know if that's not the case with gcc?

pixel-stuck:
gcc indeed is a lot less wasteful of stack space, and unused arguments don't appear to affect the actual function itself. Obviously it can affect codegen of functions that call this one,  though. But that's generally pretty obvious.

Mr-Wiseguy:
In that case, is there any changes you'd like me to make? Quite frankly, I don't think I'd be able to go through all of those functions one by one and determine what arguments they use if the caller function hasn't been decompiled yet. A quick glance at a few showed that they seem to be on average unused, but obviously that's not a comprehensive check.

pixel-stuck:
No, as I mentioned in my second bullet point, this will likely be handled as the callers to these functions are decompiled.

pixel-stuck (approved):

ethteck:
Thanks for your PR @Mr-Wiseguy . are you in the discord? We'd love to have you there if you wanna join.

PR #169, opened 2021-02-06
clean up coconut

nanaian:


ethteck (changes requested):
Looks good, but let's rename aux since windows hates files named aux and we were gonna change it anyway

ethteck (approved):

src/battle/item/coconut/coconut.c
```diff
@@ -31,14 +31,14 @@ Gfx N(coconutDL)[] = {
     gsSPEndDisplayList(),
 };
 
-s32 D_802A1B50_7304B0[] = {
+s32 N(coconutEntity)[] = {
```
pixel-stuck, line 34: Not sure I like the name? Maybe coconutIconScript/coconutIconCommandList or something?

pixel-stuck (dismissed):

src/battle/item/coconut/coconut.c
nanaian: An 'icon script' is a different thing, and a 'command list' is really vague imo. We need to decide what to call these scripts anyway - 'entity animation' may work?

ethteck (approved):

PR #170, opened 2021-02-06
Name many funcs to match upcoming Star Rod 0.4.0 release

nanaian:


ethteck (approved):

ethteck (approved):

src/func_801489B8.s
```diff
@@ -1,7 +1,7 @@
 .set noat      # allow manual use of $at
```
pixel-stuck, line 1: what's up with moving this file?

pixel-stuck (changes requested):

src/func_801489B8.s
nanaian: whoops, will fix

PR #175, opened 2021-02-09
Delete nonmatchings that aren't included in c files

Alto1772:
if you do need these files, just cancel this.


nanaian (approved):
I'm surprised `coverage.py` didn't catch these - thanks!

ethteck:
Indeed, thanks! A lot of these I believe are files that exist elsewhere in the repo but weren't deleted when c files were split. They would probably be caught eventually when they're matched, but deleting them ahead of time is cleaner anyway.

PR #177, opened 2021-02-10
Modify message syntax to match Star Rod

nanaian:
The `.msg` syntax now matches that of `.str` in Star Rod in order to facilitate a work-in-progress GUI message editing feature. Future Star Rod versions will begin transitioning to use "message" instead of "string" to match symbols in this repo, but other than the file extension and `#message` rather than `#string` directives, message files should behave the same no matter if they are compiled via this repo or via Star Rod's existing compiler. (Fingers crossed.)

Messages can now be given names in `splat.yaml`, similar to sprites. This results in `MSG_myname` defines being output to a `message_ids.h` which must be included where needed (I'd rather not include this in `common.h`, because it would mean modifying any dialogue would cause the entirety of src/ to be recompiled).

(@ethteck raised earlier that holding names like this in `splat.yaml` would balloon its file size, but I've since considered that this a) isn't a real problem, there is no hard cap on file sizes! and b) workarounds would add complexity, likely break existing features such as `.splat_cache`, and probably cause issues when we move to versioning [`splat.usa.yaml` etc.]; c) we can easily change this in the future if required, and d) most editors support folding big blocks, plus this one is right at the bottom of the yaml.)

### DIFF
--- a/include/messages.h
+++ b/include/messages.h
@@ -5,12 +5,7 @@
 
 typedef s32 MessageID;
 
+// Prefer editing splat.yaml's msg ids section than using this directly!
 #define MESSAGE_ID(section, index) (((section << 0x10) + index))
-
-#define MessageID_TATTLE_KMR_03 MESSAGE_ID(0x19, 0x3B)
-#define MessageID_TATTLE_KMR_12 MESSAGE_ID(0x19, 0x40)
-
-#define MessageID_SIGN_MUSHROOM_GOOMBA_TRAP MESSAGE_ID(0x1D, 0x167)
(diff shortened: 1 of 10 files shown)


include/messages.h
```diff
@@ -5,12 +5,7 @@
 
 typedef s32 MessageID;
 
+// Prefer editing splat.yaml's msg ids section than using this directly!
 #define MESSAGE_ID(section, index) (((section << 0x10) + index))
```
pixel-stuck, line 9: I know this isn't in-scope for this PR, but the shift in this macro should be in decimal

ethteck (approved):

PR #179, opened 2021-02-10
actor-related structs/messages

nanaian:


nanaian:
Not sure what to do with `slot_machine_start_dup1` to `dup3`; it's only named this way because its configured name/tattle messages are the same as `slot_machine_start`, but the actor ID itself is unused and it's likely it was just a copy-paste from the final time the actor table was expanded. Something similar seems to be the case with `the_master` duplicates.

Do I remove the `#define` for these dummy actor IDs?

pixel-stuck (approved):

Issue #183, opened 2021-02-13
Remove global-ref macros like BATTLE_STATUS


nanaian:
The layer of abstraction is pointless and arguably confusing for people new to the codebase.

```diff
- BattleStatus* battleStatus = BATTLE_STATUS;
+ BattleStatus* battleStatus = &gBattleStatus;
```

nanaian:
Closed in #189

PR #189, opened 2021-02-19
Odds and Ends

ethteck:
Some decomp
Got rid of many unnecessary macros (see #183)
Improved assist.py


include/functions.h
```diff
@@ -252,4 +256,10 @@ void func_80070B50(s32 arg0, f32 arg1, f32 arg2, f32 arg3, f32 arg4);
 void func_800718D0(s32 arg0, f32 arg1, f32 arg2, f32 arg3, f32 arg4, s32 arg5);
 void func_80070BB0(s32 arg0, f32 arg1, f32 arg2, f32 arg3, f32 arg4, s32 arg5);
 
+// Dead functions:
```
nanaian, line 259: these should probably definitely get their own header that the dead sections explicitly include

include/macros.h
```diff
@@ -23,12 +23,6 @@
 #define PANIC() ASSERT(0)
 #define STATIC_ASSERT(condition) enum { static_assert_fail = 1/(!!(condition)) } // Causes division by zero ("not integer constant") if false
 
-#define BATTLE_STATUS (&gBattleStatus)
-#define PLAYER_ACTOR (gBattleStatus.playerActor)
-#define GAME_STATUS (*gGameStatusPtr)
-#define PLAYER_STATUS (&gPlayerStatus)
-#define UI_STATUS (&gUIStatus)
-#define PLAYER_DATA (&gPlayerData)
```
nanaian, line 31: 🦀

include/variables.h
```diff
@@ -324,4 +325,7 @@ extern UNK_PTR D_800F7B94;
 
 extern s32 D_802C05CC;
 
+// Dead code
```
nanaian, line 328: ditto for dead

include/functions.h
nanaian: maybe prefix em with dead_ too? like `dead_atan2`

PR #194, opened 2021-02-23
jp cod (Created the first segment's c files for jp)

ethteck:
jp cod


configure.py
```diff
@@ -548,13 +555,19 @@ def add_generated_header(h: str):
...
-    for c_file in glob("src/**/*.c", recursive=True):
-        if c_file.endswith(".inc.c"):
-            continue
 
-        status = await shell_status(f"grep -q SCRIPT\( {c_file}")
+    for version in versions:
+        if version == "us":
+            c_files = glob("src/**/*.c", recursive=True)
+        elif version == "jp":
+            c_files = glob("ver/jp/src/**/*.c", recursive=True)
```
nanaian, line 563: are we just gonna not have jp load anything from src/ for now

configure.py
```diff
@@ -548,13 +555,19 @@ def add_generated_header(h: str):
         n.build("generated_headers_" + version, "phony", generated_headers)
         n.newline()
 
-    for c_file in glob("src/**/*.c", recursive=True):
-        if c_file.endswith(".inc.c"):
-            continue
 
-        status = await shell_status(f"grep -q SCRIPT\( {c_file}")
+    for version in versions:
+        if version == "us":
```
nanaian, line 560: this should probably be the `else` case

configure.py
```diff
@@ -548,13 +555,19 @@ def add_generated_header(h: str):
...
+            c_files = glob("src/**/*.c", recursive=True)
+        elif version == "jp":
+            c_files = glob("ver/jp/src/**/*.c", recursive=True)
+
+        for c_file in c_files:
+            if c_file.endswith(".inc.c"):
+                continue
+
+            status = await shell_status(f"grep -q SCRIPT\( {c_file}")
```
nanaian, line 569: this was outside of the version loop to avoid grepping the same file twice, i wonder if we could keep it that way and just add to c_files specifically for jp in a second loop? i can try and do this

configure.py
nanaian: (imo it should read from both dirs, and should be the case for all versions)

ethteck: yeah I don't think it's worth it cause splat isn't set up to work with it anyway

nanaian (approved):

PR #197, opened 2021-02-28
JP rom beginnings

ethteck:
Matched some jp stuff and started documenting the splat yaml for jp with my findings
Split an existing C file into two and moved some c files into os/ (thanks @JoshDuMan)
Also includes a splat update which prevents writing two labels when things are both jtbl labels and labels created by direct branches in the same file


nanaian (approved):
Looks good.
There's a `return (HeapNode*) - 1;` which has some weird spacing, but I'm it looks like you just moved that code in this PR so it's not really in-scope to fix

PR #199, opened 2021-03-08
Matched chill_out functions

tashi-lp:
Matched a few functions in `battle/star/chill_out_7900D0`


src/battle/star/chill_out_7900D0.c
```diff
@@ -2,20 +2,81 @@
...
+        *D_801512F0 = 1;
+        set_background_color_blend(0, 0, 0, 0);
+        script->functionTemp[0].s = 0;
+    }
+
+    set_background_color_blend(0, 0, 0, script->functionTemp[0].s);
+    script->functionTemp[0].s += 10;
+    temp = script->functionTemp[0].s;
+
+    if (temp >= 230) {
```
nanaian, line 54: Is `temp` required? Does `if (script->functionTemp[0].s >= 230)` not match?

src/battle/star/chill_out_7900D0.c
```diff
@@ -2,20 +2,81 @@
...
+
+    if (isInitialCall) {
+        script->functionTemp[0].s = 20;
+    }
+
+    set_background_color_blend(0, 0, 0, script->functionTemp[0].s * 10);
+    script->functionTemp[0].s -= 1;
+    temp = script->functionTemp[0].s;
+
+    if (!temp) {
```
nanaian, line 73: Same here - I'm not sure `temp` is required. Also, since this isn't a boolean (rather, its an int counting down), it's probably better to leave the condition as `if (script->functionTemp[0].s == 0)`.

nanaian (changes requested):
Thank you for the PR! I've made a couple comments about the `temp` variable in two of your functions - I'm not certain they're required. (Of course, if they _are_ feel free to leave them as-is!)

ethteck:
I approve btw once @nanaian 's changes are addressed. :)

PR #202, opened 2021-03-09
Misc decomp 3

ethteck:
Some data migration + decomp


include/enums.h
```diff
@@ -237,7 +237,7 @@ typedef UNK_TYPE Easing;
 #define Easing_SIN_OUT            0x0000000B
 #define Easing_COS_IN             0x0000000C
 
-typedef UNK_TYPE SoundId;
+typedef UNK_TYPE SoundID;
```
nanaian, line 240: Good, we should rename the below defines too that are in the `SoundId` namespace

nanaian (approved):
LGTM
exciting!

PR #208, opened 2021-03-12
Rodata migration + other goodz

ethteck:
Also decomped most of the star functions, where there were many duplicates
The gEffectTable address was wrong, so I fixed it and added a size parameter so it correctly disassembles the effect funcs
I'm sure there's other stuff I forgot. woohoo


.vscode/settings.json
```diff
@@ -21,6 +21,9 @@
         "docs/doxygen": true,
         "ctx.c": true,
         "ver/current": true,
+        "ver/us/build": true,
+        "ver/jp/build": true,
```
nanaian, line 25: Might be worth adding exceptions for the map files

src/battle/star/focus_789E60.c
```diff
@@ -1,25 +1,60 @@
...
+        script->varTable[0] = 1;
+        return ApiStatus_DONE2;
+    }
+
+    deepFocusSP = is_ability_active(Ability_DEEP_FOCUS) * 64;
+    superFocusSP = is_ability_active(Ability_SUPER_FOCUS) * 128;
+
+    add_SP(deepFocusSP + superFocusSP + 128);
+
+    if (playerData2->specialBarsFilled == playerData2->maxStarPower * 256) {
```
nanaian, line 35: Macro this condition and the other uses of it
Something like
```c
if ({
    PlayerData* playerData2 = &gPlayerData;
    playerData2->specialBarsFilled == playerData2->maxStarPower * 256;
}) {
    // ....
}
```

src/battle/star/lullaby_78D150.c
```diff
@@ -1,22 +1,41 @@
...
+    }
+
+    switch (script->functionTemp[0].s) {
+        case 0:
+            script->functionTemp[2].s = -80;
+            script->functionTemp[1].s = 0;
+            script->functionTemp[0].s = 1;
+        case 1:
+            temp = script->functionTemp[2].s;
+            func_800706D0(0, temp, (s32)((sin_rad((script->functionTemp[1].s * 6.28318f) / 360.0f) * 30.0f) + 30.0f), 50.0f);
```
nanaian, line 31: what is going on here, lol

src/battle/star/peach_beam_7997B0.c
```diff
@@ -1,26 +1,40 @@
...
+    if (script->functionTemp[0].s <= 200) {
+        set_background_color_blend(0, 0, 0, 200);
+        return ApiStatus_DONE2;
+    }
+
+    return ApiStatus_BLOCK;
+}
 
-INCLUDE_ASM(s32, "battle/star/peach_beam_7997B0", func_802A247C_79AC2C);
+// this file is identical in c code to star beam
```
nanaian, line 40: like, identical identical?
the scripts must be different, perhaps we should play about with includes then

src/battle/star/peach_dash_79D8A0.c
```diff
@@ -1,22 +1,35 @@
...
+    }
+    if (is_ability_active(Ability_SUPER_FOCUS)) {
+        playerData->specialBarsFilled += 256;
+    }
+
+    playerData->specialBarsFilled += 128;
+
+    if (playerData2->specialBarsFilled >= playerData2->maxStarPower * 256) {
+        playerData2->specialBarsFilled = playerData2->maxStarPower * 256;
+    }
```
nanaian, line 32: macro

src/battle/star/peach_beam_7997B0.c
ethteck: yeah

src/battle/star/lullaby_78D150.c
nanaian: may be wise to do this
```c
case 1: {
    s32 arg1 = script->functionTemp[2].s;
    f32 arg2 = sin_rad((script->functionTemp[1].s * 6.28318f) / 360.0f) * 30.0f) + 30.0f;
    func_800706D0(0, arg1, (s32) arg2, 50.0f);
}
```

.vscode/settings.json
ethteck: I think they should be filtered out by default, it just pollutes stuff when searching in my experience

src/battle/star/lullaby_78D150.c
ethteck: yeah

src/battle/star/focus_789E60.c
ethteck: Is it just the condition or the whole block? Or just like one of the fields? it's super unclear and I think it might be clearer if we wait and reconcile many examples where playerdata2 is needed

nanaian: just the condition, I'm fairly sure

nanaian (approved):

PR #215, opened 2021-03-19
Misc decomp

ethteck:
Did some decomp of various early-rom files. Also I renamed the override flags variable accordingly and added macros for it since I'm almost positive they exist at this point.


src/code_106EF0.c
```diff
@@ -29,32 +26,23 @@ s32 func_802E56D0(Entity* entity) {
...
-
-            entityIndex = create_entity(&D_802EA934, entity->position.x, entity->position.y, entity->position.z, 0.0f, 0x80000000);
-            temp_s0->unk_12 = entityIndex;
-            someEntity = get_entity_by_index(entityIndex);
-            temp2 = someEntity->dataBuf;
-            temp2->unk_00 = entity->listIndex;
-        }
+    struct802E3650* temp_s0 = entity->dataBuf;
+
+    if (temp_s0->unk_0A != 0xFFFF && get_global_flag(temp_s0->unk_0A) == 0) {
```
nanaian, line 32: The RHS of the flag check should probably be a bool FALSE or just use `!`

src/code_1b40_len_20b0.c
```diff
@@ -34,9 +35,155 @@ Gfx D_80074230[] = {
...
+        s32* overrideFlags = &gOverrideFlags;
+
+        if (*overrideFlags & 0x100) {
+            *overrideFlags |= 0x1000;
+        } else {
+            *overrideFlags &= ~0x1000;
+        }
+    }
+
+    {
```
nanaian, line 135: No way, Ethan using anonymous blocks!?

src/code_1b40_len_20b0.c
```diff
@@ -34,9 +35,155 @@ Gfx D_80074230[] = {
...
+    *gDisplayContextPtr = &dispContexts[gCurrentDisplayContextIndex];
+    *gfx = &(*gDisplayContextPtr)->backgroundGfx[0];
+
+    gfx_init_state();
+    gfx_draw_background();
+
+    gDPFullSync((*gfx)++);
+    gSPEndDisplayList((*gfx)++);
+
+    ASSERT((s32)((u32)((*gfx) - (*gDisplayContextPtr)->backgroundGfx) << 3 >> 3) < ARRAY_COUNT((*gDisplayContextPtr)->backgroundGfx));
```
nanaian, line 182: what

src/code_43F0.c
```diff
@@ -193,31 +218,27 @@ s32 func_80029994(s32 arg0) {
...
-    }
+    arg0 = abs(arg0);
 
-    ret = 0;
     if (arg0 != 0) {
-        if (arg0 == 1) {
-            return (func_80029994(1000) < 501) ^ 1;
+        switch (arg0) {
+            case 1:
+                ret = (func_80029994(1000) < 501) == 0;
```
nanaian, line 229: Use `!`?

src/code_43F0.c
```diff
@@ -242,7 +263,32 @@ s32 round(f32 arg0) {
...
+    }
+
+    do {
+        if (theta < 0.0f) {
+            theta += 360.0f;
+        }
+        if (theta >= 360.0f) {
+            theta -= 360.0f;
+        }
+    } while (!(theta >= 0.0f) || !(theta < 360.0f));
```
nanaian, line 288: Does this match if you remove the `!`s and use `&&` instead

src/code_1b40_len_20b0.c
```diff
@@ -34,9 +35,155 @@ Gfx D_80074230[] = {
...
+    step_current_game_mode();
+    update_entities();
+    func_80138198();
+    func_8014A548();
+    update_ambient_sounds();
+    func_80149734();
+    update_windows();
+    update_curtains();
+
+    overrideFlags = &gOverrideFlags;
```
nanaian, line 89: Given that you're using anonymous blocks in subsequent sections this should probably be in one too

nanaian (changes requested):

src/code_1b40_len_20b0.c
ethteck: https://discord.com/channels/279322074412089344/704241951972524063/822051304314241034

gonna leave a note but yeah this is a workaround to something that is not quite right

src/code_43F0.c
ethteck: Thanks, I re-looked at this and realized I can do

```
            case 1:
                ret = func_80029994(1000) > 500;
                break;
```
which I think is a lot better

src/code_43F0.c
ethteck: nary

nanaian (approved):

PR #218, opened 2021-03-21
All map data + rodata disasm

ethteck:
Some other decomp as well as splat update / fixes to yaml


src/world/common/UnkNpcAIFunc19.inc.c
```diff
@@ -0,0 +1,17 @@
+#include "common.h"
+#include "map.h"
+
+void N(UnkNpcAIFunc19)(ScriptInstance* script, NpcAISettings* aiSettings, s32 arg2) {
```
nanaian, line 4: This and the other related funcs should have a snake_case name

nanaian (changes requested):

PR #219, opened 2021-03-22
Refactor Jenkinsfile

ethteck:
All building is now done on non-master agents and the progress information is stashed and then unstashed from master, which updates the public files.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -33,11 +33,11 @@ pipeline {
             }
         }
         stage('Update Progress') {
-            agent {
-                label 'master'
-            }
             when {
                 branch 'master'
+            }
+            agent {
+                label 'master'
             }
             steps {
                 unstash 'reports'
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -28,11 +21,31 @@ pipeline {
                 branch 'master'
             }
             steps {
-                sh 'python3 progress.py us --csv >> /var/www/papermar.io/html/reports/progress_us.csv'
-                sh 'python3 progress.py us --shield-json > /var/www/papermar.io/html/reports/progress_us_shield.json'
+                sh 'mkdir reports'
```
nanaian, line 24: IDK if you've tested all this already, but shouldn't this be `mkdir -p reports` so the step doesn't fail if `reports/` already exists?

nanaian: It could be that stashed directories are never existing by default, I'm not a Jenkins expert :shipit:

ethteck: The workspace is fresh every time, and if it's not, I'm cool with this being an error (cause that's a bad sign)

PR #221, opened 2021-03-23
manual asm cleanup + some tiny src path fixes

Alto1772:
also `os/code_3CF80.c` is a duplicate of `os/controller.c` and the older one isnt used in the build
so i deleted it and moved asm `controller` to `os/controller` to match their src equivalent


ethteck:
Thanks! This is super helpful. I know we often have redundant asm files in the repo because we often get lazy when we move or split stuff. Appreciate the cleanup

nanaian (approved):

PR #222, opened 2021-03-23
Battle items

Kelebek1:


src/battle/item/dizzy_dial/dizzy_dial.c
```diff
@@ -0,0 +1,51 @@
+#include "dizzy_dial.h"
+
+Script N(main) = SCRIPT({
+    SI_VAR(10) = c 0x9A;
```
nanaian, line 4: This should be `= c ItemID_DizzyDial` I'm pretty sure.
Same for other items - obviously different item IDs, but they should exist in `enums.h`.

nanaian (changes requested):
Whew, this is a big PR - thank you for it! ✨ 

Just one small nit in each item's main script - `SI_VAR(10)` holds that item's ID, and preferably we'd use the enum name (e.g. `ItemID_Coconut`) there rather than a magic number.

(In the future it would be good to throw the duplicated code into an `.inc.c` file and include it from each item instead, but there's no need to do that in this PR :shipit:)

nanaian:
Looks like [https://github.com/pmret/papermario/pull/222/commits/daeaa6ce41d4e275c800206f504cf2f32c8f803c](daeaa6c "Volt Shroom") included running `format.sh`. It still says in CONTRBUTING.md to do this, but we've stopped doing it recently because of some inconsistencies between `astyle` versions. The changes it did make look fine, though, although it might be worth rebasing/force-pushing away changes outside of `src/battle/item/`. cc @ethteck

Kelebek1:
> 
> 
> Looks like [daeaa6c "Volt Shroom"](https://github.com/pmret/papermario/pull/222/commits/daeaa6ce41d4e275c800206f504cf2f32c8f803c) included running `format.sh`. It still says in CONTRBUTING.md to do this, but we've stopped doing it recently because of some inconsistencies between `astyle` versions. The changes it did make look fine, though, although it _might_ be worth rebasing/force-pushing away changes outside of `src/battle/item/`... cc @ethteck for opinion

How do I do that? I'm not good enough with git to figure out how to do it.

nanaian:
Don't worry about doing it for now, we'll see what @ethteck says

Kelebek1:
Is there a list anywhere of possible variables? Like I see "$damage"  for example, how/where is that defined?

src/battle/item/dried_shroom/lib.c
```diff
@@ -0,0 +1,175 @@
...
+    SetAnimation(ActorID_PLAYER, 0, PlayerAnim_RUNNING);
+    PlayerRunToGoal(0);
+    SetAnimation(ActorID_PLAYER, 0, PlayerAnim_2);
+    UseIdleAnimation(ActorID_PLAYER, 1);
+});
+
+Script N(EatItem) = SCRIPT({
+    spawn {
+        loop 4 {
+            PlaySoundAtActor(ActorID_PLAYER, 8341);
```
ethteck, line 158: We probably want the second argument to instances of PlaySoundAtActor to be in hex, assuming it's a sound ID, and we might even have the define for it

ethteck:
I'm cool with the formatting changes. After all, they are what we want basically. Overall I'm fine with this PR as-is, and I'll inc.c the duplicate funcs later on. Amazing work.

ethteck (approved):

nanaian:
> Is there a list anywhere of possible variables? Like I see "$damage"  for example, how/where is that defined?

These are just compiled into SI_VAR(0), SI_VAR(1), etc sequentially.
See the GitHub wiki page on the script DSL for more info.

nanaian (approved):

PR #223, opened 2021-03-24
Decompiled some functions in world.c

Mr-Wiseguy:
* get_asset_offset
* get_current_map_config
* get_current_map_header
* load_asset_by_name


Mr-Wiseguy:
There are some ROM address constants in the file that I didn't know what to name, so let me know and I'll update the PR with names for them.

ethteck (approved):

src/world/world.c
```diff
@@ -2,8 +2,23 @@
 #include "ld_addrs.h"
 #include "map.h"
 
+#define ASSET_TABLE_ROM_START 0x1E40000
```
nanaian, line 5: This should probably be the linker symbol for the assets segment.

src/world/world.c
```diff
@@ -24,15 +39,54 @@ static const s32 rodata_73DA0[] = {
 
 INCLUDE_ASM(s32, "world/world", load_map_by_IDs);
 
-INCLUDE_ASM(s32, "world/world", get_current_map_config);
+s32 get_current_map_config() {
+    return D_800A41E0;
+}
 
-INCLUDE_ASM(MapConfig*, "world/world", get_current_map_header);
+MapConfig *get_current_map_header() {
```
nanaian, line 46: Nit: our codestyle is to have the asterisk of pointer types hug the left, not the right. i.e. `s32*` not `s32 *`

nanaian (changes requested):

src/world/world.c
ethteck: He said it wouldn't match to use that since it's not doing addiu

src/world/world.c
ethteck: Oh yeah forgot that

src/world/world.c
Mr-Wiseguy: Yeah, lui + ori means it was a compile time constant because the mips lo16 and hi16 relocs are not compatible with that pair of instructions.

src/world/world.c
Mr-Wiseguy: Fixed.

src/world/world.c
nanaian: Gotcha, I didn't realise.
Perhaps its worth commenting that at some point as it could possibly confuse people once we get shiftability (or use the linker symbol when defined(NON_MATCHING)). No need to do that in this PR though!

nanaian (approved):

PR #224, opened 2021-03-24
Integrated kmc gcc for matching libgultra

Mr-Wiseguy:
Any source files in the src/ultra directory will be compiled using it.
Moved epiread and sptaskyield as examples.


Mr-Wiseguy:
Don't merge this PR yet, I swapped to a different version of mips ld to fix an issue with modern mips ld not behaving well with KMC gcc compiled object files. I will look into a better fix tomorrow.

Mr-Wiseguy:
I have resolved the ld issue by stripping the problematic symbols from the kmc gcc output files.

configure.py
```diff
@@ -561,14 +569,24 @@ def add_generated_header(h: str):
...
             s_glob = "ver/" + version + "/" + re.sub("src/", "asm/nonmatchings/", c_file)[:-2] + "/*.s"
-            n.build(
-                obj(c_file),
-                "cc_dsl" if status == 0 else "cc",
-                c_file,
-                implicit = None if not args.depend_on_s else glob(s_glob),
-                order_only="generated_headers_" + version,
-                variables={ "version": version }
-            )
+            if c_file.startswith("src/ultra/"):
```
ethteck, line 572: It'd be nice to isolate the change to this if/else and do the n.build afterward, using a variable that gets set.

So like if it's a file in src/ultra, set the rule to "cc_kmc", else if status == 0, set it to "cc_dsl", else "cc"

Mr-Wiseguy: Fixed

ver/us/splat.yaml
```diff
@@ -107,7 +107,7 @@ segments:
     - [0x3BD40, hasm, os/osSetIntMask]
     - [0x3bde0, c, os/osCreatePiManager]
     - [0x3C160, c, os/osEPiWriteIo] # MOVE_ADDU
-    - [0x3c1c0, c, os/osEPiReadIo] # MOVE_ADDU
+    - [0x3c1c0, c, ultra/epiread]
```
ethteck, line 110: I would say we might wanna keep my `# MOVE_ADDU` comment, but I guess everything in the ultra/ dir will eventually all implicitly have the addu for move anyway...

I was using those comments to track which files had the difference, but I guess we can just move them into ultra/ and the new compiler will generate the correct code? This will also get rid of the need for the different asm macro right?

Mr-Wiseguy: Yeah, there's no need for the macros in those files anymore. Anything compiled with kmc gcc will use addu for moves, and in fact the macros probably wouldn't even work with kmc gcc as it seems to not have as robust support for inline asm (for example, INCLUDE_ASM doesn't work with it).

ethteck (approved):

ethteck:
thanks so much! Excited to match libultra after this

nanaian (approved):
There are a couple correctness issues I forsee with the new ninja stuff but nothing that needs to be fixed right now. Thanks for the PR!!

ethteck:
Going to revert this for now since we're not actively working on libultra so much and the wine dependency is heavy and seems to be acting up weirdly sometimes. But we will definitely revisit this in the near future. :)

PR #225, opened 2021-03-24
Update battle item enums

Kelebek1:


src/battle/item/coconut/lib.c
```diff
@@ -112,7 +112,7 @@ Script N(UseItemWithEffect) = SCRIPT({
 
 Script N(UseItem) = SCRIPT({
     UseCamPreset(19);
-    SetBattleCamTarget(-85, 1, 0);
+    SetBattleCamTarget(0xFFFFFFAB, 1, 0);
```
ethteck, line 115: could you change this back? this is an x position apparently

PR #229, opened 2021-03-26
415D90 and Audio

ethteck:
415D90 data / rodata migration and some decomp work
Audio naming from clover and some more audio decomp


nanaian (approved):
Dang, seeing all these funcs and unknowns finally named is really cool

PR #236, opened 2021-03-28
world/area_arn/arn_02

Kelebek1:
Putting this up just as a test for now, to see what needs to be fixed. Looking for feedback on this before I go off and do a bunch more.


src/world/area_kmr/kmr_12/events.c
```diff
@@ -188,7 +188,7 @@ StaticNpc N(goombaNpc) = {
 };
 
 NpcGroupList N(npcGroupList) = {
-    NPC_GROUP(N(goombaNpc), BATTLE_ID(1, 0, 3)),
+    NPC_GROUP(N(goombaNpc), BATTLE_ID(0, 1, 0, 3)),
```
nanaian, line 191: Why did you add `unk` to the macro if you weren't going to use it? I've never seen the most significant byte be used it battle IDs.

src/world/common/UnkNpcAIFunc13_2.inc.c
```diff
@@ -14,13 +14,13 @@ void N(UnkNpcAIFunc13_2)(ScriptInstance* script, NpcAISettings* aiSettings, s32
     tempAngle = atan2(npc->pos.x, npc->pos.z, gPlayerStatusPtr->position.x, gPlayerStatusPtr->position.z);
     angleDiff = get_clamped_angle_diff(npc->yaw, tempAngle);
 
-    if (aiSettings->unk_10 < fabsf(angleDiff)) {
+    if (aiSettings->unk_10.s < fabsf(angleDiff)) {
```
nanaian, line 17: This has to be loaded as if it were an s32?? 🤢
I'm fairly certain it should always be f32, not X32... maybe try some casts

nanaian (changes requested):

nanaian:
I'm still unsure about (my) events/header.c pattern for naming files. A main.c might be better than header, and if you end up seeing a pattern feel free to give them better name

src/world/area_arn/arn_02/header.c
```diff
@@ -0,0 +1,30 @@
...
+    .main = N(Main),
+    .entryList = N(entryList),
+    .entryCount = ENTRY_COUNT(N(entryList)),
+    .background = &gBackgroundImage,
+    .tattle = MSG_arn_02_tattle,
+};
+
+Script N(PlayMusic) = SCRIPT({
+    match SI_SAVE_VAR(0) {
+        < 0xFFFFFFE3 {
```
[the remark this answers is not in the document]
nanaian: Yeah, they are.
These are story progress bytes, we actually have names for all of them somewhere; It's a big enum but with negative values so they can fit more than 255 states iirc.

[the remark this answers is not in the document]
nanaian: (by 'somewhere' I mean in a pinned comment on Discord or something)

src/world/area_kmr/kmr_12/events.c
Kelebek1: It is used in area_02. https://github.com/pmret/papermario/blob/b84a86d8eb95ca9cb4a85a5b31fa631e825fa897/src/world/area_arn/arn_02/events.c#L225

I had to add it for this, and then it threw an error for goomba due to not enough args.

nanaian: Ah, I missed this... no idea what it does; guess we should keep it anyways

src/world/common/UnkNpcAIFunc13_2.inc.c
Kelebek1: The problem is in this function:

https://github.com/pmret/papermario/blob/b84a86d8eb95ca9cb4a85a5b31fa631e825fa897/src/world/area_arn/arn_02/events.c#L269-L286

For `func_800490B4`, unk_10 is loaded and saved with `lw`, meaning that to be a float I have to change the arguments for the function to also be float. Doing that breaks pretty much every single include. I got through fixing a bunch of them but got bored, and went this "simpler" way instead. Look into `func_800490B4` if you want and let me know if it should be a flaot arg or not.

Kelebek1:
> 
> 
> I'm still unsure about (my) events/header.c pattern for naming files. A main.c might be better than header, and if you end up seeing a pattern feel free to give them better name

I was wondering what to do with the functions. From what I see from kmr_12, it has no actual functions, but this area does. Should I move the functions out somewhere on their own? I linked the header/events thing, but I'm not sure what should go within a "events" part, vs something else. I also thought about making a main, or a arn_02.c, but I don't know how the split would go.

If we follow the padding as file splits instead, as both arn_02 and kmr_12 has 2 paddings (I had 3 before making header.c), should we make up something for that?

ethteck (approved):

src/world/common/UnkNpcAIFunc13_2.inc.c
nanaian: `func_800490B4`'s assembly has this line in the prologue:
```s
/* 244C4 800490C4 4487C000 */  mtc1      $a3, $f24
```
So yes, that arg needs to be float... so we shouldn't be doing it the 'simpler' way

Kelebek1: Ok, I forgot why this didn't work, but now trying to fix it again I see.

So unk_10 is currently typed as `s32` in the repo. It's used here in that function so you say it must be float, however in the linked code you gave from `UnkNpcAIFunc13_2.inc.c`, it's required to load an s32, convert it to float and then do the checking. Here's the diff of that `if (aiSettings->unk_10 < fabsf(angleDiff)) {` line:

Untitled-1

If we keep it as s32 then there's an issue with the `func_800490B4` call again, as s32 forces a cvt.

nanaian: Yeah, you seem to be right there... X32 feels like the wrong abstraction to me but it's fine until we figure out (read: decomp) all these functions that use `NPCAISettings`.

nanaian: Feel free to leave this as-is in the PR

nanaian: (it's also possible this is purely an `s32` and was typed wrongly as a float? iirc all we've seen so far is `0.0f`, which is the same as `0`)

include/enums.h
```diff
@@ -4,6 +4,237 @@
...
+    STORY_CH7_BEGAN_PEACH_MISSION            =  0x58,
+    STORY_CH7_STAR_SPRIT_DEPARTED            =  0x59,
+
+    STORY_CH8_OPENED_PATH_TO_STAR_WAY        =  0x5A,
+    STORY_CH8_REACHED_STAR_HAVEN             =  0x5B,
+    STORY_CH8_STAR_SHIP_ACTIVATED            =  0x5C,
+    STORY_CH8_REACHED_BOWSERS_CASTLE         =  0x5E,
+    STORY_CH8_REACHED_PEACHS_CASTLE          =  0x5F,
+    STORY_EPILOGUE                           =  0x60,
+
```
nanaian, line 235: probably remove this empty line?

src/world/common/UnkNpcAIFunc13_2.inc.c
Kelebek1: Not sure how that could be, because I was showing that we need both s32 and f32 behaviour from it. If it was typed as  float we should see lwc1 everywhere and the screenshot above should be matching.

nanaian (approved):

PR #240, opened 2021-03-28
Some of code_a5dd0_len_114e0.c

nanaian:


ethteck:
I don't like single line ifs, but I wanna merge this so we can revisit it later

PR #243, opened 2021-03-29
Fix disasm script

Kelebek1:
Update disasm script to work with the new enum names. Add STORY_PROGRESS and WORLD_LOCATION defines in too.


nanaian (approved):
Looks great!

PR #251, opened 2021-04-02
area_arn

Kelebek1:


include/enums.h
```diff
@@ -603,7 +603,7 @@ enum Cams {
 
 typedef s16 ItemID;
 enum ItemIDs {
-    ITEM_NONE,
+    ITEM_NONE                   = 0,
```
nanaian, line 606: Why?

include/map.h
```diff
@@ -210,19 +210,35 @@ typedef struct EnemyTerritoryThing {
...
+typedef struct {
+    /* 0x00 */ Vec3i point;
+    /* 0x0C */ s32 wanderSizeX;
+    /* 0x10 */ s32 wanderSizeZ;
+    /* 0x14 */ s32 moveSpeedOverride;
+    /* 0x18 */ s32 unk_18; // 0 if WanderCircle, 1 otherwise
+    /* 0x1C */ Vec3i detect;
+    /* 0x28 */ s32 detectSizeX;
+    /* 0x2C */ s32 detectSizeZ;
+    /* 0x30 */ s32 unk_30; // 0 if DetectCircle, 1 otherwise
```
nanaian, line 222: Name this `detectShape` in both structs and use the enum as the type

include/map.h
```diff
@@ -210,19 +210,35 @@ typedef struct EnemyTerritoryThing {
...
-    /* 0x2C */ s32 sizeZ;
-    /* 0x30 */ s32 shape;
-    /* 0x34 */ s32 unk_34;
-} EnemyTerritory; // size = ???
+typedef struct {
+    /* 0x00 */ Vec3i point;
+    /* 0x0C */ s32 wanderSizeX;
+    /* 0x10 */ s32 wanderSizeZ;
+    /* 0x14 */ s32 moveSpeedOverride;
+    /* 0x18 */ s32 unk_18; // 0 if WanderCircle, 1 otherwise
```
nanaian, line 218: Name this `wanderShape` and give it the type `enum { SHAPE_CYLINDER, SHAPE_RECT }`

include/si.h
```diff
@@ -110,6 +110,7 @@ typedef enum ScriptOpcode {
 #define SI_ARRAY(v) ((v - 190000000))
 #define SI_ARRAY_FLAG(v) ((v - 210000000))
 #define SI_FIXED(v) (((v * 1024.0f) + -230000000)) // See float_to_fixed_var
+#define ARGS_END 0x80000000
```
nanaian, line 113: AFAIK this is only for MakeEntity, and we should probably name it as such. (`make_entity` (the internal non-API function MakeEntity wraps) is a real variable-argument fn using `va_list` and friends, which is a bit scary!)

src/battle/item/insecticide_herb/lib.c
```diff
@@ -75,20 +75,28 @@ ApiStatus N(func_802A1280_72A9D0)(ScriptInstance* script, s32 isInitialCall) {
     return ApiStatus_DONE2;
 }
 
+// This is a duplciate, search for others
```
nanaian, line 78: typo (duplciate)

src/code_EB9D90.c
```diff
@@ -7,9 +7,32 @@ INCLUDE_ASM(s32, "code_EB9D90", func_80240FB0_EB9E00);
...
+    return ApiStatus_DONE2;
+}
+*/
 
 INCLUDE_ASM(s32, "code_EB9D90", func_802413AC_EBA1FC);
 
 INCLUDE_ASM(s32, "code_EB9D90", func_802413CC_EBA21C);
 
-INCLUDE_ASM(s32, "code_EB9D90", func_802413EC_EBA23C);
+INCLUDE_ASM(s32, "code_EB9D90", func_802413EC_EBA23C);
```
nanaian, line 38: Please add a newline at EOF

nanaian (changes requested):

nanaian:
Some general notes as things to do ASAP before or after this is merged (cc @JoshDuMan, who may be interested in some of these):
1. Name the `boo`, `tubba_blubba`, etc. palettes and animations that are used in this area
2. Name all the messages used in this area, preferably we should never be using MESSAGE_ID and instead using named MSGs
3. Solve all the commented-out functions that Maide matched (cc @ethteck; this will also raise our % by a big chunk I expect)

PR #255, opened 2021-04-10
More image splits

JoshDuMan:
Know I'm gonna need to adapt feedback, opening now for it regardless

### DIFF
--- a/ver/us/splat.yaml
+++ b/ver/us/splat.yaml
@@ -2473,34 +2473,84 @@ segments:
     subsections:
     - [0x3F9E50, c]
     - [0x3FA480, bin]
-  - [0x3FA4B0, ci4,     star_cards/card_front, 32, 32]
+  - start: 0x3FA4B0
+    type: ci4
+    name: star_cards/card_front
+    flip: v
+    width: 32
+    height: 32
   - [0x3FA6B0, palette, star_cards/card_front]
   - [0x3FA6D0, bin] # PAD
-  - [0x3FA8B0, ci4,     star_cards/card_back, 32, 64] # FLIP
+  - start: 0x3FA8B0
+    type: ci4
+    name: star_cards/card_back
+    flip: v
+    width: 32
+    height: 64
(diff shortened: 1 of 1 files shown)


ver/us/splat.yaml
```diff
@@ -2485,6 +2527,19 @@ segments:
...
+  - [0x404F40, ci4,     world/cloud, 32, 32] # FLIP
+  - [0x405140, palette, world/cloud]
+  - [0x405340, ci4,     world/waterblock, 32, 32] # FLIP
+  - [0x405540, palette, world/waterblock]
+  - [0x405740, ci4,     world/yellow_carpet, 32, 32] # FLIP
+  - [0x405940, palette, world/yellow_carpet]
+  - [0x405B40, i4,     world/number_4, 32, 32] # FLIP ci4
+  - [0x405D40, i4,     world/number_3, 32, 32] # FLIP ci4
+  - [0x405F40, i4,     world/number_2, 32, 32] # FLIP ci4
+  - [0x406140, i4,     world/number_1, 32, 32] # FLIP ci4
```
nanaian, line 2539: what's going on here

JoshDuMan: They are ci4, but don't work as extracted ci4 since they share a palette.

nanaian: interesting, i should add that feature to splat

nanaian: I think these should all be one texture for now

nanaian:
For the segments marked FLIP just use the dictionary syntax instead of the shorthand list syntax

ver/us/splat.yaml
ethteck: I think I agree, but maybe leave the indivudal ones commented? Because my ideal goal would be to extract these separately probably so editing them is a tad nicer. Eventually we can have like a "spritesheet" group or something which stacks images like this together at build time but extracts them indidivudally for nicer viewing/editing

PR #259, opened 2021-04-15
Action segments

nanaian:


.gitignore
```diff
@@ -36,7 +36,6 @@ build/
 /include/message_ids.h
 /include/sprite/
 /include/map
-/tools/permuter_settings.toml
```
ethteck, line 39: Why was this removed?

src/77480.c
```diff
@@ -145,11 +145,15 @@ void func_800DFAAC(void) {
 
     check_input_midair_jump();
 
-    if (playerStatus->actionState != ACTION_STATE_SLIDING) {
+    if (playerStatus->actionState != ACTION_STATE_SLIDE) {
         collision_main_lateral();
         func_800E4508();
 
-        if ((collision_main_above() < 0) && (playerStatus->decorationList == 0) && (playerStatus->animFlags & 0x1000)) {
+        if (
```
ethteck, line 152: I don't really think this change is necessary. It'd be good to address automatic formatting soon so we don't have so much inconsistent formatting everywhere

src/a5dd0_len_114e0.c
```diff
@@ -41,6 +41,83 @@ INCLUDE_ASM(s32, "a5dd0_len_114e0", update_shadows);
 
 INCLUDE_ASM(s32, "a5dd0_len_114e0", set_entity_updatecmd);
 
+/*
+int step_entity_commandlist(entity *entity)
```
ethteck, line 45: This should probably be removed or nonmatchingd if it's close

.gitignore
ethteck: nm I guess the plan is to just include it in the repo now

ethteck (changes requested):

PR #260, opened 2021-04-15
3 down in 7B440 and 77480

ethteck:
also some non-matchings that are close


src/7B440.c
```diff
@@ -33,11 +46,99 @@ void update_player_input(void) {
...
+    playerStatus->unk_90 = 0.0f;
+    playerStatus->unk_94 = 0;
+    playerStatus->unk_98 = 0;
+    playerStatus->unk_9C = 0;
+
+    mapConfig = gAreas[gameStatus2->areaID].maps[gameStatus2->mapID].config;
+
+    if (mapConfig->entryList != NULL) {
+        if (gameStatus2->entryID < mapConfig->entryCount) {
+            playerStatus->position.x = (*mapConfig->entryList)[gameStatus2->entryID].x;
```
pixel-stuck, line 128: the GET_MAP_ENTRY() macro can probably be used here

src/7B440.c
```diff
@@ -71,6 +172,88 @@ void input_to_move_vector(f32* angle, f32* magnitude) {
...
+    if (temp1 == 0.0f) {
+        temp2 = playerStatus->targetYaw;
+    }
+
+    *arg0 = temp2;
+    *arg1 = temp1;
+}
 
 INCLUDE_ASM(s32, "7B440", func_800E24F8);
+// void func_800E24F8(void) {
```
nanaian, line 198: Throwing your comment on my PR back at ya: if this is close, wrap it in nonmatching 😉

src/7B440.c
ethteck: I mentioned this on discord but repeating here for clarity: I saw that this macro wasn't necessary in the one other place it's used, and I think it's not really giving us much and kind of hides what the code is doing, so I got rid of it

nanaian (approved):

PR #262, opened 2021-04-17
Configure updates for modding

nanaian:
.


tools/build/configure.py
```diff
@@ -59,10 +63,16 @@ def write_ninja_rules(ninja: ninja_syntax.Writer, cpp: str):
         command=f"{cross}objcopy $in $out -O binary && {BUILD_TOOLS}/rom/n64crc $out",
     )
 
-    ninja.rule("sha1sum",
-        description="check $in",
-        command=f"sha1sum -c $in && touch $out",
-    )
+    if DO_SHA1_CHECK:
```
ethteck, line 66: Why not just make sha1sum conditionally a dependency rather than changing the definition of the rule?

ethteck (changes requested):

tools/build/configure.py
nanaian: It would also mean changing the `default` rule (i.e. rules that depend on this); imo its easier to just make the check phony, as it means only changing things in one place

PR #270, opened 2021-04-18
Sprite stuff

nanaian:
Bottom 3 commits are the only unmerged ones


ethteck:
LGTM after the conflict is resolved. I am not sure why that line was removed in configure.py and if it should be there or not.

ethteck (approved):

PR #271, opened 2021-04-18
Don't reference symbols in map overlays in pause's data files

nanaian:
Pause data is an overlay that is placed in the same place that maps usually are (0x8024XXXX), so splat was getting confused.

Figured this stuff out by stripping map segments from the ROM and 'fixing' linker problems. Newer versions of splat might notice that these are in a totally different overlay and thus won't reference these bad symbols.


ethteck:
Since these get generated from splat, I think it makes more sense to try to tackle the underlying issue there. These will just be replaced every time we re-run disassembly, which is going to be inconvenient to deal with.

ethteck:
Closing on account of #277

PR #278, opened 2021-04-23
Compiler's Worst Nightmare

ethteck:
* `-fforce-addr` flag added, fixing tons of issues across the board
* `__naked__` attribute gone (we can use vanilla GCC 2.8.0/2.8.1)
* `Assorted funcs matched and some data migrated`
* some macros gone


nanaian (approved):
:shipit:

nanaian:
May be worth using vanilla gcc instead of gcc-papermario from now on

PR #291, opened 2021-05-02
disable_actor_blur

chansuke:


src/182B30.c
```diff
@@ -27,7 +27,20 @@ void mtx_mirror_y(Matrix4f arg0) {
 
 INCLUDE_ASM(s32, "182B30", enable_actor_blur);
 
-INCLUDE_ASM(s32, "182B30", disable_actor_blur);
+void disable_actor_blur(Actor* actor) {
+    ActorPart* part = actor->partsTable;
+
+    if ((part->idleAnimations != 0) && !(part->flags & 2)) {
```
nanaian, line 33: Perhaps change `(part->idleAnimations != 0)` to `part->idleAnimations != NULL` since it's a pointer (I think?).

nanaian (changes requested):

src/182B30.c
chansuke: @nanaian Thank you for the comment. I will fix this.

nanaian (approved):
Looks good; thanks for the PR!

PR #310, opened 2021-07-13
Match dma_load_string

nanaian:


src/B9D60.c
```diff
@@ -26,7 +28,18 @@ INCLUDE_ASM(s32, "B9D60", msg_copy_to_print_buffer);
...
 INCLUDE_ASM(s32, "B9D60", initialize_printer);
 
-INCLUDE_ASM(s32, "B9D60", dma_load_string);
+void dma_load_string(u32 msgID, void* dest) {
+    u8* addr = (u8*) MSG_ROM_START + (msgID >> 14); // (msgID >> 16) * 4
+    u8* offset[2]; // start, end
+
+    dma_copy(addr, addr + 4, &offset[0]); // Load section offset
+
+    addr = MSG_ROM_START + offset[0] + (msgID & 0xFFFF) * 4;
```
ethteck, line 37: the & 0xFFFF is a u16 cast, and it probably means that the type of the first arg is u16. If not, you can at least do an explicit cast

nanaian: its not a cast

nanaian: It's getting the index out of the u32 message id

ethteck: ah ok nm

PR #312, opened 2021-07-15
Split charsets, matching load_font and load_font_data

nanaian:
Depends on https://github.com/ethteck/splat/pull/88


nanaian:
Should also consider renaming load_font and load_font_data to use 'charset' instead of font; fonts are a charset plus character width data - there are 3 charsets, but 4 fonts.

PR #315, opened 2021-07-16
Message work

ethteck:
Also, new compiler, -fno-common compiler option, and lots of fixes to headers*

Note to contributors: remember not to declare struct defintions with out "typedef" in front, and please don't declare variables in headers


include/common_structs.h
```diff
@@ -50,6 +55,11 @@ typedef struct Vec2s {
     /* 0x02 */ s16 y;
 } Vec2s; // size = 0x04
 
+typedef struct Vec2su {
```
nanaian, line 58: I'm starting to not like our naming system for these vec structs... we use `u16` for short unsigned but then "su" here...

Could we consider a Vec2(T) and Vec3(T) macro (`#define Vec2(T) struct { T x; T y; }`) instead? i.e. `Vec2(u16) pos = { 0, 0 };` (not screaming caps VEC2 because they're effectively type templates)

src/world/area_mac/mac_00/7EAD10.c
```diff
@@ -5,6 +5,7 @@ static char* N(exit_str_1) = "mac_01";
 static char* N(exit_str_2) = "tik_19";
 static char* N(exit_str_3) = "kmr_20";
 static char* N(exit_str_4) = "mac_00";
+static char* N(exit_str_5) = "\0\0\0\0";
```
nanaian, line 8: Why is this required?

ethteck: because of some strange rodata padding issue. we'll probably figure it out when we actually do these files

include/common_structs.h
ethteck: that seems overcomplicated to me, personally, and I could see it making figuring out the actual types and sizes of things confusing. happy to revisit the naming convention for the vec structs though. we already have one Vec(n)(t)u so I just copied that convention for this one

nanaian: We could also typedef them as Vec2u16 to avoid the macro

PR #316, opened 2021-07-17
Match func_8003DFA0 and func_8003E1D0

ThreeMustyFears:


src/npc.c
```diff
@@ -13,6 +13,12 @@ extern s16 D_80077C1C;
 extern s16 D_80077C1E;
 extern s32 D_80077C20;
 
+s32 playFX_0C(f32, f32, f32, f32, f32); // extern
+extern u16 D_80077C30;
+extern s32 D_80077C34;
+s32 playFX_23(u32, f32, f32, f32, f32); // extern
```
nanaian, line 19: I would move the prototypes for these functions to `include/functions.h` for now. There's a big section on line 390 marked 'Effect funcs' where they can live.

src/npc.c
```diff
@@ -1331,11 +1337,42 @@ INCLUDE_ASM(void, "npc", func_8003DA38, Npc* npc, s32 arg1);
...
 INCLUDE_ASM(s32, "npc", func_8003DC38);
 
-INCLUDE_ASM(s32, "npc", func_8003DFA0);
+void func_8003DFA0(Npc* npc) {
+    f32 temp_f20;
+    f32 subroutine_argA;
+    f32 subroutine_argB;
+    u32 phi_v0;
+
+    phi_v0 = (s32) (s16) (D_80077C30++) < 4;
```
nanaian, line 1346: You can remove the cast to `s32` here, its unneeded to match.

src/npc.c
```diff
@@ -1331,11 +1337,42 @@ INCLUDE_ASM(void, "npc", func_8003DA38, Npc* npc, s32 arg1);
...
+    if (phi_v0 == 0) {
```
nanaian, line 1347: Perhaps compare against the macro `FALSE` instead of `0` here, since we know phi_v0 is a boolean.

src/npc.c
```diff
@@ -1331,11 +1337,42 @@ INCLUDE_ASM(void, "npc", func_8003DA38, Npc* npc, s32 arg1);
...
+        D_80077C30 = 0U;
+        temp_f20 = (clamp_angle(-npc->yaw) * TAU) / 360.0f;
+        subroutine_argA = sin_rad(temp_f20);
```
nanaian, line 1350: You can name this var `x` or something; it's the x component of the NPC's facing direction (`npc-yaw`). Ditto for `subroutine_argB` but called `z`.

src/npc.c
```diff
@@ -1331,11 +1337,42 @@ INCLUDE_ASM(void, "npc", func_8003DA38, Npc* npc, s32 arg1);
...
+        subroutine_argB = cos_rad(temp_f20);
+        playFX_0C(npc->pos.x + (npc->collisionRadius * subroutine_argA * 0.2f), npc->pos.y + 1.5f, npc->pos.z + (npc->collisionRadius * subroutine_argB * 0.2f), -npc->yaw, (f32) D_80077C34);
```
nanaian, line 1352: The cast of `D_80077C34` to `f32` is implicit because of the function prototype, so it can be removed

src/npc.c
```diff
@@ -1331,11 +1337,42 @@ INCLUDE_ASM(void, "npc", func_8003DA38, Npc* npc, s32 arg1);
...
+        phi_v0 = D_80077C34 == 0;
+        D_80077C34 = phi_v0;
```
nanaian, line 1354: `phi_v0` doesn't need to be a temp; `D_80077C34 = D_80077C34 == 0` matches by itself.

nanaian (changes requested):
Thank you for the contributions!!!

I've made a few notes on func_8003DFA0 to clean it up a little. If you could fix those nits and also apply similar changes to func_8003E1D0 where applicable it'd be awesome :shipit:

nanaian (approved):
LGTM 🎉

ethteck:
I hope you don't mind I pushed a commit that cleans things up a bit more and removes the temp for the condition at the beginning of the functions. Thanks for your contribution, and feel free to connect with us on Discord! :)

PR #328, opened 2021-07-20
Various funcs in 77480, 1A5830, 1776B0

ethteck:


src/197F40.c
```diff
@@ -1224,15 +1224,15 @@ ApiStatus SetActorRotation(ScriptInstance* script, s32 isInitialCall) {
 
     actor = get_actor(actorID);
 
-    if (x != -250000000) {
+    if (x != SI_LIMIT) {
```
nanaian, line 1227: Clover calls this the 'ignore' value but it shouldn't(?) be a macro as it changes between different functions (e.g. SetGoalPos uses 0xFF439EB2 as the ignore value whereas this one is 0xF1194D80)

We could call them SI_IGNORE_1 and SI_IGNORE_2 etc

nanaian (changes requested):

nanaian (approved):

PR #329, opened 2021-07-20
add remaining opcode support to script DSL

nanaian:
- update dsl caserange and bind syntax
- add bind_padlock
- use dsl buf and arr commands
- clean up foliage.inc.c
- disasm rest of battle/partner/goombario scripts
- fix rest of SI_CMD uses


tools/build/configure.py
```diff
@@ -322,7 +322,7 @@ def build(object_paths: Union[Path, List[Path]], src_paths: List[Path], task: st
                 else:
                     with entry.src_paths[0].open() as f:
                         s = f.read()
-                        if "SCRIPT(" in s or "#pragma SCRIPT" in s:
+                        if "SCRIPT(" in s or "#pragma SCRIPT" or "#include \"world/common/foliage.inc.c\"" in s:
```
ethteck, line 325: this is unfortunate

ethteck: by "this" I mean needing to add an edge case, but I agree with doing it for now

PR #330, opened 2021-07-20
Miscellaneous battle/area functions

Brotenko:
grafik


src/common/UnkBattleFunc5.inc.c
```diff
@@ -0,0 +1,15 @@
+#include "common.h"
+
+ApiStatus N(UnkBattleFunc5)(ScriptInstance* script, s32 isInitialCall) {
+    Bytecode* args = script->ptrReadPos;
+    s32 actorID = get_variable(script, *args++);
+    s32 partID = get_variable(script, *args++);
+
+    if (actorID == -127) {
```
ethteck, line 8: If you look at the actor ID enum, I think this has a define for it.

src/battle/area_flo/5B28D0.c
```diff
@@ -2,19 +2,128 @@
...
+extern char* D_80231FA8_5CB458[];
+extern char* D_80232428_5CB8D8[];
 
-INCLUDE_ASM(s32, "battle/area_flo/5B28D0", func_802194E0_5B2990);
+ApiStatus func_80219420_5B28D0(ScriptInstance* script, s32 isInitialCall) {
+    s32 var1 = 0;
+    u32 counter;
 
-INCLUDE_ASM(s32, "battle/area_flo/5B28D0", func_802195A0_5B2A50);
+    for (counter = 0; counter < 11; counter++) {
```
ethteck, line 20: This applies to all instances of this function:

* `counter` should probably just be `i`
* for 11, you can do `ARRAY_COUNT(D_802311D8_5CA688)` for example. You'll need to specify the size of each array in the extern as well.
* 0 and 1 should probably be FALSE and TRUE since it looks like tihs function is passing a bool to fx 82

ethteck (changes requested):
🎊

src/battle/area_flo/5B28D0.c
nanaian: Agree with what ethan said except perhaps maybe the last bullet point; we have no idea what kinds of args these fx funcs take yet

src/battle/partner/goombario.c
```diff
@@ -122,11 +137,49 @@ ApiStatus func_80238B60_6F1C40(ScriptInstance* script, s32 isInitialCall) {
...
+            }
+            break;
+        case 135:
+            if (partnerActor->isGlowing >= 99) {
+                script->varTable[0] = 1;
+            }
+            break;
+        case 136:
+            if (partnerActor->isGlowing >= 99) {
+                script->varTable[0] = 1;
```
nanaian, line 174: Use TRUE and FALSE for this function, it's like an `is glowing?` routine

src/common/UnkAreaFunc.inc.c
```diff
@@ -0,0 +1,11 @@
+#include "common.h"
+
+ApiStatus N(UnkAreaFunc)(ScriptInstance* script, s32 isInitialCall) {
```
nanaian, line 3: You can name this and the file `GetModelPos` or something, since that's what it obviously does

src/common/UnkBattleFunc3.inc.c
```diff
@@ -0,0 +1,10 @@
+#include "common.h"
+
+ApiStatus N(UnkBattleFunc3)(ScriptInstance* script, s32 isInitialCall) {
```
nanaian, line 3: Same here; call it `GetBootsHammerLevel` or something

src/common/UnkBattleFunc4.inc.c
```diff
@@ -0,0 +1,10 @@
+#include "common.h"
+
+ApiStatus N(UnkBattleFunc4)(ScriptInstance* script, s32 isInitialCall) {
```
nanaian, line 3: `GetJumpHammerCharge`

src/common/UnkBattleFunc5.inc.c
nanaian: ACTOR_SELF

nanaian (changes requested):

src/battle/area_flo/5B28D0.c
ethteck: yeah fair enough, in the context of the effect func it might not be a bool

PR #337, opened 2021-07-21
Decomp btl_are_all_enemies_defeated

Nalfein16:


src/190B20.c
```diff
@@ -1072,7 +1072,30 @@ INCLUDE_ASM(s32, "190B20", func_80263268);
...
+
+        // If currentEnemyFlags signify that the enemy isn't dead yet...
+        if (currentEnemyFlags != 0) {
+            if ((*currentEnemyFlags & flagEnemyDefeated) == 0) {
+                // Countinue the battle
+                enemiesStillAlive = TRUE;
+            }
+        }
+    }
+    enemiesStillAlive ^= 1; // if enemiesStillAlive = 0, then enemiesStillAlive will be TRUE
```
nanaian, line 1096: Perhaps try `return !enemiesStillAlive;`

src/190B20.c
```diff
@@ -1072,7 +1072,30 @@ INCLUDE_ASM(s32, "190B20", func_80263268);
...
+
+    enemiesStillAlive = FALSE;
+    for (i = 0; i < ARRAY_COUNT(bs->enemyActors); i++) {
+        flagEnemyDefeated = 0x404000;
+        bs = &gBattleStatus;
+        currentEnemyFlags = bs->enemyActors[i];
+
+        // If currentEnemyFlags signify that the enemy isn't dead yet...
+        if (currentEnemyFlags != 0) {
+            if ((*currentEnemyFlags & flagEnemyDefeated) == 0) {
```
nanaian, line 1090: What's up with the dereference of `currentEnemyFlags` here? It's not a pointer...
Also we write falsey flag checks like `if (!(flags & flag))`

src/190B20.c
```diff
@@ -1072,7 +1072,30 @@ INCLUDE_ASM(s32, "190B20", func_80263268);
...
+    s32 flagEnemyDefeated;
+
+    enemiesStillAlive = FALSE;
+    for (i = 0; i < ARRAY_COUNT(bs->enemyActors); i++) {
+        flagEnemyDefeated = 0x404000;
+        bs = &gBattleStatus;
+        currentEnemyFlags = bs->enemyActors[i];
+
+        // If currentEnemyFlags signify that the enemy isn't dead yet...
+        if (currentEnemyFlags != 0) {
```
nanaian, line 1089: Compare to NULL instead of 0 since its a pointer

src/190B20.c
```diff
@@ -1072,7 +1072,30 @@ INCLUDE_ASM(s32, "190B20", func_80263268);
 
 INCLUDE_ASM(s32, "190B20", func_80263300);
 
-INCLUDE_ASM(s32, "190B20", btl_are_all_enemies_defeated);
+s32 btl_are_all_enemies_defeated(void) {
+    s32* currentEnemyFlags;
+    s32 i;
+    BattleStatus* bs;
```
ethteck, line 1078: Super minor things, but 

* we name battlestatus temps `battleStatus` - we may want to shorten that at some point but it's probably best to be consistent for now.

* Since that for loop is toward the top of the function, could you move the i decl to be the last one? We do that so it's as close as can be (since we can't declare it in the for loop definition, grumble grumble)

* could you try intializing the battle status temp at the beginning of the function? It may work, but it may not. no problem if it doesn't. 

* it probably makes sense to initialize enemiesStillAlive on the same line it's declared, just to save space. 

None of these are huge deals, but they're just things we typically do style-wise. Great job on this!

nanaian (approved):
lgtm 🚀

PR #342, opened 2021-07-24
Match func_800397E8

ThreeMustyFears:


src/npc.c
```diff
@@ -358,7 +358,45 @@ INCLUDE_ASM(s32, "npc", npc_do_player_collision, Npc* npc);
...
+
+        if (npc->flags & 0x800) {
+            npc->flags &= ~0x1000;
+            return 0;
+        }
+
+        temp_v1 = fabsf(arg1)+16;
+        subroutine_arg = temp_v1;
+        x = npc->pos.x;
+        y = npc->pos.y+13;
```
JoshDuMan, line 379:
```suggestion
        y = npc->pos.y + 13;
```

src/npc.c
```diff
@@ -358,7 +358,45 @@ INCLUDE_ASM(s32, "npc", npc_do_player_collision, Npc* npc);
 
 INCLUDE_ASM(void, "npc", npc_do_gravity, Npc* npc);
 
-INCLUDE_ASM(s32, "npc", func_800397E8);
+s32 func_800397E8(Npc* npc, f32 arg1) {
+    if (!(npc->flags & 0x208))
+    {
```
ethteck, line 363: We don't use the allman brace style here - could you put this on the same line as the if?

src/npc.c
```diff
@@ -358,7 +358,45 @@ INCLUDE_ASM(s32, "npc", npc_do_player_collision, Npc* npc);
...
+        z = npc->pos.z;
+
+        if (!(npc->flags & 0x4000000)) {
+            phi_v0 = npc_raycast_down_sides(npc->unk_80, &x, &y, &z, &subroutine_arg);
+        } else {
+            phi_v0 = npc_raycast_down_ahead(npc->unk_80, &x, &y, &z, &subroutine_arg, npc->yaw,
+                                            npc->collisionRadius);
+        }
+
+        if (phi_v0 && subroutine_arg <= temp_v1) {
```
ethteck, line 389: Because phi_v0 is a number here, rather than a pointer, this should be `phi_v0 != 0`

src/npc.c
```diff
@@ -358,7 +358,45 @@ INCLUDE_ASM(s32, "npc", npc_do_player_collision, Npc* npc);
...
+            npc->pos.y = y;
+            npc->unk_84 = D_8010C97A;
+            npc->flags |= 0x1000;
+            return 1;
+        }
+    } else return 0;
```
ethteck, line 395:
```
else {
    return 0;
}
```

ethteck (changes requested):
Nice job! Some style comments

nanaian (changes requested):
Looks good, thanks for the match!
It would be nice to use [the NPC flag enum](https://github.com/pmret/papermario/blob/ed9727a34e8a1533d7bd5cc01c2e71dbef2236a9/include/enums.h#L1556) for the flag checks/sets in this function if possible. Even for those flags that we haven't named yet it's helpful to parameterise them so they can be reverse-engineered in the future :)

ThreeMustyFears:
For the flags that haven't yet been named should I add them to enum.h, or is there some other way of doing this.

ethteck:
I spoke with @nanaian, and we agreed that adding flags isn't totally necesesary at this step. What he said is ideal, but yeah don't worry too much about it. For future reference, yeah, you can add them to an existing enum in enums.h as like `WHATEVER_FLAG_10000` for example

PR #348, opened 2021-07-31
battle/action_cmd

nanaian:


ethteck:
`shink`

ethteck:
gg! Love the ac names. I'm not sure I like `cmd` as part of the directory name but we can deal with that later if it gets to become a thing

nanaian:
Yeah me neither, although "action_command" felt a little long

PR #352, opened 2021-08-02
Match world/action/jump funcs

ThreeMustyFears:


include/enums.h
```diff
@@ -1606,6 +1607,7 @@ enum NpcFlags {
 enum PlayerAnimFlags {
     PLAYER_ANIM_FLAG_HOLDING_ITEM    = 0x00000001,
     PLAYER_ANIM_FLAG_2               = 0x00000002,
+    PLAYER_ANIM_FLAG_3               = 0x00000003,
```
ethteck, line 1610: since this is a flag, it's actually 2 | 1, so instead of a "3" flag you would just do `(PLAYER_ANIM_FLAG_HOLDING_ITEM | PLAYER_ANIM_FLAG_2)`

src/world/action/jump.c
```diff
@@ -1,15 +1,250 @@
 #include "common.h"
 
-INCLUDE_ASM(void, "world/action/jump", func_802B6000_E24040, void);
+extern void phys_adjust_cam_on_landing();
```
ethteck, line 3: Function declarations don't need `extern` before them, and they can actually move to functions.h provided these functions live in other files.

src/world/action/jump.c
```diff
@@ -1,15 +1,250 @@
...
+    playerStatus->unk_3C = playerStatus->position.x;
+    playerStatus->unk_40 = playerStatus->position.z;
+    playerStatus->unk_4C = playerStatus->position.y;
 
-INCLUDE_ASM(void, "world/action/jump", func_802B6348_E24388, void);
+    phys_init_integrator_for_current_state();
+    collisionStatus = &gCollisionStatus;
 
-INCLUDE_ASM(void, "world/action/jump", func_802B647C_E244BC, void);
+    if ((playerStatus->animFlags & PLAYER_ANIM_FLAG_8BIT_MARIO)) {
```
ethteck, line 31: Unnecessary double parens here

src/world/action/jump.c
```diff
@@ -1,15 +1,250 @@
...
+        temp_v1 = 0x90005;
+    }
+    else {
+        phi_a0 = (playerStatus->animFlags & PLAYER_ANIM_FLAG_3);
```
ethteck, line 35: This can be flag 2 bitwise ored with flag 1 as mentioned above. Also, I don't think you need this temp and you can just put the expression into the if condition. You may have to invert the condition for it to match

src/world/action/jump.c
```diff
@@ -1,15 +1,250 @@
...
 
-INCLUDE_ASM(void, "world/action/jump", func_802B60B4_E240F4, void);
+extern f32 D_8010C960;
+extern f32 D_8010C97C;
 
-INCLUDE_ASM(void, "world/action/jump", func_802B6198_E241D8, void);
+void func_802B6000_E24040() {
+    s32 temp_v1;
+    s32 phi_a0;
+    PlayerStatus* playerStatus = &gPlayerStatus;
```
ethteck, line 16: We typically put the temps for globals before everything else in the function. So the playerStatus and collisionStatus can go before the temp / phi. Also, you may be able to initialize collisionStatus when you declare it as well.

src/world/action/jump.c
```diff
@@ -1,15 +1,250 @@
...
 
-INCLUDE_ASM(void, "world/action/jump", func_802B6508_E24548, void);
+void func_802B60B4_E240F4() {
+    s32 temp_v1;
+    s32 phi_a0;
+    PlayerStatus* playerStatus = &gPlayerStatus;
+    Camera* cam;
+
+    if (playerStatus->flags < 0) {
+        playerStatus->flags &= 0x7FFFFFFF;
```
ethteck, line 53: 0x7FFFFFFF is the same as ~0x1, and we prefer to write the flags this way (this unsets flag 1)

src/world/action/jump.c
```diff
@@ -1,15 +1,250 @@
...
+    }
+
+    playerStatus->decorationList++;
+    phys_update_interact_collider();
+    sub_arg1 = &sp28;
+    sub_arg2 = &sp2C;
+    sub_arg3 = &sp30;
+    sub_arg4 = &sp34;
+    sub_arg6 = sub_arg6;
+    sp28 = playerStatus->position.x;
```
ethteck, line 198: I don't think any of the sub_args are needed here and you cna just do &sp28, &sp2C, etc as arguments in the function call below

src/world/action/jump.c
```diff
@@ -1,15 +1,250 @@
...
+          phi_a0 = 0x6000A;
+
+          if (!(playerStatus->animFlags & PLAYER_ANIM_FLAG_3)) {
+            phi_a0 = 0x10008;
+          }
+
+          temp_s1 = phi_a0;
+          suggest_player_anim_clearUnkFlag(temp_s1);
+          cam = &gCameras;
+          cam->moveFlags |= 1;
```
ethteck, line 122: Are you sure this camera temp is needed? You should be able to do `gCameras[0].moveFlags`. If that doesn't work, the value of cam should actually be `&gCameras[0]` since it's a pointer to the first camera in the array, rather than a pointer to the array itself.

ethteck (changes requested):
Great job with these! There are some style things that I commented on, many of which apply to all of the functions and not just the ones that I mentioned. If you have questions, feel free to reach out here or on discord. Some suggestions may not match, in which case you can disregard those and I'll take a look later.

Brotenko:
It seems like Jenkins didn't check for .s files being deleted for some reason. Nevertheless, you should run `coverage.py` before opening a PR to remove the decompiled assembly functions!
Besides that a solid PR my man!

For @ethteck - What happened here?:
grafik

src/world/action/jump.c
```diff
@@ -1,15 +1,250 @@
...
+    if (playerStatus->flags < 0) {
+        playerStatus->flags &= 0x7FFFFFFF;
+        playerStatus->flags &= ~0xA;
+        playerStatus->flags |= 4;
+        temp_v1 = 0xA0006;
+
+        if (!(playerStatus->animFlags & PLAYER_ANIM_FLAG_PEACH_PHYSICS)) {
+            temp_v1 = 0x10008;
+        }
+        suggest_player_anim_clearUnkFlag(temp_v1);
```
Kelebek1, line 226:
```suggestion
        if (playerStatus->animFlags & PLAYER_ANIM_FLAG_PEACH_PHYSICS) {
            suggest_player_anim_clearUnkFlag(0xA0006);
        } else {
            suggest_player_anim_clearUnkFlag(0x10008);
       }
```

GCC likes to break up multiple of the same function call in different branches like this. It just pulls out the arguments and then calls the function at the end. Try play around with the branching here and you can get rid of a bunch of these temps you're using, this applies to all the `suggest_player_anim_clearUnkFlag` here, and probably more functions besides.

src/world/action/jump.c
ThreeMustyFears: I'm attempting to makes this change but if I simply substitute ~0x1 in it switches a lui and ori to li

ethteck: Oh wait, is it ~0x80000000? I may have gotten screwed up... if that doesn't work, don't worry about it

src/world/action/jump.c
nanaian: There are further double paren instances elsewhere too that can be fixed e.g. in func_802B60B4_E240F4

src/world/action/jump.c
```diff
@@ -1,15 +1,250 @@
...
+    sub_arg6 = sub_arg6;
+    sp28 = playerStatus->position.x;
+    sp2C = playerStatus->position.y;
+    sp30 = playerStatus->position.z;
+    sp34 = playerStatus->colliderHeight;
+
+    if (((get_collider_type_by_id(player_raycast_below_cam_relative(playerStatus, sub_arg1, sub_arg2, sub_arg3, sub_arg4,
+                                    &sub_arg5, &sub_arg6, &sub_arg7, &sub_arg8)) & 0xFF) - 2 >= 2u) && check_input_jump()) {
+
+        set_action_state(3);
```
nanaian, line 206:
```suggestion
        set_action_state(ACTION_STATE_JUMP);
```

src/world/action/jump.c
```diff
@@ -1,15 +1,250 @@
...
+    sub_arg3 = &sp30;
+    sub_arg4 = &sp34;
+    sub_arg6 = sub_arg6;
+    sp28 = playerStatus->position.x;
+    sp2C = playerStatus->position.y;
+    sp30 = playerStatus->position.z;
+    sp34 = playerStatus->colliderHeight;
+
+    if (((get_collider_type_by_id(player_raycast_below_cam_relative(playerStatus, sub_arg1, sub_arg2, sub_arg3, sub_arg4,
+                                    &sub_arg5, &sub_arg6, &sub_arg7, &sub_arg8)) & 0xFF) - 2 >= 2u) && check_input_jump()) {
```
nanaian, line 204: Usually unsigned-literal comparisons like these are the result of gcc merging `||` operands into a more efficient instruction; not sure what `k - 2 >= 2U` would be but perhaps @ethteck has ideas

src/world/action/jump.c
```diff
@@ -1,15 +1,250 @@
...
+
+void func_802B6508_E24548() {
+    PlayerStatus* playerStatus = &gPlayerStatus;
+    Camera* cam;
+
+    if (playerStatus->flags < 0) {
+        playerStatus->flags &= 0x7FFFFFFF;
+        playerStatus->flags &= ~0xA;
+        playerStatus->flags |= 4;
+        cam = &gCameras;
```
nanaian, line 241:
```suggestion
        cam = &gCameras[0];
```

nanaian (changes requested):
Not much to add ontop of the existing reviews, but very cool

src/world/action/jump.c
ethteck: I've never done one of these with a greater than ...huh. It's probably an or of some sort, but it'd require a temp and I'm not sure.......

src/world/action/jump.c
```diff
@@ -1,15 +1,220 @@
...
+}
 
-INCLUDE_ASM(void, "world/action/jump", func_802B647C_E244BC, void);
+void func_802B60B4_E240F4() {
+    PlayerStatus* playerStatus = &gPlayerStatus;
+    s32 phi_a0;
 
-INCLUDE_ASM(void, "world/action/jump", func_802B6508_E24548, void);
+    if (playerStatus->flags < 0) {
+        playerStatus->flags &= (~0x80000000);
```
ethteck, line 43: No parens are necessary around the constant here

src/world/action/jump.c
```diff
@@ -1,15 +1,220 @@
...
+          playerStatus->flags &= 0x7FFFFFF5;
+          playerStatus->flags |= 4;
+
+          phi_a0 = 0x6000A;
+
+          if (!(playerStatus->animFlags & (PLAYER_ANIM_FLAG_HOLDING_ITEM | PLAYER_ANIM_FLAG_2))) {
+            phi_a0 = 0x10008;
+          }
+
+          temp_s1 = phi_a0;
```
ethteck, line 107: are you able to just use phi_a0 here and remove temp_s1 entirely?

src/world/action/jump.c
```diff
@@ -1,15 +1,220 @@
...
+void func_802B6294_E242D4() {
+    PlayerStatus* playerStatus = &gPlayerStatus;
+    s32 phi_a0;
+
+    if (playerStatus->animFlags & PLAYER_ANIM_FLAG_PEACH_PHYSICS) {
+        func_802B647C_E244BC();
+        return;
+    }
+
+    if ( playerStatus->flags < 0) {
```
ethteck, line 126: There's an extra space after `(`

ethteck (changes requested):
Thanks for all the changes! Looking much better. I found a few more things...and I think after this round it'll be all set.

nanaian (approved):
LGTM once @ethteck's nits are cleaned up! :shipit:

src/world/action/jump.c
ThreeMustyFears: No I tried removing the temp_s1 in this function and replacing it with phi_a0 but it causes reg alloc issues and adds a mov where an nop should be.

ethteck (approved):

PR #357, opened 2021-08-04
decomp functions in d5a50 and fix type for widthScale and heightScale

Rainchus:


src/d5a50_len_5fd0.c
```diff
@@ -68,13 +70,40 @@ void draw_hud_element_clipped(s32 arg0) {
...
+    draw_hud_element(arg0, 1);
+}
 
 void draw_icon_2(s32 iconID) {
     draw_hud_element(iconID, 2);
 }
 
-INCLUDE_ASM(s32, "d5a50_len_5fd0", set_hud_element_script);
+void set_hud_element_script(s32 arg0, s32 *arg1) {
+    HudElement *hudElement;
```
ethteck, line 82: Pointer on the left (`HudElement*`) and you can initialize this on the same line you declare it

src/d5a50_len_5fd0.c
```diff
@@ -68,13 +70,40 @@ void draw_hud_element_clipped(s32 arg0) {
...
+    hudElement->screenPosOffset[0] = 0;
+    hudElement->screenPosOffset[1] = 0;
+    hudElement->worldPosOffset[0] = 0;
+    hudElement->worldPosOffset[1] = 0;
+    hudElement->flags &= ~4;
+    hudElement->uniformScale = 1.0f;
+    hudElement->flags &= ~0x930;
+    load_hud_element(hudElement, arg1);
+    
+    while (hud_element_update(hudElement) != 0) {}
```
ethteck, line 104: we may want to write this differently, like as an assert or something. I'll let alex weigh in

src/d5a50_len_5fd0.c
```diff
@@ -122,7 +151,18 @@ INCLUDE_ASM(s32, "d5a50_len_5fd0", ALT_clear_hud_element_cache);
 
 INCLUDE_ASM(void, "d5a50_len_5fd0", set_hud_element_scale, s32 index, f32 scale);
 
-INCLUDE_ASM(s32, "d5a50_len_5fd0", set_hud_element_size);
+void set_hud_element_size(s32 arg0, s8 arg1) {
+    HudElement *hudElement;
```
ethteck, line 155: same with the pointer and init comment above

src/d5a50_len_5fd0.c
```diff
@@ -158,7 +198,21 @@ INCLUDE_ASM(s32, "d5a50_len_5fd0", create_hud_element_transform_B);
...
 
-INCLUDE_ASM(s32, "d5a50_len_5fd0", free_hud_element_transform);
+void free_hud_element_transform(s32 arg0) {
+    HudElement* hudElement;
+    s32* hudTransform;
+    
+    hudElement = gHudElementList[arg0 & (~0x800)];
+    hudTransform = hudElement->hudTransform;
+    
+    if ((hudElement->flags & 0x20000) == 0) {
```
ethteck, line 208: ! for flag check

src/d5a50_len_5fd0.c
```diff
@@ -158,7 +198,21 @@ INCLUDE_ASM(s32, "d5a50_len_5fd0", create_hud_element_transform_B);
...
+        func_8013A854(*hudTransform);
+    }
+
+    heap_free(hudElement->hudTransform);
+    hudElement->hudTransform = 0;
```
ethteck, line 213: NULL instead of 0

src/d5a50_len_5fd0.c
```diff
@@ -158,7 +198,21 @@ INCLUDE_ASM(s32, "d5a50_len_5fd0", create_hud_element_transform_B);
 
 INCLUDE_ASM(s32, "d5a50_len_5fd0", create_hud_element_transform_C);
 
-INCLUDE_ASM(s32, "d5a50_len_5fd0", free_hud_element_transform);
+void free_hud_element_transform(s32 arg0) {
+    HudElement* hudElement;
+    s32* hudTransform;
+    
+    hudElement = gHudElementList[arg0 & (~0x800)];
```
ethteck, line 205: inits on decl lines, no parens around ~0x800

ethteck (changes requested):
gg

src/d5a50_len_5fd0.c
nanaian: I don't think an assert would work here, because the function gets called for each loop iteration. I'm good with how it is

nanaian (approved):
LGTM!

PR #360, opened 2021-08-13
curtains.c data work, assorted funcs, new file, etc and cleanup: misc decomp

ethteck:
Updated splat which includes the new `vtx` segment, allowing us to specify vertices to extract and include from c files

Also did a couple of the insane copy protection funcs. We'll need to do the rest to even dream of shiftability (unless we just don't run them)

Made some good progress on PlayEffect

Also named and updated the PartnerActionStatus struct, added a file split, and did various cleanup


include/common_structs.h
```diff
@@ -1442,13 +1442,13 @@ typedef struct ActorPart {
...
 typedef struct ColliderTriangle {
-    /* 0x00 */ f32* v1[3]; /* note: the order of v1,2,3 is reversed from the ijk in the hit file */
-    /* 0x04 */ f32* v2[3];
-    /* 0x08 */ f32* v3[3];
+    /* 0x00 */ Vec3f* v1; /* note: the order of v1,2,3 is reversed from the ijk in the hit file */
+    /* 0x04 */ Vec3f* v2;
+    /* 0x08 */ Vec3f* v3;
     /* 0x0C */ f32 e13[3]; /* = v3 - v1 */
     /* 0x18 */ f32 e21[3]; /* = v1 - v2 */
     /* 0x24 */ f32 e32[3]; /* = v2 - v3 */
```
nanaian, line 1450: These should be Vec3f also, then.

src/316C00.c
```diff
@@ -1,3 +1,62 @@
...
+    UNK_FUN_PTR(generalHeapCreate);
+    u32 thisInsn;
+    u32* it;
+    u32 prevInsn;
+    s32 seed;
+    u32 blah;
+    u32 hash;
+
+    readFunc = osEPiReadIo;
+    seed = 0x4BF15BF6; // 0x3C016C07 + 0xFEFEFEF;
```
nanaian, line 28: Does it not match to just evaluate the expression at compile-time, like
```suggestion
    seed = 0x3C016C07 + 0xFEFEFEF;
```

src/316d90.c
```diff
@@ -1,3 +1,63 @@
...
-INCLUDE_ASM(s32, "316d90", func_802AE000);
+#define JAL   0x03
+#define ADDIU 0x09
+#define LUI   0x0F
+#define LW    0x23
+#define LOWER(x) ((x) & 0xFFFF)
+#define UPPER(x) ((x) >> 16)
+#define OPCODE(x) ((x) >> 26)
+#define GET_RS(x) ((x >> 0x15) & 0x1F)
+#define GET_RT(x) (UPPER(x) & 0x1F)
```
nanaian, line 13: Perhaps we should put these into a header to deduplicate them

src/362a0_len_2f70.c
```diff
@@ -67,7 +67,20 @@ INCLUDE_ASM(void, "362a0_len_2f70", load_hit_data, s32 idx, HitAsset* hit);
 
 INCLUDE_ASM(void, "362a0_len_2f70", parent_collider_to_model, s32 colliderID, s16 modelIndex);
 
-INCLUDE_ASM(s32, "362a0_len_2f70", _add_hit_vert_to_buffer);
+void _add_hit_vert_to_buffer(f32** buf, f32* vert, s32* bufSize) {
+    s32 i;
+
+    for (i = 0; i < *bufSize; i++) {
+        if (buf[i] == vert) {
+            break;
```
nanaian, line 75: Interesting... would this match as `return`? It seems logically consistent with what the below `if` does (check if the loop terminated early)

src/7E9D0.c
```diff
@@ -0,0 +1,317 @@
...
+
+extern s32 D_8010C924;
+extern s32 D_8010C964;
+extern s32 D_8010C96C; // npc list index
+extern s32 gSpinHistoryBufferPos; // SpinHistoryBufferPos
+extern s16 D_8010C9B0;
+extern s32 gSpinHistoryPosY[5]; // SpinHistoryPosY
+extern s32 gSpinHistoryPosX[5]; // SpinHistoryPosX
+extern s32 gSpinHistoryPosZ[5]; // SpinHistoryPosZ
+extern s16 gSpinHistoryPosAngle[5]; // SpinHistoryAngle
```
nanaian, line 12: Looks like these comments can be removed, since you named the symbols...

src/7E9D0.c
```diff
@@ -0,0 +1,317 @@
...
+        gSpinHistoryPosAngle[i] = 180;
+        gSpinHistoryPosY[i] = 0x80000000;
+    }
+}
+
+f32 phys_get_spin_history(s32 lag, s32* x, s32* y, s32* z) {
+    s32 idx = gSpinHistoryBufferPos;
+
+    // Can't get this if/else to match otherwise...
+    if (idx - lag >= 0) {
```
nanaian, line 35: What's the operator precedence here? `idx - (lag >= 0)`? 🤨

src/7bb60_len_41b0.c
```diff
@@ -110,42 +105,25 @@ void phys_init_integrator_for_current_state() {
...
+    PlayerStatus* playerStatus;
     do {} while (0);
     playerStatus = &gPlayerStatus;
+
     if (playerStatus->flags & 0x40000) {
-        playerStatus->gravityIntegrator[0] = (*(floats++)) / 12.0f;
-        playerStatus->gravityIntegrator[1] = (*(floats++)) / 12.0f;
-        playerStatus->gravityIntegrator[2] = (*(floats++)) / 12.0f;
-        playerStatus->gravityIntegrator[3] = (*(floats++)) / 12.0f;
+        playerStatus->gravityIntegrator[0] = *(floats++) / 12.0f;
```
nanaian, line 116: The parens around `(floats++)` aren't required here, I'd vote remove them like we do with `*args++` in script funcs.

src/7bb60_len_41b0.c
```diff
@@ -266,7 +291,18 @@ void collision_check_player_overlaps(void) {
...
+    Shadow* playerShadow = get_shadow_by_index(gPlayerStatus.shadowID);
+    f32 rotZ = playerShadow->rotation.z + 180.0;
+    f32 rotX = playerShadow->rotation.x + 180.0;
+    s32 ret = TRUE;
+
+    if (fabsf(rotZ) < 20.0f && fabsf(rotX) < 20.0f) {
+        ret = FALSE;
+    }
+
+    return ret;
```
nanaian, line 304: Gonna cry if `return !(fabsf(rotZ) < 20.0f && fabsf(rotX) < 20.0f);` doesn't match

src/7bb60_len_41b0.c
```diff
@@ -356,7 +392,26 @@ void phys_update_interact_collider(void) {
...
+            }
+        } else {
+            ret = FALSE;
+        }
+    } else if (!(gPlayerStatus.actionState == ACTION_STATE_IDLE ||
+                 gPlayerStatus.actionState == ACTION_STATE_WALK ||
+                 gPlayerStatus.actionState == ACTION_STATE_RUN)) {
+        ret = FALSE;
+    }
+    return ret;
```
nanaian, line 413: Does it work to `return FALSE` in each of the branches and `return TRUE` at the end (probably in an `else` to make flow clear)?

src/curtains.c
```diff
@@ -1,14 +1,181 @@
...
+    gsSPEndDisplayList(),
+};
+
+#include "vtx/curtains1.vtx.inc.c"
+#include "vtx/curtains2.vtx.inc.c"
+#include "vtx/curtains3.vtx.inc.c"
+#include "vtx/curtains4.vtx.inc.c"
+#include "vtx/curtains5.vtx.inc.c"
+#include "vtx/curtains6.vtx.inc.c"
+#include "vtx/curtains7.vtx.inc.c"
```
nanaian, line 47: Very cool!

src/curtains.c
```diff
@@ -1,14 +1,181 @@
...
+    gsSP2Triangles(0, 1, 2, 0, 0, 2, 3, 0),
+    gsSP2Triangles(4, 5, 6, 0, 4, 6, 7, 0),
+    gsSP2Triangles(8, 9, 10, 0, 8, 10, 11, 0),
+    gsSP2Triangles(12, 13, 14, 0, 12, 14, 15, 0),
+    gsSP2Triangles(15, 16, 17, 0, 15, 17, 18, 0),
+    gsSP2Triangles(19, 20, 21, 0, 19, 21, 22, 0),
+    gsSP2Triangles(13, 23, 24, 0, 13, 24, 14, 0),
+    gsSP2Triangles(23, 25, 26, 0, 23, 26, 24, 0),
+    gsSP2Triangles(25, 27, 28, 0, 25, 28, 26, 0),
+    gsSP2Triangles(27, 29, 30, 0, 27, 30, 28, 0),
```
nanaian, line 75: Yeah... I definitely think a `tris` or `dl` segment is in demand, here, as discussed. Editing the model would be too cumbersome otherwise. (For the future, of course.)

src/world/partners.c
```diff
@@ -377,13 +377,13 @@ void switch_to_partner(s32 arg0) {
...
-    Temp8010EBB0* temp8010EBB0 = &D_8010EBB0;
+    PartnerActionStatus* actionStatus = &gPartnerActionStatus;
 
     if (D_8010CFD8 != arg0) {
         D_8010CFE0 = 1;
         D_8010CFE4 = arg0;
-        temp8010EBB0->unk_00 = 0;
-        temp8010EBB0->unk_01 = 0;
+        actionStatus->actionState.b[0] = ACTION_STATE_IDLE;
+        actionStatus->actionState.b[1] = 0;
```
nanaian, line 386: What's the rationale for using a constant for b[0] but not b[1]

src/7E9D0.c
ethteck: it'd be (idx - lag) >= 0

src/7bb60_len_41b0.c
ethteck: I don't think it does, because it stores 1 into a register really early on which a single return statement didn't do, from what I remember. I can try again though

src/316C00.c
ethteck: It does, but I thought it's useful to show both the actual value and how it's composed in the comment

src/362a0_len_2f70.c
ethteck: If you mean changing it to a return and then the if below is optimization or something, then no that doesn't work

src/7bb60_len_41b0.c
ethteck: does not work, unfortunately

src/7bb60_len_41b0.c
ethteck: No. In this file, 1 is stored into a register way early on in many of these funcs, which I don't know how to replicate without the way it currently is

src/world/partners.c
ethteck: Well we don't know what b[1] represents, but we also don't know what b[0] represents so this was a mistake. I changed it back to 0.

(we do know that b[0] being 0 means idle, but that specific enum shouldn't be used here since partner action states are different)

src/7E9D0.c
nanaian: ok im stupid

src/world/partners.c
nanaian: Thought so, thanks for the fix

PR #362, opened 2021-08-14
Match some funcs in world/action/walk

ThreeMustyFears:


src/world/action/walk.c
```diff
@@ -1,13 +1,125 @@
...
+            if (temp_v1 < 0) {
+                playerStatus->targetYaw = sp10;
+            } else {
+                xAxis = 0x80000000;
+                playerStatus->animFlags |= xAxis;
+            }
+
+        }
+
+        if (!is_ability_active(0xB)) {
```
nanaian, line 96:
```suggestion
        if (!is_ability_active(ABILITY_SLOW_GO)) {
```

src/world/action/walk.c
```diff
@@ -1,13 +1,125 @@
...
+            xAxis = playerStatus->stickAxis[0] * playerStatus->stickAxis[0];
+            yAxis = playerStatus->stickAxis[1] * playerStatus->stickAxis[1];
+            if (55.0f >= sqrtf(xAxis + yAxis)) {
```
nanaian, line 99: Yuck, a yoda condition. Does it match this way?
```suggestion
            if (sqrtf(xAxis + yAxis) <= 55.0f) {
```

src/world/action/walk.c
```diff
@@ -1,13 +1,125 @@
 #include "common.h"
+#include "world/actions.h"
+
+s32 D_802B6910_E23FF0[] = {0x000A0002, 0x000A002B, 0x000A002D, 0x000A002F, 0x000A0031, 0x000A0033, 0x000A0035, 0x000A0037, 0x000A0039, 0x000A003B, 0x000A003D, 0x000A003F,
```
nanaian, line 4: These are peach player animation IDs.... it would be good to add a comment to that effect so we can parameterize them later

src/world/action/walk.c
```diff
@@ -1,13 +1,125 @@
...
+    player_input_to_move_vector(&sp10, &sp14);
+    phys_update_interact_collider();
+    if (check_input_jump() == FALSE) {
+        if (phi_s3 == 0) {
+            if (check_input_hammer()) {
+                return;
+            }
+        }
+
+        player_input_to_move_vector(&sp10, &sp14);
```
nanaian, line 73: Not required, but you could name sp10 and sp14 `moveX` and `moveY` due to this call and declare them in this block since they're only(?) used here

src/world/action/walk.c
```diff
@@ -45,4 +157,58 @@ void func_802B65E8_E23CC8(void) {
...
+
+    if (playerStatus->flags < 0) {
+        playerStatus->flags &= ~0x80000000;
+        playerStatus->unk_60 = 0;
+        if (!(playerStatus->flags & 0x00004000)) {
+            playerStatus->currentSpeed = playerStatus->runSpeed;
+        }
+
+        if (!(playerStatus->animFlags & PLAYER_ANIM_FLAG_IN_DISGUISE))
+        {
```
nanaian, line 174:
```suggestion
        if (!(playerStatus->animFlags & PLAYER_ANIM_FLAG_IN_DISGUISE)) {
```

src/world/action/walk.c
nanaian: Looking at line 207 of this file, this should be able to match without the xAxis and yAxis temps. Could you also use the SQ (square) macro?

src/world/action/walk.c
```diff
@@ -45,4 +157,58 @@ void func_802B65E8_E23CC8(void) {
...
+    phys_update_interact_collider();
+    if (sp14 == 0.0f) {
+        set_action_state(ACTION_STATE_IDLE);
+        return;
+    }
+
+    playerStatus->targetYaw = sp10;
+
+    if (sqrtf(playerStatus->stickAxis[0] * playerStatus->stickAxis[0] +
+              playerStatus->stickAxis[1] * playerStatus->stickAxis[1]) < 55.0f) {
```
nanaian, line 208:
```suggestion
    if (sqrtf(SQ(playerStatus->stickAxis[0]) + SQ(playerStatus->stickAxis[1]) < 55.0f) {
```

nanaian (changes requested):
Looks good! Just a few nits that might be worth looking into

src/world/action/walk.c
ThreeMustyFears: I can change the names but I can't move the declaration as I have to call player_input_to_move_vector at line 64 as well

nanaian: nw, i missed that

nanaian (approved):

ethteck:
great job, and thanks!

PR #373, opened 2021-08-21
Decomp functions from various files

imaandrew:


src/190B20.c
```diff
@@ -1144,11 +1144,47 @@ INCLUDE_ASM(s32, "190B20", func_80263FE8);
 
 INCLUDE_ASM(s32, "190B20", func_80264084);
 
-INCLUDE_ASM(void, "190B20", add_xz_vec3f, Vec3f* vector, f32 speed, f32 angleDeg);
+void add_xz_vec3f(Vec3f* vector, f32 speed, f32 angleDeg) {
+    f32 newVectorZ;
+    f32 angleRad;
+    f32 sinAngleRad;
+
+    angleRad = angleDeg * 6.28318f;
```
ethteck, line 1152: Here and below, the 6.28... can be `TAU`, and you can initialize angleRad on the line you declare it (along with putting it as the first variable declared):

`f32 angleRad = angleDeg * TAU / 360.0f;`

you can then declare and initialize sinAngleRad right after on the next line

src/190B20.c
```diff
@@ -1144,11 +1144,47 @@ INCLUDE_ASM(s32, "190B20", func_80263FE8);
...
+    angleRad /= 360.0f;
+    sinAngleRad = sin_rad(angleRad);
+    newVectorZ = vector->z + (-speed * cos_rad(angleRad));
+    angleRad = speed * sinAngleRad;
+    vector->x = vector->x + angleRad;
```
ethteck, line 1157: any time you have x = x op y, you can write it as x op= y, so `vector->x += angleRad;`

That being said, I have a feeling there are temps for the sin_rad  and cos_rad and then the rest of the function is just 
```
vector->x += speed * sinAngleRad;
vector->z += speed * cosAngleRad;
```

could be wrong though, but it'd make this look a lot cleaner if it matches

src/190B20.c
```diff
@@ -1640,7 +1705,14 @@ INCLUDE_ASM(s32, "190B20", btl_appendGfx_prim_quad);
 //     gDPSetCombineMode(gMasterGfxPos++, G_CC_DECALRGBA, G_CC_DECALRGBA);
 // }
 
-INCLUDE_ASM(s32, "190B20", btl_draw_prim_quad);
+void btl_draw_prim_quad(s32 arg0, s32 arg1, s32 arg2, s32 arg3, s32 arg4, s32 arg5, s32 arg6, s32 arg7) {
+    u16 new_var;
+    u16 new_var2;
+
+    new_var = (arg4 + arg6) & 0xFFFF;
```
ethteck, line 1712: Because new_var and new_var2 are of type u16, the 0xFFFF is actually implicit. You should be able to remove that from both vars

src/190B20.c
```diff
@@ -1640,7 +1705,14 @@ INCLUDE_ASM(s32, "190B20", btl_appendGfx_prim_quad);
...
+    new_var2 = (arg5 + arg7) & 0xFFFF;
+    btl_appendGfx_prim_quad(arg0 & 0xFF, arg1 & 0xFF, arg2 & 0xFF, arg3 & 0xFF, arg4 & 0xFFFF, arg5 & 0xFFFF, new_var, new_var2);
```
ethteck, line 1714: These 0xFFs are most likely u8 casts, and the 0xFFFFs u16 casts, so the ideal thing to do here is to find btl_appendGfx_prim_quad, and modify its signature accordingly to be u8, u8, u8, u8, u16, u16, u16, u16 and then add a declaration for it in functions.h if that function doesn't exist in this file. If the function isn't decomped, you'll need to modify the INCLUDE_ASM line for it, and that's kind of weird since it's a macro. let us know if you'd like help in discord

after this, you should be able to remove al the & 0xFF(FF)s

src/190B20.c
```diff
@@ -1471,9 +1507,38 @@ INCLUDE_ASM(s32, "190B20", func_80266F60);
...
+    s32 flags;
+
+    actorPartsTable = actor->partsTable;
+    if (actorPartsTable != NULL) {
+        do {
+            flags = actorPartsTable->flags;
+            if (!(flags & 0x100001) && (actorPartsTable->decorationTable != NULL) && !(flags & 2) && (actorPartsTable->idleAnimations != NULL)) {
+                func_80266FD8(actorPartsTable, arg1);
+            }
+            actorPartTemp = actorPartsTable->nextPart;
```
ethteck, line 1536: This temps (actorNextPart and actorPartTemp) shouldn't be necessary. You should be able to do actorPartsTable = actorPartsTable->nextPart

that being said, I think you should change the name of this temp since it's not really used as the table but is just an iterator for the parts inside a table. so maybe just `actorPart`.

and on the first line where you initialize it, you technically want the first one in the list, so it'd be `actorPart = &actor->partsTable[0];`. You may or may not need the & there - I don't remembe the type of the table exactly

src/190B20.c
```diff
@@ -1471,9 +1507,38 @@ INCLUDE_ASM(s32, "190B20", func_80266F60);
...
+void func_80267018(Actor* actor, s32 arg1) {
+    ActorPart* actorPartsTable;
+    ActorPart* actorNextPart;
+    ActorPart* actorPartTemp;
+    s32 flags;
+
+    actorPartsTable = actor->partsTable;
+    if (actorPartsTable != NULL) {
+        do {
+            flags = actorPartsTable->flags;
```
ethteck, line 1532: You shouldn't need this flags temp either, I would think

ethteck (changes requested):
Thanks for your first PR and welcome to the cool decomp club 😎 

There are a few things that I saw that could clean these functions up. Let me know if you have any questions about anything I wrote!

src/190B20.c
```diff
@@ -1144,11 +1144,32 @@ INCLUDE_ASM(s32, "190B20", func_80263FE8);
...
+}
 
-INCLUDE_ASM(void, "190B20", add_xz_vec3f_copy2, Vec3f* vector, f32 speed, f32 angleDeg);
+void add_xz_vec3f_copy2(Vec3f* vector, f32 speed, f32 angleDeg) {
+    f32 angleRad = angleDeg * TAU / 360.0f;
+    f32 sinAngleRad = sin_rad(angleRad);
+    f32 cosAngleRad = cos_rad(angleRad);
+
+    vector->x += speed * sinAngleRad;
+    vector->z += -speed * cosAngleRad;
```
ethteck, line 1171: Sorry, one last thing: Can this be -= speed * cosAngleRad?

ethteck (approved):
Looks great! Thanks!

src/190B20.c
```diff
@@ -1471,9 +1492,29 @@ INCLUDE_ASM(s32, "190B20", func_80266F60);
 
 INCLUDE_ASM(s32, "190B20", func_80266F8C);
 
-INCLUDE_ASM(s32, "190B20", func_80266FD8);
+void func_80266FD8(ActorPart* part, s32 arg1) {
+    DecorationTable* decorationTable;
+
+    if ((part->idleAnimations != NULL) && !(part->flags & 2)) {
```
nanaian, line 1498:
```suggestion
    if (part->idleAnimations != NULL && !(part->flags & 2)) {
```

src/190B20.c
```diff
@@ -1471,9 +1492,29 @@ INCLUDE_ASM(s32, "190B20", func_80266F60);
...
+        }
+    }
+}
+
+void func_80267018(Actor* actor, s32 arg1) {
+    ActorPart* actorPart = &actor->partsTable[0];
 
-INCLUDE_ASM(s32, "190B20", func_80267018);
+        while (actorPart != NULL) {
+            if (!(actorPart->flags & 0x100001) && (actorPart->decorationTable != NULL) && !(actorPart->flags & 2) && (actorPart->idleAnimations != NULL)) {
```
nanaian, line 1512:
```suggestion
            if (!(actorPart->flags & 0x100001) && actorPart->decorationTable != NULL && !(actorPart->flags & 2) && actorPart->idleAnimations != NULL) {
```

nanaian (changes requested):

nanaian:
Great PR! Welcome to the project!! 📦 
Made a couple suggested changes to remove some extraneous parentheses but other than that it looks great :)

nanaian (approved):
Thanks for the quick fixes 👍

PR #375, opened 2021-08-22
Tidying

ethteck:
We now run undefined_syms.txt through cpp, so we can organize it better. I removed the bss and dead syms linker files and added sections for them in undefined_syms.txt. 

Decomped some various audio and other functions

Named a nusys func and decomped it

Removed many unnecessary compiler flags


src/26840_len_20d0.c
```diff
@@ -185,29 +185,81 @@ INCLUDE_ASM(s32, "26840_len_20d0", snd_set_voice_volume);
...
+void snd_SEFCmd_14(SoundManager* manager, SoundPlayer* player);
+void snd_SEFCmd_15(SoundManager* manager, SoundPlayer* player);
+void snd_SEFCmd_16(SoundManager* manager, SoundPlayer* player);
+void snd_SEFCmd_17(SoundManager* manager, SoundPlayer* player);
+void snd_SEFCmd_18(SoundManager* manager, SoundPlayer* player);
+
 INCLUDE_ASM(void, "26840_len_20d0", snd_SEFCmd_00_SetVolume, SoundManager* manager, SoundPlayer* player);
 
-INCLUDE_ASM(void, "26840_len_20d0", snd_SEFCmd_01_SetPan, SoundManager* manager, SoundPlayer* player);
+//INCLUDE_ASM(void, "26840_len_20d0", snd_SEFCmd_01_SetPan, SoundManager* manager, SoundPlayer* player);
```
nanaian, line 214:
```suggestion
```

tools/build/configure.py
```diff
@@ -13,7 +13,7 @@
 CPPFLAGS = "-w -Iver/$version/build/include -Iinclude -Isrc -Iassets/$version -D _LANGUAGE_C -D _FINALROM -D VERSION=$version " \
             "-ffreestanding -DF3DEX_GBI_2 -D_MIPS_SZLONG=32 -MD -MF $out.d"
 
-ASFLAGS = "-EB -G 0"
+ASFLAGS = "-G0"
```
nanaian, line 16: Just noticing this now but ideally we shouldn't use `-g0` (clear debug info?) when --debug is passed

nanaian (changes requested):

tools/build/configure.py
nanaian: unless ofc -G does something different for as

Issue #366, opened 2021-08-16
Fix warnings


nanaian:
Compiling the rom emits a ton of warnings that we should be able to fix - mostly implicit casts that should really be made explicit.
Additionally it would be nice to be able to enable `-Wimplicit -Wredundant-decls` so headers are used properly.

Once this is done we can enable `-Werror` on Jenkins so people fix their warnings as they match stuff :shipit:

imkhali:
Hi nanaian,
I think I can do this

ethteck:
Let us know if you have any questions! Feel free to reach out on discord as well. @imkhali

nanaian:
We should remove `-Wshadow` entirely because the DL-building macros emit `warning: declaration of `_g' shadows previous local`

ethteck:
I think this can be closed now. We've severely reduced the number of warnings and continue to do so

ethteck:
Gonna close, but someone can re-open if they disagree. I mostly think this is moot cause it was made before the warning check system / Bowser Slug entered the fray

PR #380, opened 2021-08-23
Enforce warnings count in pull requests

nanaian:
This PR:

* Steals [mm](https://github.com/zeldaret/mm)'s warnings_count tool
* Requires that the warnings count does not increase in pull requests
* Enables the following warnings: `-Wimplicit -Wredundant-decls -Wstrict-prototypes` (#366)

As of PR creation there are _17,370_ warnings emitted on a full build of both us and jp. This amount makes it kinda unacceptable for us to ask anyone to fix it so we will wait to merge this until a bunch of them -- probably those in common header files -- are fixed.

Additionally, before merge, I'd want to **not** store `warnings.txt` in the repo (as this would require that contributors update it manually and that's very annoying) - instead have Jenkins store the master's current warnings at https://papermar.io/warnings.txt which can be compared against in PR runs.


nanaian:
When #382 merges it would be nice to change this PR to have slug comment `This PR introduces the following 4 warnings:` or `This PR fixes 18 warnings!` instead of failing the build, thoughts @ethteck?

ethteck:
Yeah, sounds great to me!

BowserSlug:
⚠️ This PR introduces 17383 warnings:
See log for details.

PR #389, opened 2021-08-25
Power Shock - Adding 4/5 matching funcs, #ifdef'd nonmatching-but-close update function

jdaster64:
Adding the four matching functions for Power Shock, and #ifdef-ing a mostly-there update function that needs work on temps, etc.

Also updating symbol_addrs with the correct name for the Tidal Wave update function.


src/battle/action_cmd/power_shock.c
```diff
@@ -2,12 +2,321 @@
...
+        actionCommandStatus->unk_48 = 0;
+        actionCommandStatus->unk_46 = rand_int(100);
+        actionCommandStatus->unk_6C = 30;
+        actionCommandStatus->hudElementX = -48;
+        actionCommandStatus->unk_68 = 0;
+        actionCommandStatus->unk_66 = 0;
+        D_802A9B00 = 0;
+        actionCommandStatus->hudElementY = 80;
+
+        hudElement = create_hud_element(&D_80108B28);
```
nanaian, line 41: Can you type D_80108B28 and all the other hud element scripts as s32 arrays instead and remove the `&`?

src/battle/action_cmd/power_shock.c
```diff
@@ -2,12 +2,321 @@
...
+    BattleStatus* battleStatus = &gBattleStatus;
+    HudElement* hudElement;
+    s32 temp_v1;
 
+    battleStatus->unk_82 = 0x64;
+    battleStatus->unk_434 = &D_802943A0;
+
+    if (battleStatus->unk_83 == 0) {
+        battleStatus->actionSuccess = 0;
+        return ApiStatus_DONE2;
```
ethteck, line 25: If this return is needed, the rest of the function shouldn't be in an else block and should just be after this if

src/battle/action_cmd/power_shock.c
```diff
@@ -2,12 +2,321 @@
...
+            }
+            if (actionCommandStatus->unk_54 != 0) {
+                actionCommandStatus->unk_54--;
+                break;
+            }
+            func_80268C9C();
+            break;
+    }
+
+    // Invisible "run away"-like meter that determines success threshold.
```
nanaian, line 262: Might be funny to link to [this video](https://youtu.be/FagUpV0c67s) here :shipit: 
We can also name unk_46 "meterValue" or something.

ethteck (changes requested):

nanaian (approved):
very cool 🎸

src/battle/action_cmd/power_shock.c
jdaster64: Update; might be worth changing tidal_wave's (and others'?) as well.

src/battle/action_cmd/power_shock.c
jdaster64: Done; done for D_802943A0 as well since it looks to be an array of multipliers for how much the bar fills per press indexed by actionCommandStatus->unk_50.

src/battle/action_cmd/power_shock.c
ethteck: oh..if those are how this was then yeah that's not ideal

src/battle/action_cmd/power_shock.c
jdaster64: Added a link in the comment.

src/battle/action_cmd/power_shock.c
nanaian: Yeah it looks like this check is for a kind of universal "fail all action commands?" bool

ethteck (approved):

PR #398, opened 2021-08-25
add si_funcs.h

Brotenko:


nanaian (approved):
LGTM but may be worth including this in #403 since its very related

Brotenko:
Combined with PR #403.

PR #406, opened 2021-08-29
Match func_80144E4C

JoshDuMan:
Hopefully this builds!


BowserSlug:

✅ This PR fixes 1 warnings!

nanaian (approved):
I don't love these unions for flags, but we do do them elsewhere and I'm unsure of a way to clean them up besides casting inside func_80144E4C, which also isn't ideal.

PR #411, opened 2021-08-30
Battle Popup Decomp

JaThePlayer:
Decomps most of src/17D6A0.c


src/17D6A0.c
```diff
@@ -24,37 +119,132 @@ INCLUDE_ASM(s32, "17D6A0", func_8024F5AC);
...
 
-INCLUDE_ASM(s32, "17D6A0", func_8024F940);
+ApiStatus func_8024F940(Evt* script, s32 isInitialCall) {
+    func_8024F7C8();
+    return ApiStatus_DONE2;
+}
 
-INCLUDE_ASM(s32, "17D6A0", btl_show_battle_message);
+void btl_show_battle_message(s32 messageIndex, s32 duration) {
+    BattlePopup* popup = btl_create_popup();
```
ethteck, line 149: New line after the decl pleez

src/17D6A0.c
```diff
@@ -24,37 +119,132 @@ INCLUDE_ASM(s32, "17D6A0", func_8024F5AC);
...
+    if (popup != NULL) {
+        popup->unk_04 = &func_8024FB3C;
```
ethteck, line 151: & unneeded for function pointers

src/17D6A0.c
```diff
@@ -24,37 +119,132 @@ INCLUDE_ASM(s32, "17D6A0", func_8024F5AC);
...
+        D_8029F64A = 0;
+        D_8029F64C = 0;
+        D_8029F64E = 0;
+        D_8029F650 = 0;
+    }
+}
 
-INCLUDE_ASM(s32, "17D6A0", btl_show_variable_battle_message);
+void btl_show_variable_battle_message(s32 messageIndex, s32 duration, s32 varValue) {
+    BattlePopup* popup = btl_create_popup();
```
ethteck, line 170: Empty line

BowserSlug:
🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀 This PR matches 20 functions (+0.04%) on `us`.

✅ This PR fixes 1 warnings!

src/17D6A0.c
```diff
@@ -24,37 +119,132 @@ INCLUDE_ASM(s32, "17D6A0", func_8024F5AC);
...
-INCLUDE_ASM(s32, "17D6A0", func_8024FAFC);
+void func_8024FAFC(void) {
+    D_8029F64C = 0;
+}
 
-INCLUDE_ASM(s32, "17D6A0", close_action_command_instruction_popup);
+void close_action_command_instruction_popup(void) {
+    BattlePopup* popup = D_802838F8;
+    if (popup != NULL) {
+        if (popup->messageIndex >= 67) {
```
ethteck, line 217: can these two ifs be combined into an or? Also could you invert the condition instead of having the return?

src/17D6A0.c
```diff
@@ -24,37 +119,132 @@ INCLUDE_ASM(s32, "17D6A0", func_8024F5AC);
 
 INCLUDE_ASM(s32, "17D6A0", func_8024F768);
 
+void func_8024F7C8(void);
 INCLUDE_ASM(s32, "17D6A0", func_8024F7C8);
 
-INCLUDE_ASM(s32, "17D6A0", func_8024F84C);
+s32 func_8024F84C(Evt* script, s32 isInitialCall) {
```
JoshDuMan, line 125:
```suggestion
ApiStatus func_8024F84C(Evt* script, s32 isInitialCall) {
```

JoshDuMan (commented):
Looks almost perfect to me!

ethteck (changes requested):
Nice! Also, BattlePopup.active I think is a bool, so any time you set it to 0 or 1, you can do TRUE/FALSE and when you're testing it, you don't need != 0 for the true check. !active for the false check

I didn't put it everywhere I saw it, but please separate the decls from the rest of the func with a new line and remove & from func pointers. :)

src/17D6A0.c
JaThePlayer: Everything I tried to clean this up didn't match, including inverting the conditions. It's really weird.

JaThePlayer: Oh nevermind, combining all 3 of these if's matches lol

src/17D6A0.c
```diff
@@ -1,21 +1,116 @@
...
+    /* 0x0C */ BattlePopupCallback drawFunc;
+    /* 0x10 */ s16 active;
+    /* 0x12 */ s16 messageIndex;
+    /* 0x14 */ s16 duration;
+    /* 0x16 */ s8 unk_16;
+    /* 0x17 */ s8 unk_17;
+    /* 0x18 */ s32* message;
+} BattlePopup; // size = 0x1C
+
+extern BattlePopup BattlePopups[32];
```
ethteck, line 19: oop - didn't catch it, but this should be lowercase b `battlePopups`

ethteck (approved):

include/common_structs.h
```diff
@@ -2283,4 +2283,19 @@ typedef struct ActionCommandStatus {
     /* 0x74 */ s16 mashMeterCutoffs[6]; // upper bounds for each interval
     /* 0x80 */ s8 mashMeterIntervals;
 } ActionCommandStatus;
+
+struct PopupMessage;
+typedef void (*PopupMessageCallback)(struct PopupMessage* popup);
+typedef struct PopupMessage {
```
nanaian, line 2289: Assuming it works, I think it would better to do:
```suggestion
typedef struct PopupMessage PopupMessage;
typedef void (*PopupMesageCallback)(PopupMessage* popup);
struct PopupMessage { /* ... */ };
```

src/17D6A0.c
```diff
@@ -1,21 +1,101 @@
...
+#include "effects.h"
 
+extern PopupMessage popupMessages[32];
+extern PopupMessage* D_802838F8;
 extern s16 D_8029F64C;
+extern s16 D_8029F640;
+extern s16 D_8029F64A;
+extern s16 D_8029F64E;
+extern s16 D_8029F650;
+extern void func_8024FB3C(PopupMessage* popup);
```
nanaian, line 11: Extern is not needed for functions

src/17D6A0.c
```diff
@@ -1,21 +1,101 @@
...
+        popup->message = NULL;
+    }
+}
+
+void func_8024EDEC(void) {
+    PopupMessage* popup;
+    s32 i;
+
+    for (i = 0; i < ARRAY_COUNT(popupMessages); i++) {
+        popup = &popupMessages[i];
```
nanaian, line 30: IDM particularly but in my opinion its nicer to declare popup on this line also instead of at the top of the function

nanaian: This goes for all funcs here

PR #409, opened 2021-08-29
Decomp some pause screen map functions

Rainchus:

PR #412, opened 2021-09-01
Decomp some pause screen map functions

ethteck:
This was Rain's PR (#409), but due to a git mishap, I had to re-open it here


BowserSlug:
🚀🚀🚀🚀🚀🚀🚀 This PR matches 7 functions (+0.08%) on `us`.

⚠️ This PR introduces 4 warnings:
- src/pause/140C70.c:218: warning: initialization from incompatible pointer type
- src/pause/140C70.c:75: warning: passing arg 1 of `create_hud_element' makes pointer from integer without a cast
- src/pause/140C70.c:80: warning: assignment from incompatible pointer type
- src/pause/140C70.c:83: warning: implicit declaration of function `setup_pause_menu_tab'

JoshDuMan (approved):
My only consideration would be to rename 140C70 to pause_map. Gj again to Rain!

include/common_structs.h
```diff
@@ -2031,32 +2031,94 @@ typedef struct {
...
-    /* 0x04 */ char unk_04[4];
-    /* 0x08 */ s8* unk_08;
-    /* 0x0C */ void* fpInit;
-    /* 0x10 */ void* fpHandleInput;
-    /* 0x14 */ void* fpUpdate;
-    /* 0x18 */ void* fpCleanup;
-} MenuTab; // size = 0x1C
+typedef struct PauseMapSpace {
+    /* 0x00 */ s16 xPos;
+    /* 0x02 */ s16 yPos;
```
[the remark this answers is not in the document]
nanaian: This goes for the other xPos/yPos pairs in the other new structs

src/pause/140C70.c
```diff
@@ -1,17 +1,244 @@
 #include "common.h"
 
-INCLUDE_ASM(s32, "pause/140C70", func_8024D930);
+#define pause_map_tab_places_desc_string_id 0x1D00F2
```
nanaian, line 3: This should really go in `tools/splat_ext/msg.yaml` instead and referenced as `MSG_pause_map_tab_places_desc` from `message_ids.h`.

src/pause/140C70.c
```diff
@@ -1,17 +1,244 @@
...
+            break;
+    }
+}
 
 INCLUDE_ASM(s32, "pause/140C70", pause_map_draw_contents);
 
-INCLUDE_ASM(s32, "pause/140C70", pause_map_draw_title);
+void pause_map_draw_title(s32* arg1, s32 arg2, s32 textOffsetY, s32 textOffsetX) {
+    s32 stringWidth;
+    s32 stringID;
```
nanaian, line 50: Call these msgWidth and msgID; we're calling strings messages in most places except from the not-yet-renamed string funcs.

src/pause/140C70.c
nanaian: Actually, this is the _base_ msg ID, where the following messages are in pause_map order. It may be worth naming all the messages to that effect; i.e.
```
- [0x1D, 0x0F2, pause_map_location_desc_0]
- [0x1D, 0x0F3, pause_map_location_desc_1]
- [0x1D, 0x0F4, pause_map_location_desc_2]
...
```

nanaian (changes requested):

include/common_structs.h
```diff
@@ -2031,32 +2031,94 @@ typedef struct {
...
+    /* 0x05 */ u8 pathLength;
+    /* 0x06 */ s16 unk_06; // always 0
+    /* 0x08 */ Vec2b* path;
+    /* 0x0C */ s32 description;
+    /* 0x10 */ s32 id;
+} PauseMapSpace; //size = 0x14
+
+typedef struct MenuPanel {
+    /* 0x00 */ s8 initialized; //?
+    /* 0x01 */ s8 col; // might be backwards
```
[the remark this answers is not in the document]
ethteck: potentially, but I trust clover more than nothing - we should wait till we decomp it

src/pause/140C70.c
ethteck: I don't think the strings are consecutive due to the code that multiplies the option by 3 and then adds the base. I think it'd be best to wait to figure out what these are before creating dummy vals for them

src/pause/140C70.c
ethteck: I changed them, but it's a bit weird that all the functions pertaining to these say string and not msg

nanaian: See #413

src/pause/140C70.c
nanaian: I recall what they are: name, 'before' desc, 'after desc'.

PR #415, opened 2021-09-06
Match super boots

ThreeMustyFears:


BowserSlug:
🚀🚀 This PR matches 2 functions (+0.05%) on `us`.

⚠️ This PR introduces 6 warnings:
- src/world/action/super_boots.c:129: warning: implicit declaration of function `exec_ShakeCam1'
- src/world/action/super_boots.c:132: warning: assignment from incompatible pointer type
- src/world/action/super_boots.c:183: warning: implicit declaration of function `func_802B6648_E27428'
- src/world/action/super_boots.c:183: warning: previous implicit declaration of `func_802B6648_E27428'
- src/world/action/super_boots.c:191: warning: `func_802B6648_E27428' was previously implicitly declared to return `int'
- src/world/action/super_boots.c:191: warning: type mismatch with previous implicit declaration

ThreeMustyFears:
I could remove the other 2 warnings if I added the function declaration for func_802B6648_E27428 to a header, but I'm wondering if it should be put in actions.h or if I should create a header for this file alone.

src/world/action/super_boots.c
```diff
@@ -1,5 +1,206 @@
...
+                            set_action_state(ACTION_STATE_HIT_LAVA);
+                            playerStatus->flags &= ~0x8;
+                            return;
+                    } else {
+                        playerStatus->gravityIntegrator[1] = -3.4744f;
+                        playerStatus->gravityIntegrator[2] = 0.4704f;
+                        playerStatus->gravityIntegrator[3] = -1.1904f;
+                        playerStatus->gravityIntegrator[0] = 10.0f;
+                        suggest_player_anim_clearUnkFlag(0x1000B);
+                        playerStatus->actionState = 0xE;
```
ethteck, line 124: I think we should have an action state enum value for E - if not, could you add it?

ethteck (approved):
Lgtm! Sorry for the late review. I don't mind where the function decls go, between a specific super boot header or the actions header. I'm guessing alex will have a stronger opinion

nanaian (approved):
w.r.t. that warning....

For me it depends on whether `func_802B6648_E27428` is used only in this file or not. I did a brief search and it _seems_ that it is, in which case we should just throw the declaration at the top of `super_boots.c` IMO.

ethteck:
sounds good to me!

PR #417, opened 2021-09-10
Adding PartnerAfflictEnemy function

jdaster64:


Jenkinsfile
```diff
@@ -10,7 +10,7 @@ pipeline {
                 sh 'cp /usr/local/etc/roms/papermario.jp.z64 ver/jp/baserom.z64'
                 sh 'curl -L "https://github.com/pmret/gcc-papermario/releases/download/master/linux.tar.gz" | tar zx -C tools/build/cc/gcc'
                 sh 'curl -L "https://github.com/pmret/binutils-papermario/releases/download/master/linux.tar.gz" | tar zx -C tools/build/cc/gcc'
-                sh './configure'
```
ethteck, line 13: I think this accidentally got added back to the Jenkinsfile

src/1AC760.c
```diff
@@ -195,7 +195,82 @@ ApiStatus GetActorLevel(Evt* script, s32 isInitialCall) {
...
+    battleStatus->currentAttackEventSuppression = *args++;
+    battleStatus->currentAttackStatus = *args++;
+    battleStatus->currentAttackStatus |= evt_get_variable(script, *args++);
+    battleStatus->currentAttackDamage = evt_get_variable(script, *args++);
+    battleStatus->powerBounceCounter = 0;
+
+    flags = *args++;
+    if ((flags & 0x30) == 0x30) {
+        battleStatus->flags1 |= 0x30;
+    } else {
```
ethteck, line 219: This should be able to be mde into a series of if else/if instead of having this nested part after

src/1AC760.c
```diff
@@ -195,7 +195,82 @@ ApiStatus GetActorLevel(Evt* script, s32 isInitialCall) {
...
+        gBattleStatus.flags1 |= 0x800;
+    } else {
+        gBattleStatus.flags1 &= ~0x800;
+    }
+
+    statusChance = battleStatus->currentAttackStatus;
+    battleStatus->currentTargetID = actor->targetActorID;
+    battleStatus->currentTargetPart = actor->targetPartIndex;
+    battleStatus->statusChance = statusChance;
+    if (statusChance == 0xFF) {
```
ethteck, line 257: -1 I think

ethteck (changes requested):
Woot! Small changes

src/1AC760.c
jdaster64: Logically, I think that makes sense (there's a sentinel for 0xFE as well) but it doesn't match; perhaps the battleStatus field type is wrong?

src/1AC760.c
jdaster64: Done; will push change in a bit.

BowserSlug:
🚀 This PR matches 1 function (+0.02%) on `us`.

ethteck (approved):

PR #420, opened 2021-09-16
a5dd0 (model file) + filemenu decomp

ethteck:
I did a bunch of filemenu and model file decomp. I'm not renaming the model file yet because I believe it can be split and should be done so beforehand. However, it's gonna take some time, so I'll tackle it in the future. I started headers for both of these, too. They're not complete, but they're a start.

I also started work on a tool to sort undefined_syms


BowserSlug:
🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀 This PR matches 71 functions (+0.32%) on `us`.

PR #421, opened 2021-09-19
Fix alignment

ethteck:
Updated splat to remove the linker hack and fixed some long-standing padding issues in the repo that required it

One small step for eth, one giant leap for shiftability-kind.


ethteck:
Draft while I wait for someone to verify https://github.com/ethteck/splat/pull/102 doesn't break things.

nanaian (approved):

PR #423, opened 2021-09-21
Added some missing notes to common_structs.h

eldexterr:
Added missing notes from BattleStatus(0x172) & (0x174)


BowserSlug:
✅ This PR fixes 4178 warnings!

include/common_structs.h
```diff
@@ -714,7 +714,8 @@ typedef struct BattleStatus {
     /* 0x140 */ s16 enemyIDs[24];
     /* 0x170 */ char unk_170;
     /* 0x171 */ s8 numEnemyActors;
-    /* 0x172 */ char unk_172[6];
+    /* 0x172 */ s16 currentTurnActorID;
+    /* 0x174 */ s32 Actor* currentTurnActor;
```
ethteck, line 718: You have two types for this field declaration: `s32` and `Actor*` - only one should be there

ethteck (changes requested):
Looks good other than my comment! Where are these changes coming from, btw?

nanaian (commented):
> This PR fixes 4286 warnings!

whoa.

I'll defer to ethan on this one review-wise

include/common_structs.h
eldexterr: Hmm is the pointer of data to the enemy attacking so I guess should be like this?
`struct Actor* currentTurnActor`

I just noticed that by looking at the RAM, but the code that sets those values is at ~Func:btl_state_update_next_enemy (802472AC, 175B8C), also looks like I made some mistakes and is actually "currentTurnEnemyID" It doesn't apply for Player or Partner, only for enemies.

So I think a better approach should be
"currentTurnEnemyID" & "currentTurnEnemy"

ethteck: Ok, sounds good! And yeah I think that type makes more sense.

ethteck (approved):

PR #424, opened 2021-09-22
Set up IDO and match osFlash.c

ethteck:
One file in the repo, osFlash.c, is compiled with IDO. For this PR, I added IDO as a dependency to the project and set up compilation of this file, along with correcting some file splits we had and moving the yay0 decode function out of the non_matching dir.

I also fixed newline issues in the libultra headers. View this PR ignoring whitespace changes to preserve your sanity.


BowserSlug:
✅ This PR fixes 3511 warnings!

nanaian (approved):

PR #426, opened 2021-09-22
Decompile dead functions

Brotenko:


src/EF2680.c
```diff
@@ -3,44 +3,443 @@
...
+    y = npc->pos.y;
+    z = npc->pos.z;
+    w = 1000.0f;
+
+    npc_raycast_down_sides(npc->unk_80, &x, &y, &z, &w);
+    npc->pos.y = y + temp_f24 + (sin_deg(enemy->varTable[2]) * temp_f20);
+    enemy->varTable[2] = clamp_angle(enemy->varTable[2] + 0xC);
+
+    if (aiSettings->unk_14 >= 0) {
+        if (script->functionTemp[1] <= 0) {
```
ethteck, line 103: Can this if be &&d with the previous?

src/EF2680.c
```diff
@@ -3,44 +3,443 @@
...
-INCLUDE_ASM(s32, "EF2680", func_802424B0_EF4AC0);
+#include "world/common/set_script_owner_npc_col_height.inc.c"
 
 INCLUDE_ASM(s32, "EF2680", func_80242530_EF4B40);
 
 INCLUDE_ASM(s32, "EF2680", func_8024299C_EF4FAC);
+// This should be roughly like this and is a copy from CD7350 (func_80242918_CD9BF8), but with the little extra block
+// at the beginning. I have to assume that is has to roughly look like below, but I can't get it to match.
+/*
+ApiStatus func_8024299C_EF4FAC(Evt* script, s32 isInitialCall) {
```
ethteck, line 218: Instead of the comment, nonmatching ifdef plz

ethteck (changes requested):
LGTM aside from the comments I left. the warning added is fine cause we shouldn't be using that temp struct so it's a reminder to get rid of it

src/EF2680.c
Brotenko: Can't be done, because of `script->functionTemp[1]--;` on line 118, which only requires the first statement to be true, but not the second one.

ethteck: ahh kk

ethteck (approved):

PR #427, opened 2021-09-23
BattleStatus (0x436) added (from common_structs.h)

eldexterr:
This is defined every time an actor is created with ~Func:create_actor (802653A4, 193C84), at ~Func:create_actor(0x908)
It retrieves the ID two lines before by using function 80047608 and will get the next available index.


ethteck:
Cool! Could you start the field name with a lowercase h though? Also since you're looking at struct fields, it may be worth consolidating a few more changes into one PR instead of a having a PR per field :)

ethteck:
Also, when you rename a struct field like this, you need to update its name in the code as well. In vscode, you can do this by right clicking the field and pressing "rename references" or something like that

eldexterr:
> Also, when you rename a struct field like this, you need to update its name in the code as well. In vscode, you can do this by right clicking the field and pressing "rename references" or something like that

Oh my bad, should be fixed now 👍

eldexterr:
> Cool! Could you start the field name with a lowercase h though? Also since you're looking at struct fields, it may be worth consolidating a few more changes into one PR instead of a having a PR per field :)

Ikr XD
Isn't like I'm looking for the missing fields, I'm doing my stuff and then I found out that field is missing in decomp so I add it xd
But I will try to create PRs when I have at least a few changes not only one 😃👍

eldexterr:
Also, I took advantage of using this same PR to rename the following functions:
```
remove_status_1 > remove_status_debuff
remove_status_2 > remove_status_static
remove_status_3 > remove_status_transparent
```
I add it in the commit description too :)

ethteck (approved):

PR #434, opened 2021-09-26
Decompile E225B0.c.

eperdew:
This is my first PR for this project. I tried to follow the style of other files in the project, but I expect I missed some things.


BowserSlug:
🚀🚀🚀 This PR matches 3 functions (+0.04%) on `us`.

src/E225B0.c
```diff
@@ -1,7 +1,203 @@
...
+            temp_a0->unk_24 = 3;
+            return;
+        }
+
+        if (temp_a0->unk_18++ >= 0x10) {
+            temp_a0->scale = 0.36f;
+            temp_a0->unk_24 += 1;
+        }
+        break;
+
```
ethteck, line 174: These empty lines after `break;` aren't necessary

src/E225B0.c
```diff
@@ -1,7 +1,203 @@
...
+    struct802B7C78* temp_a0;
+    struct802B7C78* temp_v1_3;
+    s32 phi_v0;
+    PlayerStatus* playerStatus = &gPlayerStatus;
+    PartnerActionStatus* partnerActionStatus = &gPartnerActionStatus;
+
+    temp_a0 = D_802B7C78_E23228;
+    temp_f2 = temp_a0->y;
+    temp_v1 = temp_a0->unk_24;
+    temp_a0->y = temp_f2 + (((playerStatus->position.y + (f32) playerStatus->colliderHeight + 10.0f)
```
ethteck, line 143: the f32 cast may not be necessary here

src/E225B0.c
```diff
@@ -1,7 +1,203 @@
...
+    /* 0x00 */ Matrix4f* unk_00;
+    /* 0x04 */ Matrix4f* unk_04;
+    /* 0x08 */ s16 unk_08;
+    /* 0x0A */ s16 unk_0A;
+    /* 0x0C */ s16 unk_0C;
+    /* 0x0E */ s16 unk_0E;
+    /* 0x10 */ u8 unk_10;
+} UnknownCommand;
+
+void func_8013AF70(s32, UnknownCommand*, s32, Matrix4f*); // extern
```
ethteck, line 34: the //extern comment isn't necessary

src/E225B0.c
```diff
@@ -1,7 +1,203 @@
 #include "common.h"
 
-INCLUDE_ASM(s32, "E225B0", func_802B7000_E225B0);
+void func_8013AA9C(s32, s32, s32, s32, s32, s32, s32); // extern
```
ethteck, line 3: I'd move this function decl down to be with the other one (and the comment can be removed)

src/E225B0.c
```diff
@@ -1,7 +1,203 @@
...
+        }
+
+        temp = D_802B7C78_E23228->unk_20;
+        temp = temp - (temp / 12) * 12;
+        switch (temp) {
+        case 0:
+        case 1:
+        case 2:
+        case 3:
+            // spA4 = &D_802B7BA0_E23150;
```
ethteck, line 76: These comments are unnecessary

src/E225B0.c
```diff
@@ -1,7 +1,203 @@
...
+
+        localD_802B7C78_E23228 = D_802B7C78_E23228;
+        if (localD_802B7C78_E23228->unk_18 < 0x2F) {
+            localD_802B7C78_E23228->unk_20 += 1;
+        }
+
+        temp = D_802B7C78_E23228->unk_20;
+        temp = temp - (temp / 12) * 12;
+        switch (temp) {
+        case 0:
```
ethteck, line 72: Please indent all switch contents up one level (so the cases are not on the same level as the switch keyword)

src/E225B0.c
```diff
@@ -1,7 +1,203 @@
...
+        command.unk_10 = 0xFF;
+
+        func_8013AF70(0, &command, 0, &matrix2);
+        gSPPopMatrix(gMasterGfxPos++, 0);
+    }
+}
+
+
+void func_802B72C0_E22870(void) {
+    mem_clear((s8*)D_802B7C78_E23228, 0x2C);
```
ethteck, line 110: instead of 0x2C, you should be able to do `sizeof(D_802B7C78_E23228)`

src/E225B0.c
```diff
@@ -1,7 +1,203 @@
...
+
+    D_802B7C78_E23228->x = (f32)gPlayerStatus.position.x;
```
ethteck, line 112: These f32 casts shouldn't be necessary

src/E225B0.c
```diff
@@ -1,7 +1,203 @@
...
+    D_8010C93C = &func_802B735C_E2290C;
+}
+
+
+typedef struct struct8015A578 {
+    u8 unk_0[0x2];
+    u8 unk_2;
+    u8 unk_3[0x9];
+} struct8015A578;
+extern struct8015A578 D_8015A578;
```
ethteck, line 128: We should just put this into common_structs.h at this point since it's in 2 other files as well as this one

src/E225B0.c
```diff
@@ -1,7 +1,203 @@
...
+    temp_f2 = temp_a0->y;
+    temp_v1 = temp_a0->unk_24;
+    temp_a0->y = temp_f2 + (((playerStatus->position.y + (f32) playerStatus->colliderHeight + 10.0f)
+            - temp_f2) / 1.5f);
+    temp_a0->x = playerStatus->position.x;
+    temp_a0->z = playerStatus->position.z;
+
+    switch (temp_v1) {
+    case 0:
+        if (partnerActionStatus->actionState.b[0] != 0
```
ethteck, line 150: If this is too long to fit in 120 chars, we typically would put the line break after the && and align the p after the && with the p in `if (p`

src/E225B0.c
```diff
@@ -1,7 +1,203 @@
...
+        }
+        if (phi_v0 == 0) {
+            temp_v1_3 = D_802B7C78_E23228;
+            temp_v1_3->unk_24 += 1;
+            return;
+        }
+        break;
+
+    case 1:
+        if ((playerStatus->flags & 0x20) != 0) {
```
ethteck, line 164: For flag checks, you don't need the != 0 and you can get rid of the extra parens after removing it

src/E225B0.c
```diff
@@ -1,7 +1,203 @@
...
+        if (temp_a0->unk_18++ >= 0x10) {
+            temp_a0->scale = 0.36f;
+            temp_a0->unk_24 += 1;
+        }
+        break;
+
+    case 2:
+        temp_a0->scale = 0.57f;
+        temp_a0->unk_24 += 1;
+        sfx_play_sound_at_player(0x17B, 0);
```
ethteck, line 178: in enums.h there's a sound effect enum - if this one exists, please use the value there, and if not, please create one for 0x17B

src/E225B0.c
```diff
@@ -1,7 +1,203 @@
...
+        break;
+
+    case 3:
+        temp_a0->scale = 0.53f;
+        if (temp_a0->unk_18 >= 0x2F || (playerStatus->flags & 0x20) != 0) {
```
ethteck, line 183: flag check can remove the != 0 and parens

src/E225B0.c
```diff
@@ -1,7 +1,203 @@
...
+        }
+
+        if (D_802B7C78_E23228->unk_18++ >= 0x33) {
+            D_8015A578.unk_2 = 0;
+            D_8010C93C = NULL;
+            playerStatus->animFlags &= ~0x100;
+        }
+        break;
+
+    default:
```
ethteck, line 198: The default case here seems unnecessary - try removing it?

ethteck (changes requested):
Hi, great first PR and welcome to the project! I made a bunch of comments, so please feel free to ask questions on anything I mentioned. I may have more after this but this should be most of it :)

src/E225B0.c
eperdew: `sizeof(*D_802B7C78_E23228)` but yup.

src/E225B0.c
eperdew: I will rewrap my code to 120 chars, but got it.

src/E225B0.c
eperdew: Refactor for this will be in the next commit.

eperdew:
Okay, I believe everything should be addressed in my latest commit. Is there a style guide/autoformatter that you guys use that I can get set up?

include/common_structs.h
```diff
@@ -2317,4 +2317,12 @@ typedef struct unkPartnerStruct {
...
 } unkPartnerStruct; // size = 0x1C
+
+typedef struct struct8015A578 {
+    /* 0x00 */ u8 unk_00;
+    /* 0x01 */ u8 unk_01;
+    /* 0x02 */ u8 unk_02;
+    /* 0x03 */ u8 unk_03[5];
+    /* 0x08 */ f32 unk_08;
+} struct8015A578; // size = 0x0C
 #endif
```
nanaian, line 2328: Style nit: leave a empty line between declarations and the endif:
```suggestion
} struct8015A578; // size = 0x0C

#endif
```

src/E225B0.c
```diff
@@ -1,7 +1,186 @@
...
+        guScaleF(matrix1, D_802B7C78_E23228->scale, D_802B7C78_E23228->scale,
+            D_802B7C78_E23228->scale);
+        guRotateF(matrix2, -gCameras[gCurrentCameraID].currentYaw, 0.0f, 1.0f, 0.0f);
+        guMtxCatF(matrix1, matrix2, matrix1);
+        guTranslateF(matrix2, D_802B7C78_E23228->x, D_802B7C78_E23228->y, D_802B7C78_E23228->z);
+        guMtxCatF(matrix1, matrix2, matrix2);
+        guMtxF2L(matrix2, (Mtx*) &gDisplayContext->matrixStack[gMatrixListPos]);
+
+        gSPMatrix(gMasterGfxPos++, &gDisplayContext->matrixStack[gMatrixListPos++], 3);
+        gSPDisplayList(gMasterGfxPos++, (u32) &D_802B7C00_E231B0);
```
nanaian, line 64: This cast is weird; shouldn't `D_802B7C00_E231B0` be a `Gfx[]`? Then you can reference it like
```suggestion
        gSPDisplayList(gMasterGfxPos++, D_802B7C00_E231B0);
```

src/E225B0.c
```diff
@@ -1,7 +1,186 @@
...
+
+        localD_802B7C78_E23228 = D_802B7C78_E23228;
+        if (localD_802B7C78_E23228->unk_18 < 0x2F) {
```
nanaian, line 67: I would use decimal for this constant

src/E225B0.c
```diff
@@ -1,7 +1,186 @@
...
+        command.unk_10 = 0xFF;
+
+        func_8013AF70(0, &command, 0, &matrix2);
+        gSPPopMatrix(gMasterGfxPos++, 0);
+    }
+}
+
+
+void func_802B72C0_E22870(void) {
+    mem_clear((s8*)D_802B7C78_E23228, sizeof(*D_802B7C78_E23228));
```
nanaian, line 109: We should definitely retype `mem_clear` to take a `void*` as its first arg so we don't have to do this cast. (This would also match [bzero](https://man7.org/linux/man-pages/man3/bzero.3.html), which is what the func appears to be.)

nanaian (changes requested):
Thanks for the contributions!! Awesome matches, especially considering the undocumented funcs you tackle here 🚀

eperdew:
Should all be addressed.

src/EB5130.c
```diff
@@ -7,4 +7,9 @@
 
 #include "world/common/DeadUnkTexturePanFunc2.inc.c"
 
-INCLUDE_ASM(s32, "EB5130", func_8024030C_EB543C);
+extern s16 D_80169B12;
+
+ApiStatus func_8024030C_EB543C(Evt* script, s32 isInitialCall) {
+    evt_set_variable(script, *script->ptrReadPos, D_80169B12);
+    return ApiStatus_DONE2;
+}
```
Brotenko, line 15: There are 3 duplicates of these functions (including one that has already been decompiled `func_80240388_EA7358` in `EA7320.c`) so deduplicating those will probably be a thing I will do next, especially since they all access `D_80169B12`. You can keep it like this though, that is something I have to worry about!

Brotenko (approved):

PR #436, opened 2021-09-26
Decompile EB1170.

eperdew:
To get EB1170.c to match, it was necessary to move EB1920.data.s to
a the .data section of EB1170.c.

I put the EB1920 data in a separate file organizationally, but that involved including a .c. Let me know if they should be merged instead.


BowserSlug:
🚀🚀🚀🚀🚀🚀🚀🚀🚀 This PR matches 9 functions (+0.03%) on `us`.

⚠️ This PR introduces 174 warnings:
See log for details.

src/EB1170.c
```diff
@@ -36,15 +129,49 @@ ApiStatus func_802406E0_EB1540(Evt* script, s32 isInitialCall) {
...
+    event->varTable[8] = evt_get_variable(event, *readPos++);
+    event->varTable[9] = evt_get_variable(event, *readPos++);
+    event->varTable[10] = evt_get_variable(event, *readPos++);
+    event->varTable[11] = evt_get_variable(event, *readPos++);
+    return ApiStatus_DONE2;
+}
+
+
 
 //INCLUDE_ASM(s32, "EB1170", func_8024086C_EB16CC);
```
ethteck, line 151: I know you didn't do this in your PR, but that include_asm comment can be removed. I'll let brotenko review the rest of this

eperdew: Removed locally - I'll include it with the commit for the rest of brotenko's feedback.

Brotenko (commented):
Theoretically this is pretty amazing work for a first/second PR, but due to this being code of a dead map, a lot of quirks and edgecases apply here that wouldn't apply in other parts of the codebase. I could theoretically navigate you through the process of fixing it all and making it conform with how we do dead functions, but since you said that you'd rather stick to the relevant code, I'd suggest closing the PR instead so I can fix/rearrange/refactor it on a clean branch instead.

Nothing against the work you have done, seriously good work, but naming the dead functions/symbols, defininf them in dead.h correctly, deduplicating a bunch of these functions (which is especially tedious), and migrating the data correctly would take a bunch of time and revert a lot of what you have done anyway. I also made a local copy of the two .c files so I can reference your work later one, so it hasn't been in vein eitherway! 

Sorry for the inconvenience, but I honestly wouldn't want to drag you through this whole process unless you really want to!

eperdew:
Thanks Brotenko.

PR #446, opened 2021-09-28
Decompile func_80218808_5B1CB8, area_arn/4550D0.c, and area_dig/6DE1F0.c.

eperdew:
The match for func_80218808_5B1CB8 ended up being a little strange. Also I wasn't sure whether to use dec or hex for the constants. It seems like the `actorType` field is bit packed and we might be missing some macros, based on the other uses of it.


BowserSlug:
🚀🚀🚀 This PR matches 3 functions (+0.02%) on `us`.

src/battle/area_arn/4FF0D0.c
```diff
@@ -2,4 +2,25 @@
...
-INCLUDE_ASM(s32, "battle/area_arn/4FF0D0", func_80218680_4FF0D0);
+ApiStatus func_80218680_4FF0D0(Evt* script, s32 isInitialCall) {
+    Bytecode *readPos;
+    s32 var1;
+    s32 var2;
+    s32 var3;
+    f32 var4;
+    Temp8025D160* temp8025D160;
+
+    readPos = script->ptrReadPos;
```
ethteck, line 13: We call `readPos` `args`, and you can initialize it on line 6 where it's declared too. You can the initialize the temp struct pointer as well on the following line, and vars 1-4 on the following lines where they're declared

src/battle/area_dig/6DE1F0.c
```diff
@@ -2,6 +2,42 @@
...
+    f32 angle;
+
+    s32 destBytecode1;
+    s32 destBytecode2;
+    s32 destBytecode3;
+
+    // Seems to be necessary to get the stack allocations to line up.
+    s8 unused[0x40];
+
+    s32* readPos = script->ptrReadPos;
```
ethteck, line 18: args

src/battle/area_dig/6DE1F0.c
```diff
@@ -2,6 +2,42 @@
...
+
+    angle = (f64)evt_get_variable(script, *readPos++) / 10.0f;
+    angle = angle * TAU;
+    angle = angle / 360.0f;
+
+    destBytecode1 = *readPos++;
+    destBytecode2 = *readPos++;
+    destBytecode3 = *readPos++;
+
+    y = sin_rad(angle) * 150.0f + 0.0f;
```
ethteck, line 28: I think the names of y and x should be switched here

src/battle/area_flo/5B1A10.c
```diff
@@ -14,4 +14,48 @@
 
 #include "common/StartRumbleWithParams.inc.c"
 
-INCLUDE_ASM(s32, "battle/area_flo/5B1A10", func_80218808_5B1CB8);
+ApiStatus func_80218808_5B1CB8(Evt* script, s32 isInitialCall) {
+    Actor* actor;
+    s32 actorType;
+    s32 value;
+
+    Bytecode* readPos = script->ptrReadPos;
```
ethteck, line 22: `args` and this is usually the first line as well

src/battle/area_flo/5B1A10.c
```diff
@@ -14,4 +14,48 @@
...
+    ActorID actorId = evt_get_variable(script, *readPos++);
+
+    if (actorId == -0x7F) {
```
ethteck, line 25: this is checking if the actor id is self. see BindNextTurn for an example (and please call these `actorID`)

src/battle/area_flo/5B1A10.c
```diff
@@ -14,4 +14,48 @@
...
+
+    // These gotos are weird, but they produce the least weird looking match
+    // that I have found. Alternatives are an
+    //    evt_set_variable(script, *readPos, 0);
+    // in every possible else of every if, or a top level switch with breaks
+    // instead of gotos.
+    //
+    // Likewise, combining the conditions in any way will break the compiler's
+    // choice of the bnezl instruction.
+    if (actorType != 0x47) {
```
ethteck, line 40: This is almost certainly a switch over actorType

ethteck (changes requested):
I want to make sure we can get rid of the gotos before merging this in. I can take a look at it today

ethteck:
but good job on the functions!

src/battle/area_dig/6DE1F0.c
eperdew: I matched y/x to sin/cos, but yeah it seemed fishy given that it stored y then x later. Will do.

src/battle/area_flo/5B1A10.c
eperdew: That was my first thought as well. I'll play around with it some more and see if I can get a clean match with a switch.

eperdew:
I'll address the style changes/small things real quick, and see if I can get a match with a switch after.

ethteck:
Awesome, thanks. I'm taking a look as well. Also, I noticed that at least the last function is also present in 18 other c files:
`./tools/find_duplicates.py func_80218808_5B1CB8`

It would be ideal here to create an .inc.c file for this in world/common and include it from all the relevant maps. I can explain more if you want. are you on our discord?

src/battle/area_flo/5B1A10.c
eperdew: I got a janky match by doing the following:

```c
    switch (actorType) {
        case 0x4D: case 0x4E: case 0x4F: case 0x50:
        case 0x51: case 0x52: case 0x53: case 0x54:
        case 0x55: case 0x56: case 0x57: case 0x58:
        case 0x59: case 0x5A: case 0x5B: case 0x5C:
        case 0x5D: case 0x5E: case 0x5F: case 0x60:
        case 0x61: case 0x62: case 0x63: case 0x64:
            evt_set_variable(script, *args, -1);
            break;

        case 0x47:
            if (actor->varTable[8] == 1) {
                evt_set_variable(script, *args, -1);
                break;
            }
        default:
            evt_set_variable(script, *args, 0);
    }
```

Any thoughts? Seems a bit better.

ethteck: I think that's better, yeah! although it does take up a lot of space, could you put each case statement on a separate line?

src/battle/area_dig/6DE1F0.c
ethteck: good point. I'm not sure..

src/battle/area_arn/4FF0D0.c
```diff
@@ -2,4 +2,24 @@
 
 #define NAMESPACE b_area_arn
 
-INCLUDE_ASM(s32, "battle/area_arn/4FF0D0", func_80218680_4FF0D0);
+ApiStatus func_80218680_4FF0D0(Evt* script, s32 isInitialCall) {
+    Bytecode *args = script->ptrReadPos;
+    s32 var1;
```
ethteck, line 7: TempBlah can be declared here next and initialized on the same line. and then you can initialize vars 1, 2, 3, 4 on the same lines they're declared

eperdew: Will be in the next commit.

ethteck: ah kk, sorry - thought you had intended this to be resolved already

nanaian (approved):

eperdew:
Okay, actor type stuff is in the latest commit.

src/battle/area_flo/5B1A10.c
eperdew: A better version of this is in the final commit.

ethteck (approved):

PR #447, opened 2021-09-28
Decomp most of the pause badge menu

Dalrik:
Decomp everything on the pause badge menu except for `pause_badges_draw_contents`

`pause_badges_draw_contents` should be coming soon in a future PR, it's just big and taking a while and I don't want to wait any more on the rest of this


BowserSlug:
🚀🚀🚀🚀🚀🚀🚀🚀 This PR matches 8 functions (+0.08%) on `us`.

include/enums.h
```diff
@@ -1746,4 +1753,10 @@ enum EffectGfxDataFlags {
     FX_GRAPHICS_2 = 2,
 };
 
+typedef u8 MoveID;
```
ethteck, line 1756: Thanks for making this enum! I'm not sure we want / should be using the typedefs, because things are inconsistently typed sometimes, and they can make things more complicated than they're worth. Thoughts @nanaian ?

src/pause/135EE0.c
```diff
@@ -148,7 +148,19 @@ INCLUDE_ASM(s32, "pause/135EE0", pause_handle_input);
...
 INCLUDE_ASM(s32, "pause/135EE0", pause_cleanup);
 
-INCLUDE_ASM(s32, "pause/135EE0", pause_get_total_equipped_bp_cost);
+s32 pause_get_total_equipped_bp_cost(void) {
+    s32 totalCost = 0;
+    s32 i;
+
+    for (i = 0; i < ARRAY_COUNT(gPlayerData.equippedBadges); i++) {
+        ItemID itemID = gPlayerData.equippedBadges[i];
+        if (itemID != ITEM_NONE) {
```
ethteck, line 157: Empty line before the if would be good if you don't mind

src/pause/138CC0.c
```diff
@@ -103,19 +103,386 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+    for (i = 0; i < ARRAY_COUNT(gPlayerData.equippedBadges); ++i, ++currentSlot) {
+        if (badgeID == *currentSlot) {
+            slotToRemove = currentSlot;
+            result = 1;
+            break;
+        }
+    }
+
+    // Remove and shift array contents
+    for (; i < ARRAY_COUNT(gPlayerData.equippedBadges)-1; ++i, ++currentSlot) {
```
ethteck, line 125: spaces around ` - ` and we do postincrement for loop iterators (and pretty much everywhere in the code)

src/pause/138CC0.c
```diff
@@ -103,19 +103,386 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+        *currentSlot = *(currentSlot+1);
```
ethteck, line 126: you should be able to do *currentSlot++ (parens may be necessary around *currentSlot - I can't remember here)

src/pause/138CC0.c
```diff
@@ -103,19 +103,386 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+        badgeID = 0;
+    }
+    if (badgeID == 0x7FFF) {
+        badgeID = 0;
+    }
+    if (badgeID == 0) {
+        return EQUIP_RESULT_NOT_ENOUGH_BP;
+    }
+
+    for (i = 0; i < ARRAY_COUNT(playerData->equippedBadges); ++i, ++badgeSlot) {
```
ethteck, line 160: `i++, badgeSlot++` - also, have you tried not using badgeSlot + co and just using thing[i] syntax for these loops?

src/pause/138CC0.c
```diff
@@ -103,19 +103,386 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+            return EQUIP_RESULT_ALREADY_EQUIPPED;
+        }
+    }
 
-INCLUDE_ASM(s32, "pause/138CC0", pause_badges_try_equip);
+    totalEquippedBP = pause_get_total_equipped_bp_cost();
+    if (badgeID != 0) {
+        u8 moveID = gItemTable[badgeID].moveID;
+        s32 requiredBP = totalEquippedBP + gMoveTable[moveID].costBP;
+        if (playerData->maxBP < requiredBP) {
```
ethteck, line 170: Empty line before this if please

src/pause/138CC0.c
```diff
@@ -103,19 +103,386 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+            return EQUIP_RESULT_NOT_ENOUGH_BP;
+        }
+    }
 
-INCLUDE_ASM(s32, "pause/138CC0", pause_badges_draw_bp_orbs);
+    badgeSlot = &playerData->equippedBadges[0];
+    for (i = 0; i < ARRAY_COUNT(playerData->equippedBadges); ++i, ++badgeSlot) {
```
ethteck, line 176: i++, etc

src/pause/138CC0.c
```diff
@@ -103,19 +103,386 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+    if (i == ARRAY_COUNT(playerData->equippedBadges)) {
+        return EQUIP_RESULT_TOO_MANY_BADGES;
+    } else {
+        return EQUIP_RESULT_SUCCESS;
+    }
+}
+
+void pause_badges_draw_bp_orbs(s32 orbState, s32 x, s32 y) {
+    s32 orbSize = 8;
+    switch (orbState) {
```
ethteck, line 191: Empty line before switch

src/pause/138CC0.c
```diff
@@ -103,19 +103,386 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+        equippedBadges = playerData->equippedBadges;
+        for (i = 0; i < ARRAY_COUNT(playerData->equippedBadges); ++i) {
+            // This can be written more clearly as equippedBadges[i],
+            // but that causes some instruction reordering
+            s16 badgeItemID = *equippedBadges;
+            if (badgeItemID != 0) {
+                gBadgeMenuItemIDs[numItems] = badgeItemID;
+                numItems += 1;
+            }
+            ++equippedBadges;
```
ethteck, line 237: thing++

src/pause/138CC0.c
```diff
@@ -103,19 +103,386 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+    gBadgeMenuSelectedIndex = 0;
+    gBadgeMenuSelectedItemID = 0;
+    D_8027037C = 0;
+    D_80270394 = 0;
+    gBadgeMenuCurrentPage = 0;
+
+    page = &gBadgeMenuPages[0];
+    i = 0;
+   
+    while (1) {
```
ethteck, line 258: This should be able to be re-written as a for loop or at least while loop

src/pause/138CC0.c
```diff
@@ -103,19 +103,386 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+        page->numRows = 8;
+        page->enabled = 1;
+        page->startIndex = i*8;
+        page->count = 8;
+
+        i += 1;
+        page += 1;
+    }
+
+    if (((u16) gBadgeMenuNumItems & 7) != 0) {
```
ethteck, line 275: This doesn't need the != 0 since it's a flag check. though it may actually be a mod % 8 operation..?

src/pause/138CC0.c
```diff
@@ -103,19 +103,386 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+        page->listStart = i * 8;
+        page->numCols = 1;
+        page->enabled = 1;
+        page->startIndex = i * 8;
+
+        menuNumItems = gBadgeMenuNumItems;
+        count = menuNumItems - (menuNumItems / 8 * 8);
+        page->count = count;
+        page->numRows = page->count;
+        i += 1;
```
ethteck, line 287: i++;

src/pause/138CC0.c
```diff
@@ -103,19 +103,386 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+        ++page;
+    }
+
+    gBadgeMenuTargetScrollPos = gBadgeMenuCurrentScrollPos = pause_badges_get_pos_y(0, 0);
+}
+
+void pause_badges_init(MenuPanel *panel) {
+    s32 i;
+
+    (void)pause_badges_count_all();
```
ethteck, line 304: Is there a reason the void cast is here? I think it can be removed

src/pause/138CC0.c
```diff
@@ -103,19 +103,386 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+            case EQUIP_RESULT_TOO_MANY_BADGES:
+                sfx_play_sound(SOUND_MENU_BADGE_ERROR);
+                gBadgeMenuBShowNotEnoughBP = 2;
+                return;
+            case EQUIP_RESULT_SUCCESS:
+                sfx_play_sound(SOUND_MENU_BADGE_EQUIP);
+                return;
+            default:
+                return;
+        }
```
ethteck, line 369: You should be able to just put a return statement after the switch (right before the end of this block) and change all the returns to breaks and remove the default case

src/pause/138CC0.c
```diff
@@ -103,19 +103,386 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+    if (gPauseMenuPressedButtons & BUTTON_A) {
+        s16 badgeID;
+        if (gBadgeMenuBShowNotEnoughBP != 0) {
+            gBadgeMenuBShowNotEnoughBP = 0;
+            return;
+        }
+        if (gBadgeMenuLevel == 0) {
+            if (gBadgeMenuItemIDs[selectedIndex] == 0x7FFE) {
+                sfx_play_sound(SOUND_MENU_BADGE_ERROR);
+                return;
```
ethteck, line 344: This can be an if/else

src/pause/138CC0.c
```diff
@@ -103,19 +103,386 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+
+    s32 selectedCol = selectedIndex % numCols;
+    s32 selectedRow = selectedIndex / numCols;
+
+    if (gPauseMenuPressedButtons & BUTTON_A) {
+        s16 badgeID;
+        if (gBadgeMenuBShowNotEnoughBP != 0) {
+            gBadgeMenuBShowNotEnoughBP = 0;
+            return;
+        }
```
ethteck, line 340: instead of the return above, you can do else if (gBadgeMenuLevel == 0)

src/pause/138CC0.c
```diff
@@ -103,19 +103,386 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+        if (gBadgeMenuLevel == 0) {
+            if (gBadgeMenuItemIDs[selectedIndex] == 0x7FFE) {
+                sfx_play_sound(SOUND_MENU_BADGE_ERROR);
+                return;
+            }
+            gBadgeMenuLevel = 1;
+            sfx_play_sound(SOUND_MENU_NEXT);
+            return;
+        }
```
ethteck, line 349: instead of returns here, you can do } else {

src/pause/138CC0.c
```diff
@@ -103,19 +103,386 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+    }
+
+    if ((gPauseMenuPressedButtons != 0) || (gPauseMenuHeldButtons != 0)) {
+        gBadgeMenuBShowNotEnoughBP = 0;
+    }
+
+    if (gBadgeMenuNumItems != 0) {
+        if (gBadgeMenuLevel == 0) {
+            s32 oldTab = gBadgeMenuCurrentTab;
+            if (gPauseMenuHeldButtons & (BUTTON_STICK_UP | BUTTON_STICK_DOWN)) {
```
ethteck, line 379: Empty line after decls

src/pause/138CC0.c
```diff
@@ -103,19 +103,386 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+                    selectedRow = (s32) gBadgeMenuPages[gBadgeMenuCurrentPage].listStart;
+                }
+            }
+
+            // Need to re-read button state here for strange regalloc reasons
+            heldButtons2 = gPauseMenuHeldButtons;
+            if (heldButtons2 & (BUTTON_STICK_DOWN | BUTTON_R)) {
+                if (heldButtons2 & BUTTON_STICK_DOWN) {
+                    PauseItemPage* page = &gBadgeMenuPages[gBadgeMenuCurrentPage];
+                    selectedRow += 1;
```
ethteck, line 415: Empty line after decl

src/pause/138CC0.c
```diff
@@ -103,19 +103,386 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+                    if (selectedRow >= (page->listStart + page->numRows)) {
+                        gBadgeMenuCurrentPage += 1;
+                        if (gBadgeMenuPages[gBadgeMenuCurrentPage].enabled == 0) {
```
ethteck, line 418: == FALSE

src/pause/138CC0.c
```diff
@@ -103,19 +103,386 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+                            gBadgeMenuCurrentPage -= 1;
+                            selectedRow -= 1;
+                        }
+                    }
+                } else {
+                    // R button press
+                    PauseItemPage* newPage = &gBadgeMenuPages[++gBadgeMenuCurrentPage];
```
ethteck, line 425: thing++

src/pause/138CC0.c
```diff
@@ -103,19 +103,386 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+        } else {
+            gPauseMenuCurrentDescMsg = 0;
+            gPauseMenuCurrentDescIconScript = NULL;
+        }
+    } else {
+        gPauseMenuCurrentDescMsg = pause_get_menu_msg(gBadgeMenuCurrentTab == 0 ? 0x4C : 0x4D);
+        gPauseMenuCurrentDescIconScript = NULL;
+    }
+
+    if ((gPauseMenuPressedButtons & BUTTON_B) != 0) {
```
ethteck, line 472: no need for != 0 here

src/pause/138CC0.c
```diff
@@ -103,19 +103,386 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+            if (gBadgeMenuSelectedIndex != selectedIndex) {
+                sfx_play_sound(SOUND_MENU_CHANGE_SELECTION);
+            }
+            gBadgeMenuSelectedItemID = gBadgeMenuItemIDs[gBadgeMenuSelectedIndex];
+        }
+    }
+
+    if (gBadgeMenuLevel == 1) {
+        s32 itemID = gBadgeMenuSelectedItemID;
+        if (((itemID != 0x7FFE) && (itemID != 0x7FFF) && (itemID != 0))) {
```
ethteck, line 461: we should make defines for these 0x7FFE and 0x7FFF vals since they're everywhere...not sure what they are though. @nanaian any thoughts?

src/pause/138CC0.c
```diff
@@ -142,4 +509,9 @@ void pause_badges_update(void) {
     *currentScrollPos += pause_interp_vertical_scroll(gBadgeMenuTargetScrollPos - *currentScrollPos);
 }
 
-INCLUDE_ASM(s32, "pause/138CC0", pause_badges_cleanup);
+void pause_badges_cleanup(void) {
+    s32 i;
+    for (i = 0; i < ARRAY_COUNT(gBadgeMenuIconIDs); ++i) {
```
ethteck, line 514: Empty line before for, i++

ethteck (changes requested):
Great job with these!

src/pause/138CC0.c
Dalrik: 0x7FFE and 0x7FFF are hinted at more by the draw function (not included here). `0x7FFE` is a fake item ID used for the "None" line if you select "Equipped Badges" but have none equipped. 0x7FFF is empty slot, looks like checks for it are mostly defensive coding.

I can add macros for `BADGE_NONE` and `BADGE_INVALID`

src/pause/138CC0.c
Dalrik: `*currentSlot++` would not be well formed here, iirc in C there's no sequence point between expressions on the left and right side of an equals sign. It wouldn't be well defined whether the RHS uses the incremented or not incremented version of `currentSlot`.

Regardless, the increment being in the for loop header does actually matter, it changes where the increment operation gets emitted in the asm (I had to massage this loop a bunch to get it right)

src/pause/138CC0.c
Dalrik: Force of habit, I usually include it when a function's return value is being discarded. In this case it looks like they call count_all for its side effect of sorting the badge list, rather than to actually get the badge count

src/pause/138CC0.c
Dalrik: Yeah I have, the obvious `badgeSlot[i]` messes with register allocation a bit (relocates `badgeSlot` from `a1` to `v1`, probably something to do with pushing the final write to `badgeSlot` farther up in the function), I couldn't find a better way to fix it.

src/pause/135EE0.c
Dalrik: Done

src/pause/138CC0.c
Dalrik: Done

src/pause/138CC0.c
Dalrik: preincrement -> postincrement is done

src/pause/138CC0.c
Dalrik: Done

src/pause/138CC0.c
Dalrik: Done

src/pause/138CC0.c
Dalrik: Done

src/pause/138CC0.c
Dalrik: Done

src/pause/138CC0.c
Dalrik: Turns out it is mod 8

include/enums.h
nanaian: Agreed, no need to bother with the typedef. I'm fairly sure we're going to be getting rid of most of them soon anyway

src/pause/138CC0.c
```diff
@@ -103,19 +108,393 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+    return result;
+}
+
+typedef s32 BadgeEquipResult;
+enum BadgeEquipResults {
+    EQUIP_RESULT_ALREADY_EQUIPPED,
+    EQUIP_RESULT_NOT_ENOUGH_BP,
+    EQUIP_RESULT_TOO_MANY_BADGES,
+    EQUIP_RESULT_SUCCESS,
+};
```
nanaian, line 147: Should be able to do this:
```suggestion
typedef enum BadgeEquipResult {
    EQUIP_RESULT_ALREADY_EQUIPPED,
    EQUIP_RESULT_NOT_ENOUGH_BP,
    EQUIP_RESULT_TOO_MANY_BADGES,
    EQUIP_RESULT_SUCCESS,
} BadgeEquipResult;
```
Good enum btw :+1:

src/pause/138CC0.c
ethteck: kk thanks!

src/pause/138CC0.c
ethteck: sorry I should have said to just do `!` instead of == 0. my bad

src/pause/138CC0.c
```diff
@@ -103,19 +108,393 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+
+    switch (orbState) {
+        case 0:
+            pause_draw_rect(x * 4, y * 4, (x + orbSize) * 4, (y + orbSize) * 4, 0, 0, 0x100, 0x400, 0x400);
+            break;
+        case 1:
+            pause_draw_rect(x * 4, y * 4, (x + orbSize) * 4, (y + orbSize) * 4, 0, 0, 0, 0x400, 0x400);
+            break;
+        default:
+            pause_draw_rect(x * 4, y * 4, (x + orbSize) * 4, (y + orbSize) * 4, 0, 0, 0x200, 0x400, 0x400);
```
nanaian, line 206: Given the clipping args (the last 4 args of pause_draw_rect) are in pixels, I'd prefer them in decimal rather than hex

ethteck:
Thanks for all your changes! I re-opened one thing cause I was bieng silly and forgot we typically do ! for checks of a bool to be false

src/pause/138CC0.c
```diff
@@ -103,19 +108,393 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+    s32 i;
+
+    D_80270388 = 0;
+    if (!onlyEquipped) {
+        for (i = 0; i < ARRAY_COUNT(playerData->badges); i++) {
+            s16 badgeItemID = playerData->badges[i];
+
+            if (badgeItemID == 0) {
+                continue;
+            } else if (badgeItemID >= ITEM_PARTNER_ATTACK) {
```
nanaian, line 226: Worth turning this into a LAST_BADGE macro?

src/pause/138CC0.c
```diff
@@ -103,19 +108,393 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+
+    page = &gBadgeMenuPages[0];
+    i = 0;
+   
+    for (i = 0; i < gBadgeMenuNumItems / 8; i++, page++) {
+        page->listStart = i*8;
+        page->numCols = 1;
+        page->numRows = 8;
+        page->enabled = 1;
+        page->startIndex = i*8;
```
nanaian, line 275:
```suggestion
        page->listStart = i * 8;
        page->numCols = 1;
        page->numRows = 8;
        page->enabled = TRUE;
        page->startIndex = i * 8;
```

src/pause/138CC0.c
```diff
@@ -103,19 +108,393 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+        count = menuNumItems - (menuNumItems / 8 * 8);
+        page->count = count;
+        page->numRows = page->count;
+        i++;
+        page++;
+    }
 
-INCLUDE_ASM(s32, "pause/138CC0", pause_badges_init);
+    while (i < ARRAY_COUNT(gBadgeMenuPages)) {
+        page->enabled = 0;
```
nanaian, line 297:
```suggestion
        page->enabled = FALSE;
```

src/pause/138CC0.c
```diff
@@ -103,19 +108,393 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+        page->count = 8;
+    }
+
+    if ((gBadgeMenuNumItems % 8) != 0) {
+        s16 count;
+        s32 menuNumItems;
+
+        page->listStart = i * 8;
+        page->numCols = 1;
+        page->enabled = 1;
```
nanaian, line 285:
```suggestion
        page->enabled = TRUE;
```

src/pause/138CC0.c
```diff
@@ -103,19 +108,393 @@ s32 pause_badges_scroll_offset_x(s32 x) {
...
+    gBadgeMenuCurrentPage = 0;
+
+    page = &gBadgeMenuPages[0];
+    i = 0;
+   
+    for (i = 0; i < gBadgeMenuNumItems / 8; i++, page++) {
+        page->listStart = i*8;
+        page->numCols = 1;
+        page->numRows = 8;
+        page->enabled = 1;
```
nanaian, line 274:
```suggestion
        page->enabled = TRUE;
```

nanaian (changes requested):
TYSM for this

src/pause/138CC0.c
Dalrik: I'm not sure `LAST_BADGE` is the correct name for that, all of the items and key items, and hammers/boots come earlier as well (and are technically equippable by this code lul). Badges start at `ITEM_SPIN_SMASH` (224).

Maybe it's just an oversight that they don't also check for `FIRST_BADGE`, and we should throw that in too for `ITEM_SPIN_SMASH`

src/pause/138CC0.c
Dalrik: That actually causes a mismatch unless I introduce a cast to `s32` in the switch in `pause_badges_handle_input`. It looks like the optimizer notices that there's only 4 valid enum values and gets rid of checks for <0 and >3 in the switch conditions.

Maybe it'd be better for `pause_badges_try_equip` to return s32 and still drop the typedef

Dalrik:
While adding macros for `ITEM_FIRST_BADGE` and `ITEM_LAST_BADGE`, I also went ahead and added them for `KEY_ITEM` and `REGULAR_ITEM`

ethteck (approved):

PR #448, opened 2021-09-29
Match get_map_IDs_by_name

JoshDuMan:
https://decomp.me/scratch/6xc2J


BowserSlug:

ethteck (changes requested):
So we don't forget about it, could you leave a comment saying like `TODO: likely fake match`

nanaian (approved):

nanaian:
The match looks valid to me w.r.t the for condition; it's a null-terminated array so it makes sense

JoshDuMan:
I added the comment, but no guarantees that this isn't the answer. Although it's obviously bad, I don't know what you can do to avoid it. The compiler insists using that temp for the conditional, but it also needs set each loop iteration.

ethteck:
Thanks. The reason I think it's at least somewhat fake is that they're using a temp for gAreas[i].maps but not gAreas[i]. the latter temp would make more sense, since later in the loop they have to access something else from gAreas[i] anyway. We can revisit and remove the comment later though if no one can find something more sensible

PR #453, opened 2021-09-30
Decomp pause_sort_item_list and pause_draw_rect

Dalrik:
pause_sort_item_list is a very ugly match, I feel like the original code would have been based on indices instead of pointer arithmetic, especially given the optimization behavior around the do/while loop (I think the second half of the condition might avoid optimization if it's looking at indices), but I couldn't find anything that works. https://decomp.me/scratch/TRZek if you want to explore.

This also has close matches on two other functions in 135EE0, only issue with each is delay slot related and would probably be fixed by adding a data section.


BowserSlug:
🚀🚀 This PR matches 2 functions (+0.02%) on `us`.

src/pause/135EE0.c
```diff
@@ -22,9 +22,80 @@ void pause_set_cursor_opacity(s32 val) {
...
+            || get_game_mode() == GAME_MODE_EXIT_FILE_SELECT
+            || get_game_mode() == GAME_MODE_EXIT_LANGUAGE_SELECT) {
+        if (D_8024EFB4 != 0) {
+            s32 i;
+
+            for (i = 0x16; i < 0x2C; i++) {
+                Window* window = &gWindows[i];
+                s8 parent = window->parent;
+
+                if ((parent == -1 || parent == 0x16) && (window->flags & 8) != 0) {
```
ethteck, line 42: != 0 unneeded for flag check

src/pause/135EE0.c
```diff
@@ -22,9 +22,80 @@ void pause_set_cursor_opacity(s32 val) {
...
+                }
+            }
+            if (i >= 0x2C) {
+                D_8024EFB4 = 0;
+            }
+        }
+        gPauseMenuTargetPosX = gPauseMenuCursorPosX = posX;
+        gPauseMenuTargetPosY = gPauseMenuCursorPosY = posY;
+
+    } else if ((window->flags & 8) == 0 && (window->parent == -1 || (gWindows[window->parent].flags & 8) == 0)) {
```
ethteck, line 53: `!(gWindows[window->parent].flags & 8)`

src/pause/135EE0.c
```diff
@@ -22,9 +22,80 @@ void pause_set_cursor_opacity(s32 val) {
...
+            }
+            if (i >= 0x2C) {
+                D_8024EFB4 = 0;
+            }
+        }
+        gPauseMenuTargetPosX = posX;
+        gPauseMenuCursorPosX = posX;
+        gPauseMenuTargetPosY = posY;
+        gPauseMenuCursorPosY = posY;
+    } else if ((window->flags & 8) == 0 && (window->parent == -1 || (gWindows[window->parent].flags & 8) == 0)) {
```
ethteck, line 91: ! for flag checks

src/pause/135EE0.c
```diff
@@ -164,7 +235,62 @@ s32 pause_get_total_equipped_bp_cost(void) {
     return totalCost;
 }
 
-INCLUDE_ASM(void, "pause/135EE0", pause_draw_rect, s32 ulx, s32 uly, s32 lrx, s32 lry, s32 tileDescriptor, s32 uls,
-            s32 ult, s32 dsdx, s32 dtdy);
+void pause_draw_rect(s32 ulx, s32 uly, s32 lrx, s32 lry, s32 tileDescriptor, s32 uls, s32 ult, s32 dsdx, s32 dtdy) {
+    if (ulx < -2687 || uly < -2687 || lrx <= 0 || lry <= 0) {
```
ethteck, line 239: <= -2688 for these? (so the numbers match the second one)

ethteck (changes requested):

ethteck (approved):

nanaian (approved):

PR #456, opened 2021-10-02
Match pause_stats_init and pause_stats_handle_input

Dalrik:
I'm not a fan of those while(1) loops in `pause_stats_handle_input`, but the cleaner `do..while(panel->selected == ...)` generated hugely different code.

I'm guessing that the pause_badges init and handle_input functions have the same signature as pause_stats' versions, no evidence from callsites though.


BowserSlug:
🚀🚀 This PR matches 2 functions (+0.03%) on `us`.

✅ This PR fixes 1 warnings!

src/pause/138CC0.c
```diff
@@ -1,15 +1,145 @@
...
+                break;
+            }
+        }
+    }
+
+    if (gPauseMenuHeldButtons & BUTTON_STICK_RIGHT) {
+        while (1) {
+            panel->col++;
+            if (panel->col >= panel->numCols) {
+                panel->col = panel->numCols-1;
```
ethteck, line 62: spaces around the minus sign

src/pause/138CC0.c
```diff
@@ -332,7 +462,7 @@ void pause_badges_init(MenuPanel *panel) {
     panel->initialized = TRUE;
 }
 
-void pause_badges_handle_input(void) {
+void pause_badges_handle_input(MenuPanel* panel) {
```
ethteck, line 465: Why did you add an arg here? Is there a call to it that passes an arg?

src/pause/138CC0.c
```diff
@@ -1,15 +1,145 @@
 #include "common.h"
 
+// Probably only used here, but could theoretically be used in the main menu too
+typedef struct {
+    s32 cursorX;
+    s32 cursorY;
+    s32 baseMsgID;
+} StatsEntryData; // size = 0xC
+
+// Probably .data or .bss in this file, not used anywhere else
```
ethteck, line 10: These are data symbols, which you can tell since they were renamed in a .data.s file as well. I don't think we need the comment. We comment bss when it's not handled yet, but this data will be migrated and it works as-defined

include/common_structs.h
```diff
@@ -1894,19 +1894,23 @@ typedef struct PauseMapSpace {
...
-    /* 0x07 */ char unk_07;
-    /* 0x08 */ s8* gridData; // user value at each 2D grid point
+    /* 0x07 */ s8 numPages; // unsure
+    /* 0x08 */ u8* gridData; // user value at each 3D grid point (page, row, col)
     /* 0x0C */ UNK_FUN_PTR(fpInit);
     /* 0x10 */ UNK_FUN_PTR(fpHandleInput);
     /* 0x14 */ UNK_FUN_PTR(fpUpdate);
     /* 0x18 */ UNK_FUN_PTR(fpCleanup);
 } MenuPanel; // size = 0x1C
+#define MENU_PANEL_SELECTED_GRID_DATA(panel) \
```
ethteck, line 1910: I'm not a huge fan of macros like this, but I see why you made it..hmm.. A new line before it would be good though

ethteck (changes requested):

src/pause/138CC0.c
Dalrik: I'm pretty sure the `_handle_input` functions are pointed to by this field in the `MenuPanel` struct, which would imply that they all have the same signature:
```c
    /* 0x10 */ UNK_FUN_PTR(fpHandleInput);
```
`pause_stats_handle_input` has that argument, so presumably `pause_badges_handle_input` has it as well.

ethteck: Ahh yeah okay, that makes sense

include/common_structs.h
Dalrik: I added the space. As I mentioned on discord, this also matches as an inline function (with a better name of course):
```c
static inline u8 MENU_PANEL_SELECTED_GRID_DATA(MenuPanel* panel) {
            return panel->gridData[(panel->page * panel->numCols * panel->numRows) +
                (panel->numCols * panel->row) +
                panel->col];
}
```

nanaian: I like the static inline function, as long as it's a lowercased name

nanaian (approved):

include/common_structs.h
ethteck: We talked in discord about this and agreed it's best to just not have the macro in this case (for now). I think it was good thinking to abstract this kinda repetitive code, but for now let's just have it be a bit bloaty and re-visit it later on.

ethteck (dismissed):
my comment on the macro thing

Dalrik:
Macro has been removed :(

PR #467, opened 2021-10-09
Decompile action_cmd/bomb.c

eperdew:
I also deduplicated a family of HUD rendering functions - they're probably drawing the action command bar for certain action commands.


BowserSlug:
🚀🚀🚀🚀🚀🚀🚀 This PR matches 7 functions (+0.07%) on `us`.

✅ This PR fixes 1 warnings!

eperdew:
Okay, I addressed your feedback, though I left the conversation about NULL'ing out function pointers open because it's weirddddd.

PR #517, opened 2021-11-03
Msg header file

Ponmander:
I thought I might as well start this pull request.
I KNOW THERE ARE ERRORS I NEED HELP.
Please help me it has had me stumped for days I checked all the files that contain the variables/structs that are from this new header file so I would include `msg.h`


BowserSlug:

nanaian (changes requested):
There is data in the header files which should not be there - it'll be redefined for every place the header file is included!

I'd suggest moving all data back to where it was in C files and only defining functions for the time being. It's less error-prone and will avoid mistakenly putting BSS-to-be symbols in header files

Brotenko (changes requested):
You have a bunch of redundant function declarations that result in the ROM not building. You'll have to resolve these in some form or another (either removing them from the fio.h file, or removing them from functions.h, or some mix of those) 
grafik

Ponmander:
Here's the 4th push.

ethteck:
> You have a bunch of redundant function declarations that result in the ROM not building. You'll have to resolve these in some form or another (either removing them from the fio.h file, or removing them from functions.h, or some mix of those) grafik

These are just warnings and aren't related to the rom not building

src/msg.h
```diff
@@ -0,0 +1,99 @@
+#ifndef _MSG_H_
+#define _MSG_H_
+
+#include "common.h"
+
+typedef struct UnkMsgStruct8;
+
+s32 MessagePlural;
+
+s16 MessageSingular;
```
JoshDuMan, line 10: These two are not externed. This may be why the build is failing.

ethteck:
Closing as discussed on discord

PR #518, opened 2021-11-04
Match count_power_plus

JoshDuMan:


BowserSlug:
🚀🚀 This PR matches 2 functions (+0.00%) on `us`.

src/190B20.c
```diff
@@ -1135,7 +1135,26 @@ INCLUDE_ASM(s32, "190B20", func_802636E4);
...
+s32 count_power_plus(s32 arg0) {
+    s32 i;
+    u8 moveID;
+    s32 pp;
+
+    if (gGameStatusPtr->peachFlags & 1) {
+        return 0;
+    }
+
+    for (pp = 0, i = 0; i < ARRAY_COUNT(gPlayerData.equippedBadges); i++) {
```
ethteck, line 1147: I'd prefer pp be set to 0 outside of the loop setup, because it's not really a loop variable in my eyes.. It only gets modified if certain conditions are met, and its scope is beyond the loop. Could we move it to the line before the loop start?

src/190B20.c
```diff
@@ -1135,7 +1135,26 @@ INCLUDE_ASM(s32, "190B20", func_802636E4);
...
+        moveID = gItemTable[gPlayerData.equippedBadges[i]].moveID;
```
ethteck, line 1148: You should be able to define moveID here (u8 moveID ...)

src/190B20.c
```diff
@@ -1135,7 +1135,26 @@ INCLUDE_ASM(s32, "190B20", func_802636E4);
 
 INCLUDE_ASM(s32, "190B20", func_80263914);
 
-INCLUDE_ASM(s32, "190B20", count_power_plus);
+s32 count_power_plus(s32 arg0) {
+    s32 i;
```
ethteck, line 1139: could you put i last? We typically do that lately

ethteck (changes requested):

ethteck (approved):

nanaian (approved):

PR #519, opened 2021-11-05
Match _heap_malloc and _heap_malloc_tail

LightningTH:
I'm fairly certain _heap_malloc should be returning a void* but am unsure how many places assume s32*. It is common for malloc to return a void* that is then cast to the appropriate structure or type when called.


BowserSlug:
⚠️ This PR introduces 2 warnings:
- src/43F0.c:124: warning: assignment from incompatible pointer type
- src/43F0.c:183: warning: assignment from incompatible pointer type
- src/43F0.c:314: warning: passing arg 1 of `nuPiReadRom' makes integer from pointer without a cast
- src/43F0.c:318: warning: passing arg 1 of `nuPiReadRom' makes integer from pointer without a cast
- src/43F0.c:329: warning: passing arg 2 of `dma_write_block' makes integer from pointer without a cast
- src/43F0.c:333: warning: passing arg 2 of `dma_write_block' makes integer from pointer without a cast
- src/43F0.c:344: warning: passing arg 1 of `osWritebackDCache' makes pointer from integer without a cast
- src/43F0.c:349: warning: assignment makes pointer from integer without a cast

src/43F0.c
```diff
@@ -1,7 +1,7 @@
 #include "common.h"
 #include "nu/nusys.h"
 
-s32 D_80074270 = 0;
+u16 D_NEXT_HEAP_MALLOC_ID = 0;
```
ethteck, line 4: We typically name global symbols a "namespace" prefix, followed by the name in camel case, separated by an underscore. so for this I'd probably say `heap_nextMallocID` or something. Also, make sure to find and replace project-wide when renaming symbols, because there are places other than the c files that the symbols exist (symbol_addrs.txt and often undefined_syms.txt )

src/43F0.c
```diff
@@ -74,9 +74,128 @@ HeapNode* _heap_create(s32* addr, u32 size) {
...
+    }
+
+    smallestBlockFound = 0;
+    nextHeapNode = 0;
+
+    //find the smallest block we can fit into in the free list
+    for (curHeapNode = head;;curHeapNode = curHeapNode->next) {
+        if (curHeapNode->allocated == 0) {
+            curBlockLength = curHeapNode->length;
+            if ((curBlockLength >= size) && ((curBlockLength < smallestBlockFound) || (smallestBlockFound == 0))) {
```
ethteck, line 105: The inner sub conditions here don't need parens around them

ethteck (changes requested):
Awesome job, and thanks for your contribution :)

src/43F0.c
```diff
@@ -74,9 +74,128 @@ HeapNode* _heap_create(s32* addr, u32 size) {
...
+
+void* _heap_malloc(HeapNode* head, u32 size) {
+    HeapNode* nextHeapNode;
+    HeapNode* pPrevHeapNode = 0;
+    u32 newBlockSize;
+    u32 curBlockLength;
+    HeapNode* curHeapNode;
+    u32 smallestBlockFound;
+
+    //must allocate 16 bytes or more at minimum or fail
```
nanaian, line 92: Bit of a nit, but for all comments could you put a space after the `//`? i.e.
```suggestion
    // must allocate 16 bytes or more at minimum or fail
```

src/43F0.c
```diff
@@ -74,9 +74,128 @@ HeapNode* _heap_create(s32* addr, u32 size) {
...
+    size = ALIGN16(size);
+    if (size == 0) {
+        return (s32 *)pPrevHeapNode;
```
nanaian, line 95: Is this cast required? Additionally, does it not match as `return NULL`?

src/43F0.c
```diff
@@ -74,9 +74,128 @@ HeapNode* _heap_create(s32* addr, u32 size) {
...
+#define _heap_alloc_and_update_id(node) { \
+    u16 HeapEntryID = D_NEXT_HEAP_MALLOC_ID; \
+    node->allocated = 1; \
+    D_NEXT_HEAP_MALLOC_ID = HeapEntryID + 1; \
+    node->entryID = HeapEntryID; \
+}
+
+void* _heap_malloc(HeapNode* head, u32 size) {
+    HeapNode* nextHeapNode;
+    HeapNode* pPrevHeapNode = 0;
```
nanaian, line 86:
```suggestion
    HeapNode* pPrevHeapNode = NULL;
```

src/43F0.c
```diff
@@ -74,9 +74,128 @@ HeapNode* _heap_create(s32* addr, u32 size) {
...
+    u32 smallestBlockFound;
+
+    //must allocate 16 bytes or more at minimum or fail
+    size = ALIGN16(size);
+    if (size == 0) {
+        return (s32 *)pPrevHeapNode;
+    }
+
+    smallestBlockFound = 0;
+    nextHeapNode = 0;
```
nanaian, line 99:
```suggestion
    nextHeapNode = NULL;
```

src/43F0.c
```diff
@@ -74,9 +74,128 @@ HeapNode* _heap_create(s32* addr, u32 size) {
...
+
+    //find the smallest block we can fit into in the free list
+    for (curHeapNode = head;;curHeapNode = curHeapNode->next) {
+        if (curHeapNode->allocated == 0) {
```
nanaian, line 103:
```suggestion
        if (!curHeapNode->allocated) {
```

src/43F0.c
```diff
@@ -74,9 +74,128 @@ HeapNode* _heap_create(s32* addr, u32 size) {
...
+            //update previous to the proper size for the block being returned
+            pPrevHeapNode->next = (HeapNode *)((u8 *)pPrevHeapNode + newBlockSize);
+            pPrevHeapNode->length = size;
+            _heap_alloc_and_update_id(pPrevHeapNode);
+
+            //setup the new heap block entry
+            curHeapNode = pPrevHeapNode->next;
+            curHeapNode->next = nextHeapNode;
+            curHeapNode->length = smallestBlockFound - newBlockSize;
+            curHeapNode->allocated = 0;
```
nanaian, line 132:
```suggestion
            curHeapNode->allocated = FALSE;
```

src/43F0.c
```diff
@@ -74,9 +74,128 @@ HeapNode* _heap_create(s32* addr, u32 size) {
     }
 }
 
-INCLUDE_ASM(s32, "43F0", _heap_malloc);
+#define _heap_alloc_and_update_id(node) { \
+    u16 HeapEntryID = D_NEXT_HEAP_MALLOC_ID; \
+    node->allocated = 1; \
```
nanaian, line 79:
```suggestion
    node->allocated = TRUE; \
```

src/43F0.c
```diff
@@ -74,9 +74,128 @@ HeapNode* _heap_create(s32* addr, u32 size) {
     }
 }
 
-INCLUDE_ASM(s32, "43F0", _heap_malloc);
+#define _heap_alloc_and_update_id(node) { \
```
nanaian, line 77: By convention we tend to write macros in SCREAMING_CAPS. I'm personally fine with this but I suspect @ethteck may not be.

It might be possible to match this as a `static inline` func also, or of course we just copy this code into the two places that use it? I'm unsure if we need a macro for 4 lines that only appear twice 📦

src/43F0.c
```diff
@@ -74,9 +74,128 @@ HeapNode* _heap_create(s32* addr, u32 size) {
...
+            curNode->length = size;
+            curNode->allocated = 1;
+
+        } else {
+            curNode->next = nextNode;
+            curNode->length = foundNodeLength;
+            curNode->allocated = 1;
+        }
+
+        return (u8*)curNode + 0x10;
```
nanaian, line 195:
```suggestion
        return (u8*)curNode + sizeof(HeapNode);
```

src/43F0.c
```diff
@@ -74,9 +74,128 @@ HeapNode* _heap_create(s32* addr, u32 size) {
...
+
+            curNode = curNode->next;
+            curNode->next = nextNode;
+            curNode->length = size;
+            curNode->allocated = 1;
+
+        } else {
+            curNode->next = nextNode;
+            curNode->length = foundNodeLength;
+            curNode->allocated = 1;
```
nanaian, line 192: The 1s and 0s relating to `allocated` should be `TRUE` and `FALSE` respectively

src/43F0.c
```diff
@@ -74,9 +74,128 @@ HeapNode* _heap_create(s32* addr, u32 size) {
...
+    foundNode = (HeapNode *) NULL;
+    if (size == 0) {
+        return NULL;
+    }
+
+    foundNodeLength = 0;
+    nextNode = 0;
+
+    for (curNode = head;;curNode = curNode->next) {
+        if (curNode->allocated == 0) {
```
nanaian, line 163:
```suggestion
        if (!curNode->allocated) {
```

src/43F0.c
```diff
@@ -74,9 +74,128 @@ HeapNode* _heap_create(s32* addr, u32 size) {
...
+    size = ALIGN16(size);
+    foundNode = (HeapNode *) NULL;
+    if (size == 0) {
+        return NULL;
+    }
+
+    foundNodeLength = 0;
+    nextNode = 0;
+
+    for (curNode = head;;curNode = curNode->next) {
```
nanaian, line 162:
```suggestion
    for (curNode = head; ; curNode = curNode->next) {
```

src/43F0.c
```diff
@@ -74,9 +74,128 @@ HeapNode* _heap_create(s32* addr, u32 size) {
...
+    for (curHeapNode = head;;curHeapNode = curHeapNode->next) {
+        if (curHeapNode->allocated == 0) {
+            curBlockLength = curHeapNode->length;
+            if ((curBlockLength >= size) && ((curBlockLength < smallestBlockFound) || (smallestBlockFound == 0))) {
+                pPrevHeapNode = curHeapNode;
+                smallestBlockFound = curBlockLength;
+                nextHeapNode = curHeapNode->next;
+            }
+        }
+        if (!curHeapNode->next) {
```
nanaian, line 111: I'd be surprised if this didn't match as the condition in the for loop

src/43F0.c
```diff
@@ -74,9 +74,128 @@ HeapNode* _heap_create(s32* addr, u32 size) {
...
+    size = ALIGN16(size);
+    if (size == 0) {
+        return (s32 *)pPrevHeapNode;
+    }
+
+    smallestBlockFound = 0;
+    nextHeapNode = 0;
+
+    //find the smallest block we can fit into in the free list
+    for (curHeapNode = head;;curHeapNode = curHeapNode->next) {
```
nanaian, line 102:
```suggestion
    for (curHeapNode = head; ; curHeapNode = curHeapNode->next) {
```

src/43F0.c
```diff
@@ -74,9 +74,128 @@ HeapNode* _heap_create(s32* addr, u32 size) {
...
+    HeapNode *nextNode;
+
+    size = ALIGN16(size);
+    foundNode = (HeapNode *) NULL;
+    if (size == 0) {
+        return NULL;
+    }
+
+    foundNodeLength = 0;
+    nextNode = 0;
```
nanaian, line 160:
```suggestion
    nextNode = NULL;
```

src/43F0.c
```diff
@@ -74,9 +74,128 @@ HeapNode* _heap_create(s32* addr, u32 size) {
...
+    return NULL;
+}
 
-INCLUDE_ASM(s32, "43F0", _heap_malloc_tail);
+void* _heap_malloc_tail(HeapNode* head, u32 size) {
+    HeapNode *curNode;
+    u32 newNodeSize;
+    u32 foundNodeLength;
+    HeapNode *foundNode;
+    HeapNode *nextNode;
```
nanaian, line 151: We left-align pointers:
```suggestion
    HeapNode* curNode;
    u32 newNodeSize;
    u32 foundNodeLength;
    HeapNode* foundNode;
    HeapNode* nextNode;
```

nanaian (changes requested):

src/43F0.c
LightningTH: I thought symbols was auto generated, fixed

src/43F0.c
LightningTH: Removed, I personally prefer them for readability but you and Alex both requested they be removed :)

src/43F0.c
LightningTH: fixed, added comments to the tail function too

src/43F0.c
LightningTH: Cast is not required, it can return null and changed. Was a left over from getting the source to line up early on

src/43F0.c
LightningTH: Changed, also changed the others with == 0

src/43F0.c
LightningTH: Putting it as part of the loop condition results in an extra check early in the code to confirm the next pointer prior to doing the for loop.

src/43F0.c
LightningTH: Fixed

src/43F0.c
LightningTH: good catch

src/43F0.c
nanaian: Gonna leave this unresolved incase @ethteck has an opinion on whether we should have the macro at all

src/43F0.c
nanaian: Perhaps it'd be a better match as `while`?

src/43F0.c
```diff
@@ -74,9 +74,138 @@ HeapNode* _heap_create(s32* addr, u32 size) {
...
+            curHeapNode->allocated = FALSE;
+        } else {
+            // take this entry out of the free linked list and mark as allocated
+            pPrevHeapNode->next = nextHeapNode;
+            pPrevHeapNode->length = smallestBlockFound;
+
+            // update the entry id on allocation
+            _HEAP_ALLOC_AND_UPDATE_ID(pPrevHeapNode);
+        }
+        return (u8 *)pPrevHeapNode + sizeof(HeapNode);
```
nanaian, line 141: Left-align pointers:
```suggestion
        return (u8*)pPrevHeapNode + sizeof(HeapNode);
```

This might also be better as `return pPrevHeapNode + 1;`?

src/43F0.c
```diff
@@ -74,9 +74,138 @@ HeapNode* _heap_create(s32* addr, u32 size) {
...
+            curNode->allocated = TRUE;
+
+        } else {
+            // just return this actual block
+            curNode->next = nextNode;
+            curNode->length = foundNodeLength;
+            curNode->allocated = TRUE;
+        }
+
+        return (u8*)curNode + sizeof(HeapNode);
```
nanaian, line 203: Ditto

src/43F0.c
LightningTH: Oops on the space, per the +sizeof, malloc functions (not this one) sometimes return a value in the middle of the struct so it is more common to see some size added instead of doing an array pointer reference. Yes an array reference to [1] would create the same result but I see that as a side effect of it stashing the full struct at the beginning instead of only a partial struct.

src/43F0.c
LightningTH: do {} while(curHeapNode->next); crc fails

LightningTH: I do agree it looks odd though

src/43F0.c
LightningTH: @ethteck which would you prefer? I'm leaving this and the next comment open for final input.

nanaian (approved):

src/43F0.c
LightningTH: Ethan chose 2 temp vars and remove the macro

Issue #391, opened 2021-08-25
Parameterise all flags as enums


nanaian:
We currently parameterise only some flags, like NPC flags, in `enums.h`. Ideally, we'd do this for _all_ flag types, such as that of BattleStatus.

Brotenko:
A list of yet to be parameterised flags:

- [ ] ColliderFlags
- [ ] ColliderBoundingBoxFlags
- [ ] PlayerStatusFlags
- [ ] PlayerStatusAnimFlags
- [ ] BattleStatusFlags1
- [ ] BattleStatusFlags2
- [ ] BattleStatusRushesFlags
- [ ] BattleStatusCurrentTargetListFlags
- [ ] BattleStatusTattleFlags
- [ ] CameraInitDataFlags
- [ ] CameraFlags
- [ ] CameraMoveFlags 
- [ ] CameraPrevFollowFlags
- [ ] CameraPrevPrevFollowFlags
- [ ] StaticItemTargetFlags
- [ ] StaticItemTypeFlags
- [ ] NpcFlags
- [ ] MessagePrintStateFlags
- [ ] MessagePrintStateDelayFlags
- [ ] GameStatusPeachFlags
- [ ] GameStatusDemoFlags
- [ ] MessageDrawStatePrintModeFlags
- [ ] MessageDrawStateEffectFlags
- [ ] ActorPartEventFlags
- [ ] ActorPartFlags3
- [ ] EvtGroupFlags (u8)
- [ ] EntityCollisionFlags (u8)
- [ ] ShopInventoryItemFlags
- [ ] EncounterResetMapEncounterFlags
- [ ] EncounterDefeatFlags
- [ ] SaveDataEnemyDefeatFlags
- [ ] SaveDataGlobalFlags
- [ ] SaveDataAreaFlags
- [ ] EnemyDropsFlags

Names have been chosen by me - The respective data structs that hold these flags are easy to find this way.

PR #525, opened 2021-11-09
Update flags and enums in enums.h

### LINKED ISSUE #506: Use hex for enums
Enum values ( = xyz) and names, in the case of dummy stuff, should use hex and not decimal

Brotenko:
Resolves #506 and resolves the first half of #391. A list of yet to be done flags will be added to the issue.


nanaian (approved):
tysm

PR #526, opened 2021-11-10
Match StartBossBattle

JoshDuMan:


JoshDuMan:
https://decomp.me/scratch/6FwDK

Not sure why slug didn't pick this function up.

Brotenko (changes requested):
Since Github is being very non-compliant, I'll have to do it like this:

- `currentEncounter->eFirstStrike = 0;` should be `currentEncounter->eFirstStrike = FIRST_STRIKE_NONE;`
- `currentEncounter->allowFleeing = 1;` I suggest `TRUE` instead, since this very much seems like a bool and every usage of `allowFleeing` so far only set it to 0 or 1

BowserSlug:
8153 matched functions / 12470 total (65.38%)
1599864 matching bytes / 3727072 total (42.93%)

nanaian (approved):

PR #529, opened 2021-11-13
replaced commented out with if-0s

Alto1772:
Hey, long time no see! It's been long since a few months. Do you guys need this one? It only applies to existing functions that are failed to match and commented out by handmade workers.


JoshDuMan:
I think this would be better if it instead did #ifdef NON_EQUIVALENT, rather than #if 0. Other than that, of course its appreciated!
=

Brotenko:
I remember @ethteck saying that he personally wants most of these commented out functions to be completely removed (I think this was regarding map files only though?)

Personally speaking though, I agree with @JoshDuMan here that they should use `NON_MATCHING` and `NON_EQUIVALENT` though. Almost all the commented out funcs in the map files are equivalent though and only need data migration to match, so it'd be `NON_MATCHING` there and `NON_EQUIVALENT` when a function doesn't actually match the ASM.

ethteck:
Hey, thanks for this! Yeah, I agree with comments above but all the map ones should be just straight-up deleted if you don't mind doing that

ethteck (changes requested):
left my feedback in the last comment - oopz

Alto1772:
wait why is this isnt merged?

JoshDuMan:
> wait, why this isnt merged _yet_?

PRs require reviews from both Ethan and Alex typically. Alex hasn't taken a look yet.

JoshDuMan:
From Ethan- "Ethan — Today at 3:27 AM
It wasn't merged because it's still waiting on the feedback to be addressed"

ethteck:
To summarize: 
* All commented-out code in maps should be deleted entirely
* All other commented-out code should be `#ifdef NON_EQUIVALENT` (not `#ifdef 0 // NON_EQUIVALENT`)

Brotenko:
> To summarize: 
> * All commented-out code in maps should be deleted entirely
> * All other commented-out code should be `#ifdef NON_EQUIVALENT` (not `#ifdef 0 // NON_EQUIVALENT`)

Unless it is actually equivalent but can't be matched due to, for example, missing data migration. Then it should be `NON_MATCHING`

ethteck (approved):
Thank you :) looks great.

nanaian (commented):
The funcs that were added as NON_MATCHING should probably be marked NON_EQUIVALENT as per #534

ethteck:
Oh yeah, oops. that was the intention. After that's done, I approve

nanaian (approved):

PR #538, opened 2021-11-22
Use designated inits for data tables and AC 0A data migration

### LINKED ISSUE #483: Use designated inits for data tables
e.g. bActorTattles

https://gcc.gnu.org/onlinedocs/gcc/Designated-Inits.html

Brotenko:
resolve #483


ethteck (approved):

nanaian (dismissed):
Should remove the hex comments I think, the info is redundant given the designated inits

PR #551, opened 2021-12-13
First overlay's BSS symbols moved to C

ethteck:
This PR migrates a all of the first overlay's BSS symbols into C and out of undefined_syms. There are still BSS issues to sort out, but this brings us one step closer to shiftability,


BowserSlug:
- src/26840_len_20d0.c:142: warning: assignment makes integer from pointer without a cast
- src/341d0.c:54: warning: initialization makes integer from pointer without a cast

JoshDuMan:
My only feedback would be to put somewhere either in bss.c or the splat splits why that files exists. Even if its removed later, its best to not let it slip that its a compiler hack to force correct bss order.

src/341d0.c
```diff
@@ -37,20 +37,20 @@ s32 D_8007F1D0[] = {
...
-s32 D_8007F200[] = {
+s32 D_8007F200[1] = {
     D_8007F1D0
 };
 
-s32 D_8007F204[] = {
-    D_8007F1D0, 0x00000000, 0x00000000
+s32 D_8007F204[1] = {
+    D_8007F1D0
 };
```
nanaian, line 54: Do these even need to be arrays at this point?

nanaian (commented):
👀

Brotenko (approved):

PR #554, opened 2021-12-17
Many func's in 20ec0_len_5040.c

JoshDuMan:
:)


BowserSlug:
🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀 This PR matches 10 functions (+0.01%) on `us`.

src/20ec0_len_5040.c
```diff
@@ -1,5 +1,60 @@
...
+typedef struct HudComplexStatusIcon {
+	/* 0x0 */ s8 active;
+	/* 0x1 */ s8 removing;
+	/* 0x2 */ s8 unk_02;
+	/* 0x3 */ s8 unk_03;
+	/* 0x4 */ s8 unk_04;
+	/* 0x5 */ s8 unk_05;
+	/* 0x8 */ s32 hudElemIndex;
+	/* 0xC */ s32 unk_0C;
+} HudComplexStatusIcon;/* size = 0x10 */
```
nanaian, line 14: Doesn't matter too much but we've gone for this style of comment so far
```suggestion
} HudComplexStatusIcon; // size = 0x10
```

src/20ec0_len_5040.c
```diff
@@ -1,5 +1,60 @@
...
+	/* 0x068 */ HudComplexStatusIcon status4;
+	/* 0x078 */ HudSimpleStatusIcon boostJump;
+	/* 0x080 */ s32 prevIndexBoostJump;
+	/* 0x084 */ HudSimpleStatusIcon boostHammer;
+	/* 0x08C */ s32 prevIndexBoostHammer;
+	/* 0x090 */ HudSimpleStatusIcon boostPartner;
+	/* 0x098 */ HudSimpleStatusIcon surprise;
+	/* 0x0A0 */ HudSimpleStatusIcon peril;
+	/* 0x0A8 */ HudSimpleStatusIcon danger;
+} HudStatusIcon;/* size = 0xB0 */
```
ethteck, line 47: I'd prefer to use the normal ` // size = ` comment format like alex said above. also not to be too pedantic, but since this only goes up to 2 hex digits, the leading 0s can be removed from the offset comments in this struct as well

src/20ec0_len_5040.c
```diff
@@ -176,68 +231,286 @@ INCLUDE_ASM(s32, "20ec0_len_5040", draw_all_status_icons);
...
 
 INCLUDE_ASM(s32, "20ec0_len_5040", func_800477F4);
 
 INCLUDE_ASM(s32, "20ec0_len_5040", func_80047820);
 
-INCLUDE_ASM(s32, "20ec0_len_5040", remove_status_debuff);
+void remove_status_debuff(s32 iconID) {
+    HudStatusIcon* statusIcon = &D_800A0F44[iconID];
+
+    if (statusIcon->status1.active != FALSE && statusIcon->status1.removing == FALSE) {
```
ethteck, line 259: this can just be `if (statusIcon->status1.active && !statusIcon->status1.removing)` (bool checks don't need to expliticly compare with false)

ethteck: also relevant in below funcs

ethteck (changes requested):
These are looking great! The only other thing I'd say is that ideally we make/use enums for the flags

PR #559, opened 2021-12-22
Match filemenu_set_cursor_goal_pos

JoshDuMan:
https://decomp.me/scratch/xnc5w
https://decomp.me/scratch/9QaEy


src/163400.c
```diff
@@ -81,7 +81,37 @@ void filemenu_set_cursor_alpha(s32 arg0) {
     filemenu_cursorGoalAlpha2 = arg0;
 }
 
-INCLUDE_ASM(s32, "163400", filemenu_set_cursor_goal_pos);
+void filemenu_set_cursor_goal_pos(s32 windowID, s32 posX, s32 posY) {
+    Window* window = &gWindows[windowID];
+
+    if (D_80249BB0 != 0
```
ethteck: I think this can all go on the same line and fit far under 120 chars

src/163400.c
```diff
@@ -81,7 +81,37 @@ void filemenu_set_cursor_alpha(s32 arg0) {
...
+            || get_game_mode() == 0xF
+            || get_game_mode() == 0xD) {
+        if (D_80249BB0 != 0) {
+            s32 i;
+
+            for (i = 0x2C; i < 0x40; i++) {
```
ethteck, line 93:
```suggestion
            for (i = 0x2C; i < ARRAY_COUNT(gWindows); i++) {
```

src/163400.c
```diff
@@ -81,7 +81,37 @@ void filemenu_set_cursor_alpha(s32 arg0) {
...
+                Window* window = &gWindows[i];
+                s8 parent = window->parent;
+
+                if ((parent == -1 || parent == 0x2C) && (window->flags & 8)) {
+                    break;
+                }
+            }
+            if (i >= 0x40) {
```
ethteck, line 101:
```suggestion
            if (i >= ARRAY_COUNT(gWindows)) {
```

src/163400.c
```diff
@@ -81,7 +81,37 @@ void filemenu_set_cursor_alpha(s32 arg0) {
...
+                D_80249BB0 = 0;
+            }
+        }
+        D_80249BA0 = posX;
+        D_80249B94 = posX;
+        D_80249BA4 = posY;
+        D_80249B98 = posY;
+
```
ethteck, line 109: I think this empty line isn't really that useful. If anything I'd add one 4 above before the first `... = posX` statement. but not really important

src/163400.c
```diff
@@ -81,7 +81,37 @@ void filemenu_set_cursor_alpha(s32 arg0) {
     filemenu_cursorGoalAlpha2 = arg0;
 }
 
-INCLUDE_ASM(s32, "163400", filemenu_set_cursor_goal_pos);
+void filemenu_set_cursor_goal_pos(s32 windowID, s32 posX, s32 posY) {
+    Window* window = &gWindows[windowID];
+
+    if (D_80249BB0 != 0
+            || get_game_mode() == 0xF
```
ethteck, line 88: Do we have a game mode enum? We probably should if we don't

ethteck (changes requested):

src/163400.c
JoshDuMan: I can do array count here, but it seems like gWindows has subdivisions within it for each range that probably need an enum.

src/163400.c
JoshDuMan: No we do not, at least from what I saw

JoshDuMan: nvm Im wrong

BowserSlug:

src/163400.c
ethteck: yeah or some defines. but I think it's safe to use this for the end of the array. Or it's multiple arrays..hm

ethteck (approved):

nanaian (approved):

PR #581, opened 2021-12-28
Decomp some player render functions

Rainchus:


BowserSlug:
🚀🚀 This PR matches 2 functions (+0.04%) on `us`.

src/77480.c
```diff
@@ -783,7 +787,59 @@ void render_player(void) {
...
+    s8 renderModeTemp;
+    void (*phi_a0)(void);
+
+    if (playerStatus->flags & PLAYER_STATUS_FLAGS_40000000) {
+        playerStatus->flags &= ~PLAYER_STATUS_FLAGS_40000000;
+        get_screen_coords(gCurrentCamID, playerStatus->position.x, playerStatus->position.y,
+                          playerStatus->position.z, &x, &y, &z);
+        if (!(playerStatus->flags & PLAYER_STATUS_FLAGS_20000)) {
+            if (playerStatus->alpha1 != playerStatus->alpha2) {
+                if (playerStatus->alpha1 < 254) {
```
ethteck, line 804: <= 255 (since that's the max alpha val)

src/77480.c
```diff
@@ -783,7 +787,59 @@ void render_player(void) {
...
+                    renderModeTemp = 22;
+                    if (!(playerStatus->animFlags & PLAYER_STATUS_ANIM_FLAGS_1000000)) {
+                        renderModeTemp = 17;
+                    }
```
ethteck, line 808: instead of setting it to 22 before, you should be able to else temp=22

src/77480.c
```diff
@@ -783,7 +787,59 @@ void render_player(void) {
...
+
+                    playerStatus->renderMode = renderModeTemp;
```
ethteck, line 810: are you sure renderModeTemp is needed? can you just do the payerStatus->renderMode= in both the if and else?

src/77480.c
```diff
@@ -783,7 +787,59 @@ void render_player(void) {
...
+
+        if (!(playerStatus->animFlags & PLAYER_STATUS_ANIM_FLAGS_IN_DISGUISE)) {
+            rtPtr->appendGfxArg = playerStatus;
+            rtPtr->distance = -z;
+            rtPtr->renderMode = playerStatus->renderMode;
+            phi_a0 = appendGfx_player_spin;
+
+            if (!(playerStatus->flags & PLAYER_STATUS_ANIM_FLAGS_20000)) {
+                phi_a0 = appendGfx_player;
+            }
```
ethteck, line 834: you can else the phi_a0 here too (and potentially don't need the temp)

src/77480.c
```diff
@@ -783,7 +787,59 @@ void render_player(void) {
...
+    void (*phi_a0)(void);
+
+    if (playerStatus->flags & PLAYER_STATUS_FLAGS_40000000) {
+        playerStatus->flags &= ~PLAYER_STATUS_FLAGS_40000000;
+        get_screen_coords(gCurrentCamID, playerStatus->position.x, playerStatus->position.y,
+                          playerStatus->position.z, &x, &y, &z);
+        if (!(playerStatus->flags & PLAYER_STATUS_FLAGS_20000)) {
+            if (playerStatus->alpha1 != playerStatus->alpha2) {
+                if (playerStatus->alpha1 < 254) {
+                    renderModeTemp = 22;
```
nanaian, line 805: We have an enum for render modes I believe

ethteck (changes requested):
Great job! Just a few things to try

src/77480.c
```diff
@@ -783,7 +787,59 @@ void render_player(void) {
...
 }
 
-INCLUDE_ASM(void, "77480", render_player_model);
+void render_player_model(void) {
+    RenderTask task;
+    RenderTask* rtPtr = &task;
+    PlayerStatus* playerStatus = &gPlayerStatus;
+    s32 x, y, z;
+    s8 renderModeTemp;
+    void (*phi_a0)(void);
```
nanaian, line 796: Type this as `void (*appendGfx)(void*)` and you can get rid of the `void*` cast on L836

src/77480.c
```diff
@@ -848,7 +904,90 @@ void appendGfx_player(void) {
...
+
+        guTranslateF(translation, 0.0f, -playerStatus->colliderHeight * 0.5f, 0.0f);
+        guMtxCatF(translation, mtx, mtx);
+        guRotateF(rotation, yaw, 0.0f, 1.0f, 0.0f);
+        guMtxCatF(mtx, rotation, mtx);
+        guRotateF(rotation, playerStatus->spriteFacingAngle, 0.0f, 1.0f, 0.0f);
+        guMtxCatF(mtx, rotation, mtx);
+        guTranslateF(translation, 0.0f, playerStatus->colliderHeight * 0.5f, 0.0f);
+        guMtxCatF(mtx, translation, mtx);
+        guScaleF(scale, 0.71428573f, 0.71428573f, 0.71428573f);
```
nanaian, line 977: 0.71428573f is `5.0/7.0` or something - see #580

nanaian (changes requested):

src/77480.c
nanaian: May also need to make `appendGfx_player` take an unused void* arg

src/77480.c
nanaian: This needs to be done as part of the merge I think

src/77480.c
Brotenko: Indeed we have!

src/77480.c
```diff
@@ -848,7 +904,90 @@ void appendGfx_player(void) {
...
+    f32 px, py, pz;
+    s32 x, y, z;
+    s32 i;
+    s32 flags;
+
+    for (i = 0; i < 2; i++) {
+        yaw = -gCameras[gCurrentCamID].currentYaw;
+
+        if (i == 0) {
+            if ((playerStatus->spriteFacingAngle > 90.0f) && (playerStatus->spriteFacingAngle <= 180.0f)) {
```
Brotenko, line 925: Are the parenthese required here?

src/77480.c
```diff
@@ -848,7 +904,90 @@ void appendGfx_player(void) {
...
+                yaw = 180.0f - playerStatus->spriteFacingAngle;
+            } else {
+                if ((playerStatus->spriteFacingAngle > 180.0f) && (playerStatus->spriteFacingAngle <= 270.0f)) {
```
Brotenko, line 928: Are the parenthese required here?

src/77480.c
```diff
@@ -848,7 +904,90 @@ void appendGfx_player(void) {
...
+                    yaw = playerStatus->spriteFacingAngle - 180.0f;
+                } else if ((playerStatus->spriteFacingAngle > 270.0f) && (playerStatus->spriteFacingAngle <= 360.0f)) {
```
Brotenko, line 930: Are the parenthese required here?

src/77480.c
Brotenko: Yeah, this is `SPRITE_PIXEL_SCALE` now, so you probably need to rebase, otherwise you introduce this faulty into the codebase

Brotenko (changes requested):

src/77480.c
Rainchus: The equivalent would be `if (playerStatus->alpha1 <= 253) {` however i believe `if (playerStatus->alpha1 < 254) {` is best

src/77480.c
Rainchus: doesn't match with if/else

PR #597, opened 2021-12-31
Disassemble KMR 2

JoshDuMan:


BowserSlug:
✅ This PR fixes 50 warnings!

ver/us/splat.yaml
```diff
@@ -3726,15 +3726,33 @@ segments:
...
+    - [0x439F90, c, actor/blue_goomba2]
+    - [0x43A210, c, actor/red_goomba2]
+    - [0x43A4C0, c, actor/goomba_king]
+    - [0x43A5A0, c, actor/goomnut_tree]
+    - [0x43A7E0, c, stage/kmr_02]
+    - [0x43A7E0, c, stage/kmr_03]
+    - [0x43AA50, c, stage/kmr_04]
+    - [0x43ACC0, c, stage/kmr_05]
+    - [0x43AF30, c, stage/kmr_06]
+    - [0x43B1E0, .data, area]
```
ethteck, line 3744: Since all the data has been disassembled, you can change this line to just `- [0x43B1E0]` and remove the following ones

ethteck (changes requested):
Amazing work! Aside from the one thing I found, LGTM

src/battle/battle.h
```diff
@@ -271,7 +271,10 @@ typedef struct ActorDesc {
 
 typedef struct FormationRow {
     /* 0x00 */ ActorDesc* actor;
-    /* 0x04 */ s32 position; ///< Home position. May also be a `Vector3*`.
+    /* 0x04 */ union {
+                   s32    index; ///< Home position. May also be a `Vector3*`.
```
nanaian, line 275: I think we can remove this comment now

src/battle/battle.h
```diff
@@ -271,7 +271,10 @@ typedef struct ActorDesc {
...
+                   Vec3i* vec;
+               } pos;
```
nanaian, line 277: You've renamed this to `pos`, I think `home` would be better

src/world/common/FoliageTransform.inc.c
```diff
@@ -0,0 +1,80 @@
+#include "common.h"
```
nanaian, line 1: This file is in src/**world**/common/, which is a bit weird considering this is used in battles too. Maybe `src/common` is better? @ethteck may have opinions

tools/star_rod_idx_to_c.py
```diff
@@ -853,23 +862,24 @@ def transform_symbol_name(symbol):
...
             out += INDENT + f".flags = {read_flags(d[0], 'ActorFlags')},\n"
-            out += INDENT + f".type = {d[1]},\n" # TODO enum
+            out += INDENT + f".type = {read_enum(d[1], 'ActorType')},\n"
             out += INDENT + f".level = {d[2]},\n"
             out += INDENT + f".maxHP = {d[3]},\n"
             out += INDENT + f".partCount = ARRAY_COUNT({read_ptr(d[5], symbol_map)}),\n"
             out += INDENT + f".partsData = {read_ptr(d[5], symbol_map)},\n"
             out += INDENT + f".script = {read_ptr(d[6], symbol_map)},\n"
-            out += INDENT + f".statusTable = {read_ptr(d[7], symbol_map)},\n"
+            out += INDENT + f".statusTable = (DictionaryEntry*) {read_ptr(d[7], symbol_map)},\n"
```
nanaian, line 882: Thinking it might be better to change the type of `statusTable` rather than do this... I believe there was a discussion in Discord regarding removing DictionaryEntry entirely? What was the outcome of that? I'd be happy to although we might have to modify some code

include/enums.h
```diff
@@ -3011,4 +3011,9 @@ enum BattleSubmenuStatus {
     BATTLE_SUBMENU_STATUS_NO_TARGETS_2      = -2, // TODO: determine difference, probably uses a different error message
 };
 
+enum DictionaryIndex{
+    DICTIONARY_KEY   = 0,
+    DICTIONARY_VALUE = 1,
+};
```
nanaian, line 3017: Unsure if this is really needed, would be nice if we could do like
```c
s32 key = defenseTable[0];
s32 value = defenseTable[1];

// ...

defenseTable += 2;
```
but idk if that would match in lookup_defense?

include/common_structs.h
```diff
@@ -13,10 +13,7 @@ typedef ApiStatus(*ApiFunc)(struct Evt*, s32);
 
 typedef Bytecode EvtSource[0];
 
-typedef struct DictionaryEntry {
-    /* 0x00 */ s32 key;
-    /* 0x04 */ s32 value;
-} DictionaryEntry; // size = 0x08
+typedef s32 DictionaryEntry[];
```
nanaian, line 16: I'd probably just get rid of this type altogether

include/enums.h
```diff
@@ -3011,4 +3011,9 @@ enum BattleSubmenuStatus {
     BATTLE_SUBMENU_STATUS_NO_TARGETS_2      = -2, // TODO: determine difference, probably uses a different error message
 };
 
+enum DictionaryIndex{
```
nanaian, line 3014:
```suggestion
enum DictionaryIndex {
```
if we keep it

nanaian (changes requested):

include/enums.h
nanaian: Actually it might be cool to do
```suggestion
    DICTIONARY_ENTRY_SIZE,
};
```
at the bottom and then increment defenseTable by DICTIONARY_ENTRY_SIZE (2)

ethteck (approved):

nanaian (approved):

PR #606, opened 2022-01-06
Improve disasm_script.py and do some sample scripts

JoshDuMan:


include/enums.h
```diff
@@ -2486,6 +2496,83 @@ enum CameraFlags {
...
+    BTL_CAM_PRESET_A  = 0,
+    BTL_CAM_PRESET_B ,
+    BTL_CAM_PRESET_C ,
+    BTL_CAM_PRESET_D ,
+    BTL_CAM_PRESET_E ,
+    BTL_CAM_PRESET_F ,
+    BTL_CAM_PRESET_G ,
+    BTL_CAM_PRESET_H ,
+    BTL_CAM_PRESET_I ,
+    BTL_CAM_PRESET_9 ,
```
nanaian, line 2509: Commas here

src/battle/common/actor/goomba.inc.c
```diff
@@ -119,7 +119,7 @@ EvtSource N(idle_80219938) = {
     EVT_LOOP(LW(0))
         EVT_LABEL(0)
         EVT_CALL(GetStatusFlags, ACTOR_SELF, LW(1))
-        EVT_IF_FLAG(LW(1), 0x35D000)
+        EVT_IF_FLAG(LW(1), (STATUS_FLAG_SLEEP | STATUS_FLAG_FROZEN | STATUS_FLAG_FEAR | STATUS_FLAG_PARALYZE | STATUS_FLAG_DIZZY | STATUS_FLAG_STONE | STATUS_FLAG_STOP))
```
nanaian, line 122: I'm pretty sure these parens aren't required anymore -- the double parens originally in flags were to tell the SCRIPT compiler things

nanaian (changes requested):
Incredible

ethteck (approved):
looks great! One note to myself is that effects actually store their id in their main function, so I'll probably end up shifting all effects to use enum values instead of just numbers there.

nanaian (approved):
ty

PR #618, opened 2022-01-13
Improve `find_duplicates.py` performance, ~30x

zbanks:
- Use length heuristic to avoid computing true Lev ratio
- Prebuild `sym_bytes` map for `sym_name` -> `query_bytes`

I think this is at least ~30x faster? I can now run a whole cross search on the repo in ~45 seconds.

### DIFF
--- a/tools/find_duplicates.py
+++ b/tools/find_duplicates.py
@@ -124,6 +124,11 @@ def diff_syms(qb, tb):
     if len(tb[1]) < 8:
         return 0
 
+    # The minimum edit distance for two strings of different lengths is `abs(l1 - l2)`
+    # Quickly check if it's impossible to beat the threshold. If it is, then return 0
+    l1, l2 = len(qb[0]), len(tb[0])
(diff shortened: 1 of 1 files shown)


tools/find_duplicates.py
```diff
@@ -261,36 +266,38 @@ def do_cross_query():
...
-                            ccount[cluster_first] += 1
-
-                        #if len(cluster) % 10 == 0 and len(cluster) >= 10:
-                        print(f"Cluster {cluster_first} grew to size {len(cluster)} - {sym_name}: {str(cluster_score)}")
-                        break
-                if not cluster_match:
-                    clusters.append([sym_name])
+                sym_bytes[sym_name] = get_symbol_bytes(map_offsets, sym_name)
+
+    for sym_name, query_bytes in sym_bytes.items():
```
zbanks, line 278: This just looks big, but it's just a de-indent & replacing `get_pair_score(...)` with `diff_syms(...)`

PR #627, opened 2022-01-19
matched func 8023F060

Mathyland:
* Match a func

* Delete func_8023F060.s


BowserSlug:

nanaian (approved):
Great first PR! The function looks good - we probably can't improve the match much more without knowing the argument types this func takes.

PR #632, opened 2022-01-22
The Great Effect Naming

ethteck:
This PR is for naming all effects and their files, functions, IDs, and structs.


include/dead.h
```diff
@@ -46,19 +46,19 @@
...
 #define subtract_hp dead_subtract_hp
 #define get_entity_by_index dead_get_entity_by_index
-#define playFX_52 dead_playFX_52
-#define playFX_33 dead_playFX_33
+#define fx_energy_orb_wave dead_playFX_52
+#define fx_radial_shimmer dead_playFX_33
 #define set_screen_overlay_color dead_set_screen_overlay_color
 #define set_screen_overlay_params_front dead_set_screen_overlay_params_front
-#define playFX_82 dead_playFX_82
+#define fx_sun dead_playFX_82
```
JoshDuMan, line 61: would be neat if these dead functions were renamed as well

JoshDuMan (commented):
Hard to see everything because its obviously a large diff, but this looks generally good to me

include/effects_internal.h
```diff
@@ -25,12 +26,137 @@ void shim_guTranslateF(float mf[4][4], float x, float y, float z);
...
 void shim_mem_clear(void* data, s32 numBytes);
 void shim_transform_point(Matrix4f mtx, f32 inX, f32 inY, f32 inZ, f32 inS, f32* outX, f32* outY, f32* outZ, f32* outS);
-
-s32 shim_load_effect(s32 effectIndex);
-
 s32 shim_npc_raycast_down_sides(s32, f32*, f32*, f32*, f32*);
 void shim_sfx_play_sound_at_position(s32 soundID, s32 value2, f32 posX, f32 posY, f32 posZ);
 
-void fx_6_main(s32 type, f32 x, f32 y, f32 z, f32 arg4);
+#define FX_MAIN(name) void name(s32); // the actual signature is going to be wrong, but it doesn't matter
```
nanaian, line 32: Can you `undef` this at the end of the file

nanaian, line 32: Why doesn't it matter? Could we do `void name();` instead then?

ethteck: I'll make the args void, but unspecified args is a compile warning

ethteck: undef FX_MAIN? why?

PR #634, opened 2022-01-29
Fix Install Script for Ubuntu

bennyman123abc:
This pull request is a simple fix which adds a check for the unsupported distro message, and adds cURL to the install dependencies for all supported distros in the event it is not installed.

The issue was found on a fresh install of Ubuntu 21.04 (Minimal install) as cURL was not included by default.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -7,7 +7,11 @@ else
 fi
 
 uname=`uname`
+unset supported
+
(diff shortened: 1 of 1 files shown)


ethteck (approved):
Thanks, looks great!

nanaian (approved):

PR #635, opened 2022-01-30
SFX funcs, configure fix

ethteck:


BowserSlug:
🚀🚀🚀🚀🚀 This PR matches 5 functions (+0.07%) on `us`.

✅ This PR fixes 14 warnings!

nanaian (approved):
LGTM. I'm iffy on the configure 'fix' because we should really be using `-M` for depfiles rather than disabling it entirely but I'll make a separate issue for it because its nbd for now

PR #640, opened 2022-02-08
fix install script

Unnunu:
First, variable assignment in bash must be without spaces. Second, downloading binutils etc was never called because of exits.


ethteck:
Ah, nice. Yeah I fixed this in my incoming PR but it's best to get it in asap.

ethteck (approved):

ethteck:
@nanaian ninja-merging this since it's a blocker to use the repo atm

PR #645, opened 2022-02-09
func_800F0CB0 matched

cfillak:


src/8a160_len_700.c
```diff
@@ -18,7 +18,32 @@ typedef struct {
...
+    struct_func_800F0CB0 *temp_a1;
+    s32 phi_v1;
+    sub_struct_func_800F0CB0 *phi_a1;
+
+    temp_a1 = &D_8010D000[arg0];
+    phi_a1 = &temp_a1->unk00[0];
+    temp_v1 = 0x39;
+
+    if (arg0 <= 0) {
+      phi_a1->unk0 = 1;
```
nanaian, line 32: Use 4 spaces for indentation:
```suggestion
        phi_a1->unk0 = 1;
```

nanaian (changes requested):
Nice match! Just a few style nits

src/8a160_len_700.c
```diff
@@ -18,7 +18,32 @@ typedef struct {
...
 extern struct_func_800F0CB0 D_8010D000[];
 
-INCLUDE_ASM(s32, "8a160_len_700", func_800F0CB0);
+void func_800F0CB0(s32 arg0, f32 arg1, f32 arg2, f32 arg3) {
+    s32 temp_v1;
+    struct_func_800F0CB0* temp_a1;
+    s32 phi_v1;
+    sub_struct_func_800F0CB0* phi_a1;
+
+    temp_a1 = &D_8010D000[arg0];
```
ethteck, line 27: We typically like to arrange variable declarations at the beginning of the function so that we can define them and declare them on the same line. So the beginning of the function would ideally look like:

```
struct_func_800F0CB0* temp_a1 = &D_8010D000[arg0];
sub_struct_func_800F0CB0* phi_a1 = &temp_a1_unk00[0];
s32 temp_v1 = 0x39;
s32 phi_v1;
```

src/8a160_len_700.c
```diff
@@ -18,7 +18,32 @@ typedef struct {
...
+        phi_a1->unk8 = arg1;
+        phi_a1->unkC = arg2;
+        phi_a1->unk18 = arg3;
+        phi_a1->unk1 = 0x28;
+        phi_a1->unk2 = 0;
+        phi_a1 = &temp_a1->unk00[1];
+        phi_v1 = 1;
+        while (phi_v1 < temp_v1) {
+            phi_a1->unk7 = (s32) (-(f32) phi_v1 * 0.0f) - 1;
+            phi_v1 = phi_v1 + 1;
```
ethteck, line 42: we like to write x = x op y as x op=y. In this case, since it's incrementing by one, we do `phi_v1++`

src/8a160_len_700.c
```diff
@@ -18,7 +18,32 @@ typedef struct {
...
+    if (arg0 <= 0) {
+        phi_a1->unk0 = 1;
+        phi_a1->unk8 = arg1;
+        phi_a1->unkC = arg2;
+        phi_a1->unk18 = arg3;
+        phi_a1->unk1 = 0x28;
+        phi_a1->unk2 = 0;
+        phi_a1 = &temp_a1->unk00[1];
+        phi_v1 = 1;
+        while (phi_v1 < temp_v1) {
```
ethteck, line 40: We can leave this for now, but I have a strong suspicion that this is actually a for-loop and should probably be cleaned up to look like one at some point

ethteck (changes requested):
Hey, great job on your first function! I left a few other style notes - you don't need to worry about the for loop thing for now, but I just wanted to mention it as it will definitely come up again in the future. mips_to_c decompiles for loops as while loops. I'll maybe do another pass over this file soon and see if I can figure some of these structs and things out.

src/8a160_len_700.c
```diff
@@ -20,28 +20,28 @@ extern struct_func_800F0CB0 D_8010D000[];
...
+        phi_a1->unk1 = 0x28;
+        phi_a1->unk2 = 0;
+        phi_a1 = &temp_a1->unk00[1];
+        phi_v1 = 1;
+    while (phi_v1 < temp_v1) {
         phi_a1->unk7 = (s32) (-(f32) phi_v1 * 0.0f) - 1;
         phi_v1 = phi_v1 + 1;
         phi_a1++;
-      }
+    }
```
nanaian: This block should be indented by 4 more spaces, as it's inside the `if`

nanaian (changes requested):

PR #657, opened 2022-03-03
Update progress-jp badge to the correct URL

Ryan-Myers:
Just happened to notice this and figured I'd suggest the change.


ethteck:
ah thanks!

PR #660, opened 2022-03-11
hud_element.c

Unnunu:


BowserSlug:
🚀🚀🚀🚀🚀🚀🚀🚀 This PR matches 8 functions (+0.41%) on `us`.

✅ This PR fixes 41 warnings!

src/16F740.c
```diff
@@ -291,7 +291,7 @@ void btl_state_update_victory(void) {
             player->transStatus = 0;
             player->koStatus = 0;
             player->koDuration = 0;
-            player->debuffIcon->ptrPropertyList[15] = 0;
+            ((s32*)player->debuffIcon->data)[15] = 0;
```
nanaian, line 294: This typing is looking weird; do we have `data` typed wrong or something? cc @ethteck

nanaian (approved):

src/16F740.c
ethteck: we ideally should cast to the appropriate effect data struct, but we may not have one for this effect? for now I think it's ok but maybe should be left with a todo (or even remove the s32* cast to keep a warning there as a reminder to us)

Unnunu: `/* 0x0C */ void* data; // often but not always EffectInstanceData`
Since I don't know anything about effects, I decided to leave it without casting to EffectInstanceData

ethteck: yeah that's fair. we can address this later

src/hud_element.c
```diff
@@ -40,94 +50,653 @@ Vec3s gHudElementSizes[26] = {
...
+
+                    if (!flipX && !flipY) {
+                        gDPLoadTextureTile(gMasterGfxPos++, imageAddr, G_IM_FMT_RGBA, G_IM_SIZ_32b, texSizeX, texSizeY, uls, ult, lrs, lrt, 0,
+                                            G_TX_NOMIRROR | G_TX_WRAP, G_TX_NOMIRROR | G_TX_WRAP, masks, maskt, G_TX_NOLOD, G_TX_NOLOD);
+                    } else {
+                        gDPLoadTextureTile(gMasterGfxPos++, imageAddr, G_IM_FMT_RGBA, G_IM_SIZ_32b, texSizeX, texSizeY, uls, ult, lrs, lrt, 0,
+                                            G_TX_MIRROR | G_TX_WRAP, G_TX_MIRROR | G_TX_WRAP, masks, maskt, G_TX_NOLOD, G_TX_NOLOD);
+                    }
+                    break;
+
```
ethteck, line 512: We don't have empty lines between switch cases elsewhere in the codebase. I think it's okay for larger cases, but we should be consistent probably

ethteck: (relevant before this point too, just saw it a few times and decided to comment here)

src/hud_element.c
```diff
@@ -229,15 +798,970 @@ void update_hud_elements(void) {
...
+    HudTransform* hudTransform = hudElement->hudTransform;
+    s32* nextPos = (s32*)hudElement->readPos;
+
+    switch (*nextPos++) {
+        case HUD_ELEMENT_OP_End:
+            hudElement->updateTimer = 60;
+            flags = hudElement->flags;
+            hudElement->flags = flags | HUD_ELEMENT_FLAGS_ANIMATION_FINISHED;
+            break;
+
```
ethteck, line 827: like here I think we don't need the new lines

src/hud_element.c
```diff
@@ -229,15 +798,970 @@ void update_hud_elements(void) {
...
+                entryPalette = gHudElementCacheTablePaletteBattle;
+            }
 
-INCLUDE_ASM(void, "hud_element", render_hud_elements_frontUI, void);
+            i = 0;
+            while (TRUE) {
+                if (entryRaster[i].id == *nextPos) {
+                    break;
+                }
+                ASSERT(++i < 192);
```
ethteck, line 958: I'm seeing 192 a lot. If we can't do ARRAY_COUNT, it might be good to have it as a define

src/hud_element.c
```diff
@@ -229,15 +798,970 @@ void update_hud_elements(void) {
...
+    }
+
+    for (i = 0; i < count; i++) {
+        hudElement = (*gHudElements)[sortedElements[i]];
+
+        if (hudElement->readPos == NULL) {
+            break;
+        }
+
+        if ((hudElement->flags & HUD_ELEMENT_FLAGS_FIXEDSCALE) == 0) {
```
ethteck, line 1254: ! for flag checks instead of == 0

src/hud_element.c
```diff
@@ -229,15 +798,970 @@ void update_hud_elements(void) {
...
+                offsetY = -drawSizeY / 2;
+            }
+
+            if ((hudElement->flags & HUD_ELEMENT_FLAGS_REPEATED) == 0) {
+                if ((hudElement->flags & HUD_ELEMENT_FLAGS_SHADOW) != 0) {
+                    draw_rect_hud_element(hudElement, texSizeX, texSizeY, drawSizeX, drawSizeY, offsetX, offsetY, TRUE, TRUE);
+                }
+                draw_rect_hud_element(hudElement, texSizeX, texSizeY, drawSizeX, drawSizeY, offsetX, offsetY, TRUE, FALSE);
+            } else {
+                if ((hudElement->flags & HUD_ELEMENT_FLAGS_SHADOW) != 0) {
```
ethteck, line 1292: no !=0 or the inner parens needed

src/hud_element.c
```diff
@@ -229,15 +798,970 @@ void update_hud_elements(void) {
...
+    Matrix4f sp1A0;
+    Matrix4f sp1E0;
+    Matrix4f sp220;
+    Matrix4f sp260;
+    FoldImageRecPart sp2A0;
+    f32 phi_f24, phi_f20;
+    s32 height, width;
+    HudTransform* transform;
+    HudTransform *new_var;
+    VtxRect* ums;
```
ethteck, line 1464: this variable (and thus this type) seems unused? can it be removed for now?

ethteck (changes requested):
Great job as always! Just some minor things

PR #663, opened 2022-03-16
Decompiled and Matched unknown_move

Nasina7:
Decompiled battle/move/unknown_move


src/battle/move/unknown_move/779C90.c
```diff
@@ -1,7 +1,34 @@
...
+
+    script->varTable[0] = 99;
+
+    enemyTargetPowerBounceChance = enemyTargetActor->actorBlueprint->powerBounceChance;
+
+    if (is_ability_active(0) != 0) {
+        enemyTargetPowerBounceChance += 7;
+    }
+
+    D_802A26D0_77B360 *= enemyTargetPowerBounceChance;
```
pixel-stuck, line 25: Is the / 100 required on a separate line, or could the entire calculation be on this line only?

src/battle/move/unknown_move/779C90.c
```diff
@@ -1,7 +1,34 @@
...
 #include "world/common/UnkMoveFunc1.inc.c"
 
-INCLUDE_ASM(s32, "battle/move/unknown_move/779C90", func_802A10E4_779D74);
+extern s32 D_802A26D0_77B360;
+
+ApiStatus func_802A10E4_779D74(Evt* script, s32 isInitialCall) {
+    Actor *enemyTargetActor;
+    u32 enemyTargetPowerBounceChance;
+
+    enemyTargetActor = get_actor((s32) get_actor(script->owner1.enemyID)->targetActorID);
```
pixel-stuck, line 15: Is the s32 cast needed here?

src/battle/move/unknown_move/779C90.c
pixel-stuck: Also, I guess this global would be chance as percentage, though maybe it's used elsewhere

BowserSlug:

src/battle/move/unknown_move/779C90.c
Nasina7: Nope, just removed it :+1:

src/battle/move/unknown_move/779C90.c
Nasina7: It seems like it does need to be on a separate line, although I could just be missing something.

pixel-stuck: Globals are a bit weird, so it probably is correct as you have it, but I figured it was worth a shot :)

Nasina7: :+1:

src/battle/move/unknown_move/779C90.c
```diff
@@ -1,7 +1,34 @@
...
+    Actor *enemyTargetActor;
+    u32 enemyTargetPowerBounceChance;
+
+    enemyTargetActor = get_actor(get_actor(script->owner1.enemyID)->targetActorID);
+
+    script->varTable[0] = 99;
+
+    enemyTargetPowerBounceChance = enemyTargetActor->actorBlueprint->powerBounceChance;
+
+    if (is_ability_active(0) != 0) {
```
nanaian, line 21:
```suggestion
    if (is_ability_active(ABILITY_DODGE_MASTER)) {
```

nanaian (changes requested):

src/battle/move/unknown_move/779C90.c
```diff
@@ -1,7 +1,34 @@
...
+
 #define NAMESPACE battle_move_unknown_move
 
 #include "world/common/UnkMoveFunc1.inc.c"
 
-INCLUDE_ASM(s32, "battle/move/unknown_move/779C90", func_802A10E4_779D74);
+extern s32 D_802A26D0_77B360;
+
+ApiStatus func_802A10E4_779D74(Evt* script, s32 isInitialCall) {
+    Actor *enemyTargetActor;
```
ethteck, line 12: Pointer on the left pleez (Actor*) and also please move the enemyTargetActor initialization into this line as well

ethteck (changes requested):

PR #670, opened 2022-03-25
Santas workshop

HailSanta:
added missing player fields, cleaned up entity data/code, standardized hud_element function names


BowserSlug:
✅ This PR fixes 4 warnings!

include/common_structs.h
```diff
@@ -564,12 +626,16 @@ typedef struct Entity {
     /* 0x3C */ UNK_PTR renderSetupFunc; // pointer to draw func(?)
     /* 0x40 */ union {
         s32* any;
+        SwitchData* swtch;
```
ethteck, line 629: unfortunate that we have to do this since `switch` is a keyword..wondering if there's a better alternative but yeah mawp

include/enums.h
```diff
@@ -1479,7 +1479,7 @@ enum ActionStates {
     ACTION_STATE_USE_SNEAKY_PARASOL             = 0x00000019,
     ACTION_STATE_SPIN                           = 0x0000001A,
     ACTION_STATE_ENEMY_FIRST_STRIKE             = 0x0000001B,
-    ACTION_STATE_GET_STAR_SPIRIT                = 0x0000001C,
+    ACTION_STATE_RAISE_ARMS                = 0x0000001C,
```
ethteck, line 1482:
```suggestion
    ACTION_STATE_RAISE_ARMS                     = 0x0000001C,
```

include/enums.h
```diff
@@ -3378,4 +3384,42 @@ enum RushFlags {
     RUSH_FLAG_POWER = 2,
 };
 
+enum FileMenuMessages {
+    FILE_MESSAGE_NONE                       = 0,
+	FILE_MESSAGE_SELECT_FILE_TO_START       = 1, // Select file to start:[End]
```
ethteck, line 3389: these should be 4 space-indented but every line except the first uses tab characters

src/165490.c
```diff
@@ -681,15 +682,15 @@ void filemenu_main_handle_input(MenuPanel* menu) {
...
+                    if (gSaveSlotHasData[menu->selected] == 0) {
                         sfx_play_sound(SOUND_MENU_NEXT);
                         filemenu_8024C098 = 2;
                         filemenu_menus[filemenu_8024C098]->page = 1;
                         set_window_update(0x2C, main_menu_window_update);
                         set_window_update(0x2F, 1);
                         set_window_update(0x32, 2);
                         fio_save_game(menu->selected);
-                        D_80077A24[menu->selected] = 1;
+                        gSaveSlotHasData[menu->selected] = 1;
```
ethteck, line 693: if this is a bool, all of the checks can be ! for == 0 and the setting can be TRUE/FALSE instead of 1/0

src/world/partner/goombario.c
```diff
@@ -74,7 +74,7 @@ s32 func_802BD1D0_3170F0(Evt* script, s32 isInitialCall) {
         D_8010C954 = 0;
     }
 
-    playerData->unk_2F4[1]++;
+    playerData->partnerUsedTime[1]++;
```
ethteck, line 77: for this array, are the indexes partner IDs? if so, we should use the enum here instead of just [1] (applies for all world/partner/*.c

ethteck (changes requested):
Great stuff! Just a few things I noticed

src/165490.c
HailSanta: good suggestion, but that code is in a NON_EQUIVALENT block and im reluctant to interefere with it until someone can match it

include/common_structs.h
HailSanta: thought about switchData, but none of the other union members have Data as a suffix. could just spell switch wrong like the typical java class/clazz thing, but i think its unclear to use swetch/swotch/whatever as opposed to just dropping the vowel.

ethteck: yeah I agree. zwitch is possible too but kinda goofy and not ideal because it starts with a different letter

nanaian (approved):
yum

ethteck (approved):

PR #678, opened 2022-04-07
Match parakarry func

ThreeMustyFears:


BowserSlug:
🚀 This PR matches 1 function (+0.13%) on `us`.

⚠️ This PR introduces 2 warnings:
- src/world/partner/kooper.c:10: warning: redundant redeclaration of `start_bounce_b' in same scope

nanaian (approved):
Damn, that's a big func. GJ on the match even if we have that icky goto to clean up at a later stage

PR #692, opened 2022-04-14
combining various AI and item check imports

HailSanta:


BowserSlug:
🚀🚀🚀🚀🚀🚀🚀 This PR matches 7 functions (+0.12%) on `us`.

✅ This PR fixes 1 warnings!

ethteck (approved):
Awesome stuff

src/world/common/atomic/MakeConsumableChoice.inc.c
```diff
@@ -0,0 +1,20 @@
+#include "common.h"
+
+ApiStatus N(BuildItemChoiceList)(Evt* script, s32 isInitialCall) {
+    Bytecode* args = script->ptrReadPos;
+    s32* allowedItemList = (s32*)evt_get_variable(script, *args++);
+    s32 i;
+
+    if (allowedItemList != NULL) {
+        for (i = 0; allowedItemList[i] != ITEM_NONE; i++) {
+            N(ItemChoiceList)[i] = allowedItemList[i];
```
nanaian, line 10: Should `N(ItemChoiceList)` not be declared near this file? Where's it come from?
Also what's the reasoning for making this atomic/ when its a single function?

HailSanta: I'm not finished with the stuff in this commit, I just kinda burned out on it for the time being. When completed, MakeConsumableChoice + MakeKeyChoice + ItemChoice_PartA (in that order) should be combined and a #define toggle should be used for enabling MakeKeyChoice, as its not always present. However, I was not able to dedupe the functions in ECAA80 which would allow me to do this.

PR #694, opened 2022-04-16
monty mole AI + various enums

HailSanta:


BowserSlug:
🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀 This PR matches 11 functions (+0.14%) on `us`.

include/evt.h
```diff
@@ -139,11 +139,30 @@ typedef s32 ApiStatus;
...
 
-typedef enum EventGroup {
-    EVT_GROUP_0     = 0x00,
-    EVT_GROUP_B     = 0x0B,
-    EVT_GROUP_1B    = 0x1B,
-    EVT_GROUP_EF    = 0xEF
-} EventGroup;
+enum EventGroupFlags {
+    EVT_GROUP_00    = 0x00,
+    EVT_GROUP_0B     = 0x0B, // 8 | 4 | 1
```
nanaian, line 144:
```suggestion
    EVT_GROUP_0B    = 0x0B, // 8 | 4 | 1
```

include/evt.h
```diff
@@ -139,11 +139,30 @@ typedef s32 ApiStatus;
...
+    EVT_GROUP_1B    = 0x1B, // 10 | 8 | 4 | 1
+    EVT_GROUP_EF    = 0xEF, // ~10
+    EVT_GROUP_01    = 0x01,
+    EVT_GROUP_02    = 0x02,
+    EVT_GROUP_04    = 0x04,
+    EVT_GROUP_08    = 0x08,
+    EVT_GROUP_10    = 0x10,
+};
```
nanaian, line 152: Those of these that have more than one bit set should probably? be changed to OR together the flags at the point of use

src/world/partner/bombette.c
```diff
@@ -181,15 +181,15 @@ void func_802BD6DC_31842C(Npc* npc) {
 }
 
 s32 world_bombette_can_use_ability(void) {
-    if (gPartnerActionStatus.actionState.b[0] != 0) {
+    if (gPartnerActionStatus.partnerActionState != 0) {
```
nanaian, line 184:
```suggestion
    if (gPartnerActionStatus.partnerActionState != PARTNER_ACTION_NONE) {
```

nanaian (changes requested):
TYSM for getting rid of PartnerActionStatus `actionStatus.b[]` 🙏 
That monty mole AI state enum is awesome too

include/evt.h
HailSanta: Not yet. It seems the flags are used individually within the event manager and only the first four "groups" are used at "point of use" in scripts -- that is, only the first four values of the enum appear in SetGroup() commands (0, B, 1B, and EF). It will be useful and probably more meaningful to deduce what these are and give them special semantic names. At a later time, we can revisit and decide if that still makes sense or whether its better to write them all with individual flags OR'd together.

HailSanta:
> TYSM for getting rid of PartnerActionStatus `actionStatus.b[]` pray That monty mole AI state enum is awesome too

I plan to continue in like fashion with the other AI.

PR #699, opened 2022-04-19
tik_05 functions

kklions1:


BowserSlug:

src/world/area_tik/tik_05/8750D0.c
```diff
@@ -10,6 +10,26 @@
 
 #include "world/common/CheckItemFlags40.inc.c"
 
+#include "battle/battle.h"
+#include "effects.h"
+
+#ifndef NON_MATCHING
```
ethteck, line 16: this would be ifdef, not ifndef

src/world/area_tik/tik_05/8750D0.c
```diff
@@ -10,6 +10,26 @@
...
+
+void play_model_animation(s32, s32);                     /* extern */
```
ethteck, line 18:
```suggestion
void play_model_animation(s32, s32);
```

ethteck: also, this function def should probably go into functions.h

src/world/area_tik/tik_05/8750D0.c
```diff
@@ -10,6 +10,26 @@
...
+extern s32 D_000001E4;
+
+ApiStatus func_8024093C_8756FC(Evt* script, s32 isInitialCall) {
+    Entity* entity = get_entity_by_index(evt_get_variable(script, *script->ptrReadPos));
```
ethteck, line 22: api funcs should define an "args" temp (see other script functions) and use that (*args++) rather than use script->ptrReadPos directly

src/world/area_tik/tik_05/8750D0.c
```diff
@@ -10,6 +10,26 @@
...
+
+    if (entity == 0) {
```
ethteck, line 24:
```suggestion
    if (entity == NULL) {
```

src/world/area_tik/tik_05/8750D0.c
```diff
@@ -10,6 +10,26 @@
...
+        return ApiStatus_BLOCK;
+    }
+
+    play_model_animation(entity->virtualModelIndex, (s32) &D_000001E4);
```
ethteck, line 28: This cast should probably just be removed for now. The warning in this case is actually useful, since we need to not forget to figure out how to handle this arg

src/world/area_tik/tik_05/8750D0.c
```diff
@@ -10,6 +10,26 @@
 
 #include "world/common/CheckItemFlags40.inc.c"
 
+#include "battle/battle.h"
```
ethteck, line 13: these main includes should go at the top of the c file

src/sprite.c
```diff
@@ -345,7 +345,27 @@ s32 func_802DE5C8(s32 arg0) {
...
+    /* 0x00 */ unk802DE748_2** unk_00;
+    /* 0x04 */ char unk_04[0x10];
+} unk802DE748; // size = 0x14
+extern unk802DE748 D_802DFA4C[2];
+
+s32 func_802DE748(s32 arg0, s32 arg1) {
+    unk802DE748_2** temp_v1 = D_802DFA4C[arg0].unk_00;
+
+    if (temp_v1 == NULL) {
+        return -1U;
```
ethteck, line 364:
```suggestion
        return -1;
```

src/sprite.c
```diff
@@ -345,7 +345,27 @@ s32 func_802DE5C8(s32 arg0) {
...
+
+typedef struct
+{
+    /* 0x00 */ unk802DE748_2** unk_00;
+    /* 0x04 */ char unk_04[0x10];
+} unk802DE748; // size = 0x14
+extern unk802DE748 D_802DFA4C[2];
+
+s32 func_802DE748(s32 arg0, s32 arg1) {
+    unk802DE748_2** temp_v1 = D_802DFA4C[arg0].unk_00;
```
ethteck, line 361: This function and the structs are super messy, unfortunately. It'd be good if you could spend some time cleaning this up a bit more and trying to figure out what these structs are. Also, the brace style for the second struct is wrong (there's an open brace on its own line) , and in the first struct, there's not a space between the word struct and the opening brace

ethteck (changes requested):

src/world/area_tik/tik_05/8750D0.c
```diff
@@ -10,6 +13,23 @@
 
 #include "world/common/CheckItemFlags40.inc.c"
 
+#ifdef NON_MATCHING
```
nanaian, line 16: Could you comment what the issue is preventing you from matching this?

PR #703, opened 2022-05-05
Misc decomp 53

ethteck:
in the truest sense of the term

some decomp, data cleanup, splat update, yadda yaddda


src/190B20.c
```diff
@@ -3502,7 +3502,7 @@ void btl_update_ko_status(void) {
     player->koDuration = player->debuffDuration;
     if (player->koDuration > 0) {
         player->koStatus = STATUS_DAZE;
-        ((s32*)player->debuffIcon->data)[15] = player->koDuration;
+        ((DisableXFXData*)player->debuffEffect->data)->unk_3C = player->koDuration;
```
nanaian, line 3505: We can totally name this `unk_3C` field `koDuration`

src/battle/action_cmd/air_raid.c
```diff
@@ -60,7 +67,168 @@ ApiStatus func_802A9000_429320(void) {
 
 #include "common/MashActionCommandInit.inc.c"
 
-INCLUDE_ASM(s32, "battle/action_cmd/air_raid", func_802A9294_4295B4);
+//INCLUDE_ASM(s32, "battle/action_cmd/air_raid", func_802A9294_4295B4);
```
nanaian, line 70:
```suggestion
```

src/sprite.c
```diff
@@ -272,26 +272,28 @@ void spr_init_sprites(s32 playerSpriteSet) {
     }
 
     for (i = 0; i < ARRAY_COUNT(spr_playerCurrentAnimInfo); i++) {
-        struct spr_playerCurrentAnimInfo* d = &spr_playerCurrentAnimInfo[i];
-        d->componentList = 0;
-        d->animID = -1;
+        PlayerCurrentAnimInfo* animInfo = &spr_playerCurrentAnimInfo[i];
+
+        animInfo->componentList = 0;
```
nanaian, line 277: Should this be `NULL`?

nanaian (changes requested):

src/190B20.c
ethteck: it's definitely some sort of duration, but there's no indication that it's specifically a ko duration just because that's used for it in this instance

ethteck: clover was gonna work through these some point soon anyway, btw

PR #710, opened 2022-05-16
battle camera

Unnunu:


BowserSlug:
🚀🚀🚀🚀🚀 This PR matches 5 functions (+0.28%) on `us`.

src/16F740.c
```diff
@@ -131,13 +131,13 @@ void btl_state_update_begin_player_turn(void) {
...
     if (gBattleState2 == BATTLE_STATE2_UNK_0) {
-        func_8024E40C(2);
+        btl_cam_use_preset(BTL_CAM_PRESET_C);
         btl_cam_move(5);
         gBattleState2 = BATTLE_STATE2_UNK_64;
     }
 
     if (gBattleState2 != BATTLE_STATE2_UNK_14) {
-        if (gBattleState2 == BATTLE_STATE2_UNK_64 && (func_8024E584() != 0)) {
+        if (gBattleState2 == BATTLE_STATE2_UNK_64 && (btl_cam_is_moving_done())) {
```
nanaian, line 140:
```suggestion
        if (gBattleState2 == BATTLE_STATE2_UNK_64 && btl_cam_is_moving_done()) {
```

src/battle_cam.c
```diff
@@ -0,0 +1,2562 @@
+#include "common.h"
+#include "script_api/battle.h"
+
+#define LERP(a, b, alpha) ((a) * (alpha) + (b) * (1.0f-(alpha)))
```
nanaian, line 4: Given this computes alpha twice could we maybe turn it into a static function

nanaian (changes requested):
I'm not sure I like the PascalCase of of the `BattleCam_` variables but looks good otherwise :+1:

src/battle_cam.c
ethteck: I don't want us to add any static functions

ethteck:
the battle cam file locals were already in the weird casing, so we can just keep them as-is and change it in a future PR

ethteck (approved):
Great stuff :) once alex's change is added, we can merge

PR #726, opened 2022-06-09
hud-script-disasm

HailSanta:


ethteck (approved):

BowserSlug:
⚠️ This PR introduces 3 warnings:
- src/C50A0.c:173: warning: assignment from incompatible pointer type
- src/C50A0.c:174: warning: assignment from incompatible pointer type
- src/C50A0.c:560: warning: passing arg 2 of `sparkle_script_init' from incompatible pointer type
- src/C50A0.c:656: warning: passing arg 2 of `sparkle_script_init' from incompatible pointer type
- src/C50A0.c:681: warning: passing arg 2 of `sparkle_script_init' from incompatible pointer type

include/sparkle_script.h
```diff
@@ -0,0 +1,20 @@
+#include "common.h"
+
+typedef s32 SparkleScript[0];
```
nanaian, line 3: Wondering if this should be unsized ie `s32 SparkleScript[];`, I recall issues with 0-sized EvtScripts.. does this match the other *Script types?

HailSanta: This is the same as EvtScript, HudScript, and ItemScript.

PR #729, opened 2022-06-10
omo2 battle map

Unnunu:


BowserSlug:
🚀🚀🚀🚀🚀 This PR matches 5 functions (+0.02%) on `us`.

ethteck (approved):

nanaian (approved):
Awesome work with that animation include. I'm wondering if we should introduce macros similar to the EVT ones for AnimScripts/AS in the future also

PR #732, opened 2022-06-12
Mystic Miscellaneous Matches

JaThePlayer:
Also fixes some types in sprite.c/.h, though there are still tons of weird things going on there.


BowserSlug:
🚀🚀🚀🚀🚀🚀 This PR matches 6 functions (+0.04%) on `us`.

- src/battle/partner/sushie.c:110: warning: assignment makes integer from pointer without a cast
- src/battle/partner/sushie.c:121: warning: initialization makes integer from pointer without a cast
- src/battle/partner/sushie.c:124: warning: passing arg 1 of `remove_effect' makes pointer from integer without a cast
- src/battle/partner/watt.c:104: warning: passing arg 6 of `fx_bulb_glow' from incompatible pointer type
- src/sprite.c:345: warning: comparison between pointer and integer

src/battle/partner/watt.c
```diff
@@ -11,6 +11,7 @@ extern s32 D_8023C1C8;
 extern EffectInstance* D_8023C1CC;
 extern EffectInstance* D_8023C1D0;
 extern EffectInstance* D_80239A0C_7054FC;
+extern EffectInstance* D_8023C1B0;
```
ethteck, line 14: Could you move this to be the first extern in the list (so it's in memory order)? B0, B4, etc

src/sprite.c
```diff
@@ -354,7 +392,7 @@ s32 func_802DE748(s32 arg0, s32 arg1) {
         return -1;
     }
 
-    return componentList[arg1]->unk_4F;
+    return (u8)componentList[arg1]->unk_4C;
```
ethteck, line 395: why was this changed?

ethteck: actually, maybe it's correct. I'm just wondering if there was something to corroborate it

ethteck: ohhhh it's the ```
if (arg2 != 0) {
                    component->unk_4C |= 0x10000000;
                } else {
                    component->unk_4C &= ~0xF0000000;
                }
                ```
                I think this indicates it's actually a bitfield or something, but we can worry about that later. Though, instead of the u8 casts, we should probably be doing & 0xFF on that field.

ethteck (changes requested):
Great jorb!

src/sprite.c
ethteck: (there's another u8 cast or two on that field elsewhere in this PR, so this applies to all of those)

JaThePlayer: changed to `& 0xFF`

src/battle/partner/watt.c
```diff
@@ -91,7 +92,26 @@ ApiStatus func_8023859C_70408C(Evt* script, s32 isInitialCall) {
...
+        fx_bulb_glow(0, x, y, z, 1.0f, &D_8023C1B0);
+    }
+
+    script->functionTemp[0]--;
+    if (script->functionTemp[0] == 0) {
+        ((BulbGlowFXData*)D_8023C1B0->data)->unk_14 = 5;
+        return ApiStatus_DONE2;
+    }
+
+    return ApiStatus_BLOCK;
```
nanaian, line 113: I'd prefer this as an else

src/sprite.c
```diff
@@ -226,7 +226,24 @@ void spr_init_component_anim_state(SpriteComponent* comp, s16*** anim) {
     comp->scale.z = 1.0f;
 }
 
+
+// matches, but has a very questionable cast, spr_init_component_anim_state args are probably wrong
```
nanaian, line 230: Does this func actually match? If so it might be wise to remove the asm/NON_EQUIVALENT wrapper and just have this as a `// TODO improve match`

JaThePlayer: yea it does match, no clue why I added the NON_EQUIVALENT thing here tbh lol

PR #735, opened 2022-06-15
Tiny funcs & cleanup & entity

ethteck:


BowserSlug:
🚀🚀🚀🚀🚀🚀🚀🚀🚀 This PR matches 9 functions (+0.05%) on `us`.

✅ This PR fixes 9 warnings!

src/17D6A0.c
```diff
@@ -94,11 +98,109 @@ void free_popup(PopupMessage* popup) {
 
 INCLUDE_ASM(s32, "17D6A0", func_8024EFE0);
 
-INCLUDE_ASM(s32, "17D6A0", func_8024F394);
+void func_8024F394(void* data) {
```
nanaian, line 101: Given that this arg is interpreted as `PopupMessage*` should we type `data` as it too?
```suggestion
void func_8024F394(PopupMessage* data) {
```

src/17D6A0.c
```diff
@@ -94,11 +98,109 @@ void free_popup(PopupMessage* popup) {
...
+    PopupMessage* popup = data;
+    Message* message = popup->message;
+    s32 found = FALSE;
+    s32 i;
+
+    for (i = 0; i < popup->messageIndex; i++, message++) {
```
nanaian, line 107: I think we should initialise `message` here too if we're going to be incrementing it here also
```suggestion
    for (i = 0, message = popup->message; i < popup->messageIndex; i++, message++) {
```

src/a5dd0_len_114e0.c
```diff
@@ -1629,7 +1636,62 @@ s32 entity_get_collision_flags(Entity* entity) {
...
+                case ENTITY_TYPE_HEALING_BLOCK:
+                case ENTITY_TYPE_1C:
+                case ENTITY_TYPE_SAVE_POINT:
+                case ENTITY_TYPE_SUPER_BLOCK:
+                    entity = get_entity_by_index(entityIdx);
+                    entity->flags |= ENTITY_FLAGS_BLOCK_BEING_HIT;
+                    ret = TRUE;
+            }
+            break;
+    }
```
nanaian, line 1692: this is really cool

src/a5dd0_len_114e0.c
```diff
@@ -1776,8 +1838,135 @@ s32 func_80111790(void) {
...
+    s32 listIndex;
+    Entity* entity;
+    s32* a;
+
+    va_start(ap, bp);
+    bpPtr = &bp;
+    *bpPtr = bp;
+
+    load_area_specific_entity_data();
+
```
nanaian, line 1860:
```suggestion
```

src/a5dd0_len_114e0.c
```diff
@@ -1776,8 +1838,135 @@ s32 func_80111790(void) {
...
+    a = &CreateEntityVarArgBuffer[3];
+
+    *a-- = 0;
+    *a-- = 0;
+    *a = 0;
+
+    for (listIndex = 3; listIndex > 0; listIndex--) {
+        s32 arg = va_arg(ap, s32);
+
+        if (arg == 0x80000000) {
```
nanaian, line 1876: I believe we have a macro for this constant

nanaian (changes requested):
bro really said "tiny funcs" and dropped this PR

src/17D6A0.c
ethteck: The signature needs to be this way for 2 reasons: the function is called indirectly via an argument that types the first arg of the function generically so other args can fit into it, and secondly, it doesn't match unless we create a temp for the first arg (which strongly indicates they were casting a void* to the proper type)

src/17D6A0.c
ethteck: I disagree - I think it makes the loop confusing to grok, and we have dozens of other loops like this in the game for which we don't initialize the struct pointer there either

PR #741, opened 2022-07-13
entities

Unnunu:


BowserSlug:
🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀 This PR matches 74 functions (+1.01%) on `us`.

✅ This PR fixes 85 warnings!

include/common_structs.h
```diff
@@ -589,26 +559,195 @@ typedef struct BlueWarpPipeData {
...
+    /* 0x04 */ Vec3f rotation;
+    /* 0x10 */ s32 unk_10;
+    /* 0x14 */ f32 spinSpeed;
+    /* 0x18 */ s32 unk_18;
+    /* 0x1C */ char unk_1C[0x0C];
+    /* 0x28 */ s16 unk_28;
+    /* 0x2A */ s16 unk_2A;
+    /* 0x2C */ s16 unk_2C;
+    /* 0x30 */ Mtx unk_30;
+} SpinningFlowerData; // size = 0x70
```
nanaian, line 708:
```suggestion
typedef struct SpinningFlowerData {
    /* 0x00 */ s16 unk_00;
    /* 0x02 */ s8 state;
    /* 0x04 */ Vec3f rotation;
    /* 0x10 */ s32 unk_10;
    /* 0x14 */ f32 spinSpeed;
    /* 0x18 */ s32 unk_18;
    /* 0x1C */ char unk_1C[0x0C];
    /* 0x28 */ s16 unk_28;
    /* 0x2A */ s16 unk_2A;
    /* 0x2C */ s16 unk_2C;
    /* 0x30 */ Mtx unk_30;
} SpinningFlowerData; // size unknown
```

include/entity.h
```diff
@@ -0,0 +1,122 @@
...
+    ENTITY_MODEL_SCRIPT_OP_ClearFlags,
+    ENTITY_MODEL_SCRIPT_OP_DrawImage,
+};
+
+#define es_End ENTITY_SCRIPT_OP_End,
+#define es_Jump(script) ENTITY_SCRIPT_OP_Jump, (s32)script,
+#define es_Call(func) ENTITY_SCRIPT_OP_Call, (s32)func,
+#define es_SetCallback(func, time) ENTITY_SCRIPT_OP_SetCallback, time, (s32)func,
+#define es_Goto(labelId) ENTITY_SCRIPT_OP_Goto, labelId,
+#define es_Restart ENTITY_SCRIPT_OP_Goto, 0,
```
nanaian, line 38: Unsure what I think of having a pseudoinstruction like this. Evt scripts have a similar case where `evt_Loop(0)` is an infinite loop, but we decided against adding an `evt_LoopForever`-like macro there. It does create nice parity with `ems_Restart` however

src/entity/HiddenPanel.c
```diff
@@ -0,0 +1,372 @@
...
+    yaw = clamp_angle(gCameras[CAM_DEFAULT].currentYaw + 45.0f);
+    if (yaw < 90.0f || yaw >= 180.0f && yaw < 270.0f) {
+        rotAngle = entity->rotation.z;
+        flipAxis = 1;
+    } else {
+        rotAngle = entity->rotation.x;
+        flipAxis = 0;
+    }
+
+    switch (data->state) {
```
nanaian, line 119: Can you add a `// TODO: enum` to the `state` field declaration?

include/common_structs.h
```diff
@@ -411,12 +411,12 @@ struct Entity;
 
 // BEGIN ENTITY-SPECIFIC STRUCTS
```
nanaian, line 412: Wondering if it would be wise to move this section to `entity.h`

src/entity/model/ArrowSign.c
```diff
@@ -0,0 +1,89 @@
+#include "common.h"
```
nanaian, line 1: Need to have a discussion about how to make some kind of splat segment for these model files, they're arguably assets (embedded Vtx, images, Gfx)

src/entity/HiddenPanel.c
```diff
@@ -0,0 +1,372 @@
...
+    entity->renderSetupFunc = entity_HiddenPanel_setupGfx;
+    data->pickupVar = 0xFFFF;
+    data->initialY = entity->position.y;
+    data->modelID = CreateEntityVarArgBuffer[0];
+    data->itemID = CreateEntityVarArgBuffer[1];
+    data->needSpawnItem = TRUE;
+    data->spawnedItemIndex = -1;
+
+    if (data->itemID == 0) {
+        data->itemID = 0x15C;
```
ethteck, line 306: is there an item for this in enums.h?

src/entity/model/ArrowSign.c
ethteck: splat can extract the individual display lists and vtx currently, but I'm not sure if it's worth it. none of this looks very asset-y to me other than the images, which area already handled. but we can def talk about it

src/entity/model/BellbellPlant_gfx.c
```diff
@@ -0,0 +1,348 @@
+#include "common.h"
+
+extern unsigned char D_0A001090_E6C240[];
+
+Vtx D_0A000000_E6B1B0[] = {
```
ethteck, line 5: This one I would definitely make a vtx splat type and #include the inc.c splat creates. also for the non-rectangle vertices below

ethteck (changes requested):
Amazing! Thanks for all your hard work on this.

include/common_structs.h
Unnunu: all these structs are not referenced anywhere outside entity folder so it might be good idea

include/entity.h
Unnunu: `es_Goto(0)` without corresponding `es_Label(0)` looks a bit confusing

nanaian (approved):

PR #745, opened 2022-08-06
effect organization

HailSanta:


BowserSlug:
🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀 This PR matches 10 functions (+0.03%) on `us`.

✅ This PR fixes 22 warnings!

src/world/common/atomic/SuperBlock.inc.c
```diff
@@ -311,9 +311,9 @@ ApiStatus N(SuperBlock_AnimateEnergyOrbs)(Evt* script, s32 isInitialCall) {
     }
 
     for (i = 0; i < SUPER_BLOCK_NUM_ORBS; i++) {
-        ((BlurFlameEffectData*)userData->orbEffects[i]->data)->position.x = userData->posX[i];
-        ((BlurFlameEffectData*)userData->orbEffects[i]->data)->position.y = userData->posY[i];
-        ((BlurFlameEffectData*)userData->orbEffects[i]->data)->position.z = userData->posZ[i];
+        ((BlurFlameEffectData*)userData->orbEffects[i]->data.unk32)->position.x = userData->posX[i];
+        ((BlurFlameEffectData*)userData->orbEffects[i]->data.unk32)->position.y = userData->posY[i];
+        ((BlurFlameEffectData*)userData->orbEffects[i]->data.unk32)->position.z = userData->posZ[i];
```
ethteck, line 316: is there not a union entry for BlurFlameEffectData? (also shouldn't it be FXData)

ethteck (changes requested):
Awesome!

src/world/common/atomic/SuperBlock.inc.c
HailSanta: Fixed in latest commit.

ethteck (approved):

include/common_structs.h
```diff
@@ -208,7 +212,7 @@ typedef struct Npc {
     /* 0x0C0 */ s8 unk_C0;
     /* 0x0C1 */ s8 paletteCount;
     /* 0x0C2 */ char unk_C2[2];
-    /* 0x0C4 */ s32* spritePaletteList;
+    /* 0x0C4 */ PAL_PTR* spritePaletteList;
```
nanaian, line 215: This has gone from s32* to u16** - is the double-indirection correct?

nanaian (changes requested):

include/common_structs.h
HailSanta: Yes. This is a pointer to a list of palette pointers in the sprite data terminated by -1. Not sure how the previous definition was OK, unless it was being cast. Going back to s32 is not OK.

PR #747, opened 2022-08-09
Effect work and rodata migration

ethteck:


BowserSlug:
🚀🚀🚀🚀🚀🚀🚀🚀 This PR matches 8 functions (+0.18%) on `us`.

src/effects/bombette_breaking.c
```diff
@@ -53,4 +202,53 @@ void bombette_breaking_render(EffectInstance* effect) {
     retTask->renderMode |= RENDER_TASK_FLAG_2;
 }
 
-INCLUDE_ASM(s32, "effects/bombette_breaking", bombette_breaking_appendGfx);
+void bombette_breaking_appendGfx(void* effect) {
```
nanaian, line 205:
```suggestion
void bombette_breaking_appendGfx(EffectInstance* effect) {
```
and then remove the casts

ethteck: nah, all the appendGfx functions take a void* (this allows them to be typed properly) and there's evidence to corroborate this in that we often need to make a temp var for a0 (cast the void* to EffectInstance*). Later on we can maybe have a macro that converts the void* effect arg to an EffectInstance*

PR #753, opened 2022-08-28
Fix some stuff for Mac installation

EllipticEllipsis:
- Updated `configure.py` to use GCC 12
- Remove obsolete capstone requirement
- Add a note to the installation readme about the weird `md5sha1sum` error (could make this a `<details>` if it's still too intrusive, of course it would be preferable to just solve the problem anyway)
- Changed the intellisense mode, since otherwise on non-x86 on startup it gives warnings about changing the mode automatically (there's probably a way to stop it doing this, but I have yet to find it)

### DIFF
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -26,7 +26,7 @@
             ],
             "cStandard": "c89",
             "cppStandard": "c++17",
-            "intelliSenseMode": "gcc-x86"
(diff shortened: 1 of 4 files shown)


EllipticEllipsis:
Oh, one other thing is that probably `install.sh` should use splat's requirements file too rather than just repeating everything, not sure if there's a good way to do this beyond adding it to all 5 OSes.

PR #809, opened 2022-10-08
[Effects] Gfx data + some light funcs

mrshigure:


BowserSlug:
🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀 This PR matches 11 functions (+0.15%) on `us`.

include/effects.h
```diff
@@ -142,6 +142,8 @@ enum EffectID {
     EFFECT_86                       = 0x86,
 };
 
+typedef struct EffectInstance EffectInstance;
```
ethteck, line 145: wut dis do

include/effects.h
```diff
@@ -2072,13 +2143,13 @@ typedef union {
     struct Effect86FXData*              unk_86;
 } EffectData;
 
-typedef struct EffectInstance {
+struct EffectInstance {
```
ethteck, line 2146: this should still be the way it beed before (I think)

ver/us/undefined_syms.txt
```diff
@@ -62,21 +62,6 @@ D_09002020 = 0x09002020;
 D_09004040 = 0x09004040;
 D_09008BE0 = 0x09008BE0;
```
ethteck, line 64: can more syms be removed from here, given you did more gfx? or maybe there aren't any added here for those effects yet?

ethteck (dismissed):

include/effects.h
mrshigure: `LightningFXData` has references to `EffectInstance`, so we need to declare it beforehand.

include/effects.h
mrshigure: Since now we're declaring `EffectInstance` at the top of the file, the compiler complains if we try to redefine the type here, so I've removed the typedef.

ver/us/undefined_syms.txt
mrshigure: Yep, this was the only effect with undefined syms in this PR. We just have a few more (around 4) to clean.

include/effects.h
ethteck: I think preferably we do `struct EffectInstance*` in LightningFXData

nanaian (approved):

PR #823, opened 2022-10-15
BSS And Libultra

ethteck:
Using the Wiseguy method :tm:, we're no longer building c files in segments containing scrambled bss with -fno-common. This allows us to define bss vars in these files and still maintain their original rom order. After doing this, I was able to much easier match some more libultra code (copied from [ultralib](https://github.com/decompals/ultralib)).


BowserSlug:
🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀 This PR matches 10 functions (+0.12%) on `us`.

✅ This PR fixes 11 warnings!

nanaian (approved):
huge

PR #827, opened 2022-10-16
3 more gfx functions

Unnunu:


BowserSlug:
🚀🚀🚀 This PR matches 3 functions (+0.22%) on `us`.

src/C50A0.c
```diff
@@ -2222,19 +2799,19 @@ void func_801363A0(ItemEntity* itemEntity) {
         case 10:
             if (!(itemData->typeFlags & ITEM_TYPE_FLAG_BADGE)) {
                 if (!(itemEntity->flags & ITEM_ENTITY_FLAGS_4000000) || (itemEntity->pickupMsgFlags & 0x4)) {
-                    itemMsg = MSG_Menus_0058;
+                    itemMsg = MESSAGE_ID(0x1D, 0x058);
```
ethteck, line 2802: Is there a reason you went back to the more primitive macros? Do the MSG_ defines not work?

ethteck (approved):

src/C50A0.c
Unnunu: I didn't decompile this function,I just copied old code from decomp.me and made some score improvement. Is it ok to leave it as is?

ethteck:
Yesh

src/C50A0.c
HailSanta: Would be much better to keep the lines with MSG_* and WINDOW_ID_*.

Unnunu: done

nanaian (approved):

PR #834, opened 2022-10-23
[Effects] Matching a couple of functions

mrshigure:


BowserSlug:
🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀 This PR matches 33 functions (+0.70%) on `us`.

src/effects/something_rotating.c
```diff
@@ -1,6 +1,66 @@
 #include "common.h"
 #include "effects_internal.h"
 
+s32 D_E0116C60[] = { 0x09004458 };
```
ethteck, line 4: These should be made into symbols and added to undefined syms instead of being raw pointers (same as the ones below), and the arrays should be typed as Gfx*[]

src/effects/something_rotating.c
```diff
@@ -66,7 +126,111 @@ EffectInstance* something_rotating_main(
...
+                    part->unk_25 = D_E0116D08[ARRAY_COUNT(D_E0116D08) - 1];
+                }
+
+                if (unk_14_2 < ARRAY_COUNT(D_E0116D3C)) {
+                    factor = D_E0116D3C[unk_14_2];
+                } else {
+                    factor = D_E0116D3C[ARRAY_COUNT(D_E0116D3C) - 1];
+                }
+
+                factor *= 0.0039215686; // 1 / 255
```
ethteck, line 200: does 1.0 / 255 work? you might need to fudge the numbers a tiny bit, but that seems better than the strange constant

src/effects/star_spirits_energy.c
```diff
@@ -23,4 +114,63 @@ void star_spirits_energy_render(EffectInstance* effect) {
...
+        }
+    }
+
+    gSPPopMatrix(gMasterGfxPos++, G_MTX_MODELVIEW);
+
+    if (unk_04 == 2 || unk_04 == 3) {
+        alpha = 0;
+
+        if (data->unk_70 == 1 || data->unk_70 == 100) {
+            if (data->unk_74 > 2) {
```
ethteck, line 154: it'd be nice to && this and the following condition together, since they're both checking if the thing is within a certain range. also applies in the second block below with 0 and 9

ethteck (changes requested):
"a couple" pshh... wow, this is a ton of g00ds. great work!

src/effects/star_spirits_energy.c
mrshigure: Putting the two together wasn't working, but it seems it works if I include the outer condition as well:
```c
(data->unk_70 == 1 || data->unk_70 == 100) && data->unk_74 > 2 && data->unk_74 < 11
```

src/effects/something_rotating.c
mrshigure: It doesn't, unfortunately. 1.0 / 255 is more precise: 0.00392156862745098

ethteck: Ahhh okh

ethteck (approved):

nanaian (approved):

PR #841, opened 2022-10-30
non-world data migration

Unnunu:


BowserSlug:
- src/17D6A0.c:2725: warning: assignment from incompatible pointer type

ethteck (changes requested):
great work as always!

ethteck (approved):

ethteck:
Despite matching a function,  B. Slug didn't comment about it because the number of *total* functions decreased in this PR. Two symbols that were mistakenly being detected as functions previously no longer are, so therefore the total number of "matched functions" decreased by one.

src/17D6A0.c
```diff
@@ -2387,6 +2405,133 @@ s32 bActorMessages[] = {
...
+    { \
+        (u32)battle_partner_##name##_ROM_START, \
+        (u32)battle_partner_##name##_ROM_END, \
+        battle_partner_##name##_VRAM, \
+        &battle_partner_##name, \
+        Y \
+    }
+
+PartnerDMAData bPartnerDmaTable[] = {
+    { },
```
nanaian, line 2521: Can you write these as `{}` instead? (no space)

src/17D6A0.c
```diff
@@ -2387,6 +2405,133 @@ s32 bActorMessages[] = {
...
+    gsSP1Triangle(0, 2, 3, 0),
+    gsDPPipeSync(),
+    gsSPEndDisplayList(),
+
+};
+EntityModelScript D_80283EE8 = STANDARD_ENTITY_MODEL_SCRIPT(D_80283DF8, RENDER_MODE_ALPHATEST);
+
+s32 D_80283F04_padding = 0;
+s32 D_80283F08_padding = 0;
+s32 D_80283F0C_padding = 0;
```
nanaian, line 2509: Perhaps make these `static`?
@ethteck I'm thinking we should maybe make a `DATA_PADDING(addr, size)` macro for this purpose

src/starpoint.c
```diff
@@ -0,0 +1,151 @@
+#include "common.h"
+#include "entity.h"
+
+Lights1 starpoint_lights1 = gdSPDefLights1(0, 0, 0, 0, 0, 0, 0, 0, 0);
+#include "starpoint/starpoint.png.inc.c"
+#include "starpoint/load_starpoint.gfx.inc.c"
+#include "starpoint/starpoint.vtx.inc.c"
+Gfx D_802A20D0_7A9A70[] =
+{
```
nanaian, line 9: Extend this change to all structs here please
```suggestion
Gfx D_802A20D0_7A9A70[] = {
```

nanaian (changes requested):

src/17D6A0.c
ethteck: I would rather just keep the padding temps personally

ethteck: explicit padding temps*

nanaian: That's fine, was just suggesting so they're easier to find and clean up later but tbh we can just look for "pad" and find them all

PR #852, opened 2022-11-04
battle partners

Unnunu:


BowserSlug:
🚀🚀🚀 This PR matches 3 functions (+0.19%) on `us`.

src/common/CalculateRotationZ.inc.c
```diff
@@ -1,7 +1,19 @@
 #include "common.h"
```
HailSanta, line 1: I think you should call this file CalculateArcsinDeg and rename the function CalculateRotationZ -> CalculateArcsinDeg accordingly.

ver/us/splat.yaml
```diff
@@ -7595,73 +7595,62 @@ segments:
     vram: 0x80238000
     subsegments:
     - [0x6F10E0, c, goombario]
-    - [0x6F2290]
```
ethteck, line 7598: These markers that explicitly say where the c code ends are still useful imo, even if they're not totally necessary

ver/us/splat.yaml
```diff
@@ -7595,73 +7595,62 @@ segments:
...
     - [0x703AF0, c, watt]
-    - [0x704A80, data]
   - name: battle_partner_sushie
     dir: battle/partner
     type: code
     start: 0x707CA0
     vram: 0x80238000
     subsegments:
     - [0x707CA0, c, sushie]
-    - [0x708B70, data]
```
ethteck, line 7638: so this one would just be `- [0x708B70]` and 70BD00 can go away

ethteck (changes requested):

PR #859, opened 2022-11-15
sprite_shading_profiles & obk_07

ethteck:


BowserSlug:
🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀 This PR matches 23 functions (+0.07%) on `us`.

tools/build/sprite/sprite_shading_profiles.py
```diff
@@ -0,0 +1,167 @@
...
+        f.write("#define SHADING_PROFILES_H\n")
+        f.write("/* This file is auto-generated from sprite_shading_profiles.json. Do not edit. */\n\n")
+        f.write("enum ShadingProfile {\n")
+        f.write("    SHADING_NONE = 0xFFFFFFFF,\n")
+
+        for g, group in enumerate(groups):
+            for i, profile in enumerate(group.profiles):
+                f.write(f"    SHADING_{profile.name.upper()} = 0x{(g << 16) | i:08X},\n")
+        f.write("};\n")
+        f.write("#endif // SHADING_PROFILES_H")
```
nanaian, line 108: Can you add a terminating \n

tools/build/sprite/sprite_shading_profiles.py
```diff
@@ -0,0 +1,167 @@
...
+                data_table.extend(struct.pack(END + "h", light.pos[2]))
+                data_table.extend(struct.pack(END + "f", light.falloff))
+                data_table.append(0)
+                data_table.append(0)
+                group_data_offset += 0x10
+        data_offset += group_data_offset
+
+    offsets_table.extend(profile_lists)
+    if matching:
+        offsets_table += b'\0' * (0x1D0 - len(offsets_table)) # Pad to 0x1D0
```
nanaian, line 148: Interesting that this is a thing we have to do, does this imply that we should have a bunch of zero groups in the input config?

tools/build/sprite/sprite_shading_profiles.py
```diff
@@ -0,0 +1,167 @@
...
+
+    final_data = offsets_table + data_table
+    if matching:
+        final_data += b'\0' * (0xE70 - len(final_data)) # Pad to 0xE70
```
nanaian, line 152: Perhaps the linker should be doing this

nanaian (changes requested):
Awesome work, a few minor suggestions

tools/build/sprite/sprite_shading_profiles.py
```diff
@@ -0,0 +1,167 @@
...
+    with open(input, "r") as f:
+        json_data = json.load(f)
+
+    groups = groups_from_json(json_data)
+
+    # Header creation
+    with open(header_out, "w") as f:
+        f.write("#ifndef SHADING_PROFILES_H\n")
+        f.write("#define SHADING_PROFILES_H\n")
+        f.write("/* This file is auto-generated from sprite_shading_profiles.json. Do not edit. */\n\n")
```
nanaian, line 100: Should use `input` here not hardcode the path

tools/build/sprite/sprite_shading_profiles.py
ethteck: That would cause more lighting data generation as well. I really think this is just padding until the actual lighting data

PR #860, opened 2022-11-16
btl_state_update_partner_menu

Unnunu:


BowserSlug:
🚀 This PR matches 1 function (+0.17%) on `us`.

include/variables.h
```diff
@@ -21,7 +21,7 @@ extern BattleStatus gBattleStatus;
 extern s32 D_800DC4D0;
 extern s32 D_800DC4D4;
 extern s32 D_800DC4D8;
-extern s32 gBattleState2;
+extern s32 gBattleSubState;
```
nanaian, line 24: Can you rename the enum and its prefix also?

src/415D90.c
```diff
@@ -14,6 +14,73 @@
 #include "sprite/npc/BattleLakilester.h"
 #include "sprite/npc/Twink.h"
 
+enum BattleSubStatesPartner {
```
nanaian, line 17: Move to enums.h or 415D90.h please

src/415D90.c
```diff
@@ -3534,60 +3610,864 @@ void btl_state_draw_player_menu(void) {
...
+        battle_menu_moveOptionCount = battleStatus->submenuMoveCount;
+        s1 = battleStatus->unk_5C[6];
+        D_802AD4A8 = s1;
+        func_802A2684();
+        gBattleSubState = BATTLE_SUB_STATE_PARTNER_FOCUS_2;
+        break;
+    case BATTLE_SUB_STATE_PARTNER_FOCUS_2:
+        numOptions = func_802A2C84();
+        if (numOptions != 0) {
+            if (numOptions == 255) {
```
nanaian, line 4167: Interesting this isn't -1. Does changing `numOptions` to 8-bit allow -1 here and not break elsewhere?

nanaian (changes requested):
Awesome, mad that this func was 0.17%. Couple comments

src/415D90.c
Unnunu: -1 and 255 are different possible results that can be returned by this kind of function. 255 means return back from submenu and -1 is for error.

PR #868, opened 2022-11-23
Misc matches & cleanup

ethteck:


BowserSlug:
🚀🚀🚀 This PR matches 3 functions (+0.05%) on `us`.

src/17D6A0.c
```diff
@@ -23,47 +23,50 @@ extern ActorBlueprint battle_partner_lakilester;
...
-    3, 11, 0, 0,
-    0, 0, 0, 0,
-    0, 0, 0, 0,
-    0, 0, 0, 0,
-    0, 0, 0, 0,
-    0, 0, 0, 0, 0, 0,
+UnkPartnerThing D_80280FC0 = {
+    10, 90,
+    {
+        50, 196619,
```
HailSanta, line 29: The second field here are player anim IDs.
e.g., 196619 = 0x1002D = ANIM_Mario_1002D

ethteck: ah nice - I thought that may be the case, but I wasn't sure. I'll add those

src/entity_model.c
```diff
@@ -1,6 +1,7 @@
 #include "common.h"
+#include "entity.h"
 
-s32 D_8014C260[] = { 0x00000000, 0x00000000 };
+EntityModelScript D_8014C260[] = { {0x00000000}, {0x00000000 }};
```
HailSanta, line 4: 0 -> ems_End ?

PR #872, opened 2022-11-26
Effect work

ethteck:
Finished one, utterly failed at the other. Help wanted on stars_shimmer


BowserSlug:
🚀🚀🚀 This PR matches 3 functions (+0.05%) on `us`.

include/effects.h
```diff
@@ -944,8 +953,8 @@ typedef struct RadialShimmerFXData {
     /* 0x1C */ f32 unk_1C;
     /* 0x20 */ f32 unk_20;
     /* 0x24 */ s32 unk_24;
-    /* 0x28 */ s32 unk_28;
-    /* 0x2C */ s32 unk_2C;
+    /* 0x28 */ s32 timeLeft;
+    /* 0x2C */ s32 lifeTime;
```
nanaian, line 957: Can you define the difference? Is lifeTime the max

src/effects/firework_rocket.c
```diff
@@ -6,6 +6,64 @@ void firework_rocket_render(EffectInstance* effect);
...
+} FireworkRocketUnk; // size = 0x8
+
+extern Gfx D_09000080_3E53D0[];
+extern Gfx D_09000158_3E54A8[];
+
+Gfx* D_E010ACC0[] = { D_09000158_3E54A8, D_09000158_3E54A8, D_09000158_3E54A8 };
+Gfx* D_E010ACCC[] = { D_09000080_3E53D0, D_09000080_3E53D0, D_09000080_3E53D0 };
+
+Color_RGB8 D_E010ACD8[] = {
+    {36, 0, 93}, {54, 216, 74}, {67, 0, 74}, {82, 220, 45}, {89, 0, 45}, {60, 190, 45}, {21, 192, 74}, {28, 171, 45},
```
nanaian, line 21: We should split these images (?) now if possible

ethteck: they're not images, they're rgb values used in the graphics func

include/effects.h
ethteck: timeLeft counts down, lifeTime counts up. most effects have these, and we decided to name them this way a while ago. we can change them at some point though if we don't like the names

PR #876, opened 2022-12-03
snaking_static effect

savaughn:
ain't much, but better than nothin :rofl:


BowserSlug:
🚀🚀 This PR matches 2 functions (+0.03%) on `us`.

src/effects/snaking_static.c
```diff
@@ -3,12 +3,142 @@
...
+            }
+        }
+        data->unk_24 = (shim_sin_deg(((data->lifeTime * 180) / data->timeLeft)) * 1024.0f);
+        if (data->unk_24 >= 256) {
+            data->unk_24 = 255;
+        }
+        data->lifeTime++;
+    }
+    data = effect->data.snakingStatic;
+    data++;
```
nanaian, line 137: Can't do + 1?

savaughn: do you mean like this?
```
data = effect->data.snakingStatic + 1;
```

ethteck: the way you have it is better imo, since it matches the way we normally use data and handle it in these loops

PR #886, opened 2022-12-13
a match for the clover

ethteck:


BowserSlug:
🚀 This PR matches 1 function (+0.04%) on `us`.

✅ This PR fixes 1 warnings!

src/world/area_obk/obk_01/BBD1A0.c
```diff
@@ -31,8 +34,179 @@ ApiStatus func_8024007C_BBD21C(Evt* script, s32 isInitialCall) {
...
+    s32 var_a0;
+    s32 temp_s0_2;
+    Model* model;
+    UnkObk01* temp_v0;
+    s32 i;
+
+    if (isInitialCall) {
+        script->functionTempPtr[1] = temp_v0 = heap_malloc(sizeof(*temp_v0));
+        temp_v0->unk_08 = (s32*) evt_get_variable(script, *args++);
+        temp_v0->unk_08[0] = (s32) temp_v0; // what is going on here?
```
nanaian, line 64: An evt variable holding a pointer to a struct isn't too uncommon

nanaian (approved):

src/world/area_obk/obk_01/BBD1A0.c
ethteck: I'm not sure what you mean. the question is why is temp-v0->unk_08 holding a pointer to itself

ethteck: the code is doing `x->y = x`

nanaian: Oh, I missed `x`/`temp_v0` was there twice. It's early in the day 😛 

I think this is probably because `unk_08` is a pointer that came from the outside environment (`evt_get_variable`), so something other than this function might need access to the allocated data. This theoretical other function won't be able to use `script->functionTempPtr[1]` after this API function is done, so instead it can dereference the evt variable

ethteck: I snee. maybe, but as far as I can tell, this is only used in this one function and barely 2 others

nanaian: Maybe this function is called multiple times from the script, ie the "other function" is just another call to this one

PR #890, opened 2022-12-16
96%

Unnunu:


BowserSlug:
🚀🚀🚀🚀🚀🚀🚀🚀🚀 This PR matches 9 functions (+0.12%) on `us`.

ethteck (approved):

include/macros.h
```diff
@@ -26,6 +26,13 @@
 #define NOP_FIX __asm__(".set nogpopt");
 #define NOP_UNFIX __asm__(".set gpopt");
 
+#define SWAP16(in, out) \
```
nanaian, line 29: Unsure what I think of this given we only use it once in this PR but perhaps I'll change my mind if it sees use elsewhere

nanaian (approved):

include/macros.h
HailSanta: I think it should go in the file its associated with, that's how its arranged in naudio.

ethteck: Sounds grood

PR #910, opened 2023-01-12
model functions

Unnunu:


BowserSlug:
🚀🚀🚀🚀 This PR matches 4 functions (+0.28%) on `us`.

include/model.h
```diff
@@ -142,6 +142,14 @@ typedef enum ShapeTypes {
     SHAPE_TYPE_SPECIAL_GROUP = 10,
 } ShapeTypes;
 
+typedef enum ExtraTileTypes {
+    EXTRA_TILE_0 = 0,
```
HailSanta, line 146: These are known and should be
EXTRA_TILE_NONE = 0,
EXTRA_TILE_MIPMAPS = 1,
EXTRA_TILE_AUX_SAME = 2, // has aux tile with same settings as main
EXTRA_TILE_AUX_INDEPENDENT = 3, // has aux tile with different settings

src/a5dd0_len_114e0.c
```diff
@@ -2988,8 +2988,638 @@ void state_render_frontUI(void) {
...
+                    break;
+            }
+        }
+    }
+
+    if (textureHeader != NULL) {
+        switch (extraTileType) {
+            case EXTRA_TILE_3:
+            case EXTRA_TILE_4:
+                prop = get_model_property(modelNode, MODEL_PROP_KEY_SPECIAL);
```
HailSanta, line 3126: The two-s32 values for this property are:
00AAABBB 000CD0EF
A = auxOffsetT
B = auxOffsetS
C = auxShiftT
D = auxShiftS
E = ModelReplaceType (an enum)
F = defaultScrollUnit

src/a5dd0_len_114e0.c
```diff
@@ -2988,8 +2988,638 @@ void state_render_frontUI(void) {
...
+        if (renderMode != RENDER_MODE_ALPHATEST && renderMode != RENDER_MODE_ALPHATEST_ONESIDED) {
+            *(*gfxPos) = D_8014B0B8[v1][combineSubType];
+        } else {
+            *(*gfxPos) = D_8014B400[v1][combineSubType];
+        }
+        (*gfxPos)++;
+    }
+
+    switch (s3) {
+        case 1:
```
HailSanta, line 3171: This is some 'texturing mode' enum, with at least these values:
1 = NO_TEXTURE
2 = NORMAL
3 = FOG
6 = FOG_MULTIPLY_COLOR

PR #911, opened 2023-01-13
misc battle cleanup III

HailSanta:


BowserSlug:
✅ This PR fixes 1 warnings!

src/16F740.c
```diff
@@ -3067,14 +3090,15 @@ void btl_state_update_end_player_turn(void) {
                 }
 
                 if (battleStatus->hustleTurns != 0 && (gBattleStatus.flags1 & BS_FLAGS1_HUSTLED)) {
-                    gBattleStatus.flags2 &= ~BS_FLAGS2_2;
 
+                    gBattleStatus.flags2 &= ~BS_FLAGS2_2;
                     btl_set_state(BATTLE_STATE_PREPARE_MENU);
+                    return;
```
ethteck, line 3096: I'm not sure I'm a fan of these extra returns. At the very least, I prefer putting the return at the end of a block where all paths will lead to a return, instead of putting it in every sub-block

PR #915, opened 2023-01-18
func_802401A0_DDC9D0 & save/load tattle flags

ethteck:
and some wip work


BowserSlug:
🚀 This PR matches 1 function (+0.04%) on `us`.

- src/190B20.c:437: warning: implicit declaration of function `create_target_list'

BowserSlug:
🚀 This PR matches 1 function (+0.04%) on `us`.

src/181810.c
```diff
@@ -442,9 +524,75 @@ s32 is_actortype_hpbar_visible(s32 actorType) {
...
+        if (list == NULL) {
+            break;
+        }
+
+        for (j = 0; list[j] != 0xFF; j++) {
+            if (actorType == list[j]) {
+                for (k = 0; list[k] != 0xFF; k++) {
+                    actorType = list[k];
+
+                    gb = get_global_byte((actorType / 8) + 0x16D);
```
nanaian, line 582: Can you add a note that GameByte(0x16D) is GB_Tattles_00? Alternatively, some macro for `(actorType / 8) + 0x16D` might be good given its done a few times

src/181810.c
```diff
@@ -13,21 +13,101 @@ BSS ActorPart* gSpeakingActorPart;
...
+    ACTOR_TYPE_DUPLIGHOST,
+    ACTOR_TYPE_GHOST_GOOMBARIO,
+    ACTOR_TYPE_GHOST_KOOPER,
+    ACTOR_TYPE_GHOST_BOMBETTE,
+    ACTOR_TYPE_GHOST_PARAKARRY,
+    ACTOR_TYPE_GHOST_BOW,
+    ACTOR_TYPE_GHOST_WATT,
+    ACTOR_TYPE_GHOST_SUSHIE,
+    ACTOR_TYPE_GHOST_LAKILESTER,
+    -1
```
nanaian, line 26: Given this is a u8 list it'd be better to use 0xFF or perhaps define an ACTOR_TYPE_LIST_END and use it in the below funcs too

src/world/area_omo/omo_11/DDC830.c
```diff
@@ -51,5 +51,130 @@ ApiStatus func_80240000_DDC830(Evt* script, s32 isInitialCall) {
...
-INCLUDE_ASM(s32, "world/area_omo/omo_11/DDC830", func_802401A0_DDC9D0);
+extern s32 D_80241760_DDDF90[6]; // TODO remove hard-coded size once data is migrated
+extern s32 D_80241778_DDDFA8[6]; // TODO remove hard-coded size once data is migrated
+
+NOP_FIX
+ApiStatus func_802401A0_DDC9D0(Evt* script, s32 isInitialCall) {
+    PlayerStatus* playerStatus = &gPlayerStatus;
+    Matrix4f sp20, sp60, spA0, spE0;
+    Npc* partner = get_npc_unsafe(NPC_PARTNER);
+    Model* model1 = get_model_from_list_index(get_model_list_index_from_tree_index(0x4A));
```
nanaian, line 68: 0x4A is MODEL_fl from mapfs/omo_11_shape.h

src/world/area_omo/omo_11/DDC830.c
```diff
@@ -51,5 +51,130 @@ ApiStatus func_80240000_DDC830(Evt* script, s32 isInitialCall) {
...
+            }
+        }
+    }
+
+    it = script->functionTempPtr[0];
+    for (i = 0; i < 6; i++, it++) {
+        loopModel = get_model_from_list_index(get_model_list_index_from_tree_index(D_80241760_DDDF90[i]));
+
+        for (j = 0; j < 4; j++) {
+            for (k = 0; k < 4; k++) {
```
nanaian, line 119: Worth ARRAY_COUNTing these to be the transformMatrix lengths probably

nanaian (changes requested):

src/181810.c
HailSanta: I disagree on this. Many different list types in the game use -1 to indicate "end of list", regardless of storage size. I don't think its a good idea to proliferate END_OF_X for each type. The only one I've been using is PTR_LIST_END for (void*)-1. Unfortunately, defining a generic LIST_END as -1 is not possible, as 0 is sometimes used for end of list as well.

src/world/area_omo/omo_11/DDC830.c
HailSanta: I think affine transform matrices can always be assumed to be 4x4; this can go either way imo.

src/181810.c
ethteck: yeah I forgot to change the checks to use -1 in the function. I prefer -1 to 0xFF because 0xFF feels like just a large number, where -1 to me indicates a non-normal value. but I'll change everything back to 0xFF for now

src/world/area_omo/omo_11/DDC830.c
ethteck: yeah idk, I don't think array_count is warranted here. if anyone is changing the dimensions of Matrix4f, good luck to them

src/181810.c
nanaian: I agree with using -1 but I'd assume that using that in the function might produce different asm? If not then yeah lets use -1 everywhere :)

include/common_structs.h
```diff
@@ -933,7 +933,7 @@ typedef struct BattleStatus {
     /* 0x194 */ u8 statusChance;
     /* 0x195 */ s8 statusDuration;
     /* 0x196 */ char unk_196;
-    /* 0x197 */ u8 targetHomeIndex; /* some sort of home idnex used for target list construction */
+    /* 0x197 */ s8 targetHomeIndex; /* some sort of home idnex used for target list construction */
```
pixel-stuck, line 936: idnex

src/181810.c
ethteck: it dont work :despair:

PR #921, opened 2023-01-24
area_omo

HailSanta:


include/common_structs.h
```diff
@@ -2478,14 +2478,14 @@ typedef struct PartnerDMAData {
...
-    /* 0x14 */ s32 treeIndex;
-} UnkOmo; // size = 0x18
+typedef struct RockingHorse {
+    /* 0x00 */ f32 posX;
+    /* 0x04 */ f32 posZ;
+    /* 0x08 */ f32 lastRockAngle;
+    /* 0x0C */ f32 rockPhaseAngularVel;
+    /* 0x10 */ f32 rockPhase;
+    /* 0x14 */ s32 modelID;
+} RockingHorse; // size = 0x18
```
nanaian, line 2488: Would it be better to move this to `omo.h`? Or even the specific `omo_xx.h` for the map it appears in? Unless of course this appears in engine code too, I don't know

HailSanta: Sounds fine. Ideally, it would go in a RockingHorse.inc.c along with the function. However, the devs have two flavors of the function among its three occurrences in the world with the only difference being the use of sin_deg vs sin_rad.

PR #925, opened 2023-01-28
Shiftability / symbol fixes

ethteck:


src/world/area_hos/hos_05/A2AAC0.c
```diff
@@ -1062,6 +1062,7 @@ void func_80243ED0_A2E110(s32 arg0, s32 idx) {
 
 #ifdef SHIFT
 #define TAPE_OFFSET title_tape_ROM_START - title_bg_1_ROM_START
+#define TAPE_OFFSET 0x2A440 // TODO remove this line once the map is split
```
nanaian, line 1065: why does the map need to be split for the title asset offsets to be valid

nanaian (approved):

src/world/area_hos/hos_05/A2AAC0.c
ethteck: because the code generated when you subtract two symbols is different, and it changes the length of a function, which makes other functions/symbols' offsets change

PR #927, opened 2023-01-31
Shiftability / modern gcc work

ethteck:


tools/build/configure.py
```diff
@@ -384,6 +393,10 @@ def build(object_paths: Union[Path, List[Path]], src_paths: List[Path], task: st
                 if entry.src_paths[0].suffixes[-1] == ".cpp":
                     task = "cxx"
 
+                top_lev_name = seg.get_most_parent().name
+                # if "evt" in top_lev_name:
+                #     task = "cc_modern"
```
nanaian, line 398: We should make this configurable instead of commented out like this. Perhaps a `--modern`?
Also perhaps add a custom splat entry property for it instead of configuring based on the name of the splat seg

ethteck: this is just me experimenting for now. it's not ready to make into a proper flag, since shifting itself isn't even confirmed to work. just got sick of maintaining a stash and thought it'd be fine to leave here commented-out

nanaian: ah fair, I saw your discord message but wasn't sure if you had cleaned it up. fine to leave as-is then ig

PR #928, opened 2023-01-31
area hos

HailSanta:


src/world/area_hos/hos_00/hos_00_3_unused.c
```diff
@@ -0,0 +1,12 @@
+#include "hos_00.h"
+
+EvtScript N(EVS_TexPan_ToggleUnused) = {
+    EVT_LABEL(0)
+		EVT_CALL(SetTexPanOffset, TEX_PANNER_D, TEX_PANNER_MAIN, 0, 0)
```
ethteck, line 5: looks like your editor switched to tabs here?

src/world/common/enemy/ai/PatrolNoAttackAI.inc.c
```diff
@@ -1,3 +1,6 @@
+#ifndef _PATROL_AI_NO_ATTACK_
```
ethteck, line 1: Is there a specific reason for adding an include guard to an inc.c file?

ethteck (commented):
:confetti_ball:

src/world/common/enemy/ai/PatrolNoAttackAI.inc.c
HailSanta: Several of them need these. Multiple enemy AIs include this, and when those enemies are included in the same maps together, only the first included PatrolNoAttackAI/etc appears on the final ROM. Based on comparisons between different maps, we can work out which parts of the larger includes are common and where guards are needed with fairly high confidence that the original sources were organized along similar lines.

ethteck: I shee. okay sounds good

PR #934, opened 2023-02-05
last audio function and more documentation

Unnunu:


BowserSlug:
🚀 This PR matches 1 function (+0.09%) on `us`.

src/audio/public.h
```diff
@@ -11,22 +11,22 @@ void snd_start_sound_with_shift(s32 soundID, u8 volume, u8 pan, s16 pitchShift);
...
-AuResult snd_ambient_80055590(s32 arg0, s32 arg1);
-AuResult snd_ambient_800555E4(s32 arg0);
-AuResult snd_ambient_80055618(s32 arg0, s32 arg1);
-AuResult au_ambience_set_volume(s32 arg0, s32 arg1, s32 arg2);
-AuResult au_ambience_disable(s32 arg0);
-AuResult au_ambience_enable(s32 arg0);
-void snd_ambient_80055760(s32 arg0);
-AuResult snd_ambient_800557CC(s32 arg0);
-AuResult snd_ambient_80055848(s32 arg0);
+AuResult au_ambient_load_sound(s32 arg0);
```
HailSanta, line 16: As a general note, I don't think these should be au_*. The intention of the original naming scheme was for au_ to be reserved for low-level audio code which should not be called from elsewhere in the engine (maps, menus, etc). These snd_amb and snd_song functions are part of the public interface for the audio system, and they usually delegate the work to an appropriate au_ function with error checking. sfx_* and bgm_* are similar and perhaps we should spend some time to hash out a consistent convention. Regardless, I think its important to maintain the distinction between different api "layers" in the audio subsystem.

ethteck (approved):
You're a wizard for this one.. what a crazy amount of RE / documentation on audio stuff. After you fix the merge conflicts & address Clover's comment, we can merge it

ethteck:
also I think since clover renamed the "advanced" door stuff, you may wanna update your PR to reflect that

PR #939, opened 2023-02-12
Build main segment in JP

marijnvdwerf:
All relevant changes are in the second commit.

`pm_msg` reading crashes, so I included that area as binary. In order to get `message_ids.h` imports resolved, I added a CPPFLAG: `-I ver/$version/include`. Message IDs in the Japanse version are different from the US version.


BowserSlug:
🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀 This PR matches 1269 functions (+95.81%) on `jp`.

src/5B320.c
```diff
@@ -3,6 +3,175 @@
 #include "item_entity.h"
 #include "world/world.h"
 
+#if !defined(VERSION_JP)
```
nanaian, line 6: Better do to `ifdef VERSION_US` I think?

src/5B320.c
```diff
@@ -3,6 +3,175 @@
...
+#define MSG_PartnerDesc_Goombaria_Battle 0
+#define MSG_PartnerDesc_Goombaria_World 0
+#define MSG_PartnerDesc_Goompa_Battle 0
+#define MSG_PartnerDesc_Goompa_World 0
+#define MSG_PartnerDesc_Twink_Battle 0
+#define MSG_PartnerDesc_Twink_World 0
+#define MSG_ShortDesc_32 0
+#endif
+
+#if !defined(VERSION_JP)
```
nanaian, line 147: Why not combine these blocks?

src/audio/2e230_len_2190.c
```diff
@@ -11,8 +11,12 @@ extern f32 AlTuneScaling[];
 #ifdef SHIFT
 #define SBN_ROM_OFFSET SBN_ROM_START
 #else
+#if defined(VERSION_JP)
```
nanaian, line 14: Prefer `#elif`

ver/us/splat.yaml
```diff
@@ -48,20 +48,21 @@ segments:
     - [0x1370, c, main_pre]
     - [0x1370, c, main]
     - [0x1B40, c, main_loop]
+    - [0x25A0, c, main_loop_2]
```
nanaian, line 51: We should find a better name for this. Perhaps something `background_gfx`y?

nanaian (changes requested):
Awesome work!!!

Throughout could we use `ifdef X` over `if defined(X)`?

src/5B320.c
marijnvdwerf: This is about compatibility defines; JP has strings for unused items. US (or CN) doesn't have strings for these. So the defines wouldn't get generated.

src/5B320.c
marijnvdwerf: I pulled them apart because they have different goals.

marijnvdwerf:
> `ifdef X` over `if defined(X)`

I figured `#if defined()` would make more sense, as there could also be cases of `#if !defined()`. Didn't know `#ifndef` existed though.

ver/jp/splat.yaml
```diff
@@ -34,199 +36,3197 @@ segments:
...
+    - [0x4A120, c, os/strcpy, gcc_272 -O1]
+    - [0x4A150, c, os/strcmp, gcc_272 -O1]
+    - [0x4A190, c, os/memset, gcc_272 -O1]
+    - [0x4A1D0, hasm, os/decode_yay0]
+    - [0x4A280, c, os/osFlash]
+    - [0x4AC70, pm_effect_loads, effect_loads]
+    - [0x4DF10, c, battle/battle]
+    - [0x4E580, bin, n_aspMain_text]
+
+    - [0x4F1E0, .data, main_pre] # TODO: find data offset
```
ethteck, line 290: you can just make these `auto` instead imo. but if you do wanna find the offset, go for it

ethteck (commented):
will do a thorough review within the next 12ish hours hopefully

include/common_structs.h
```diff
@@ -2202,7 +2202,11 @@ typedef struct {
     /* 0x00000 */ LookAt lookAt;
     /* 0x00020 */ Hilite hilite;
     /* 0x00030 */ Mtx camPerspMatrix[8]; // could only be length 4, unsure
+#ifdef VERSION_JP
+    /* 0x00230 */ Gfx mainGfx[0x2000];
```
ethteck, line 2206: I think having a define for this size, before the struct def, would be nicer

src/load_obfuscation_shims.c
```diff
@@ -6,6 +6,13 @@
 extern s8 obfuscated_obfuscation_shims_ROM_START[];
 extern s8 obfuscated_obfuscation_shims_VRAM[];
 
+#ifdef VERSION_JP
+extern Addr battle_heap_create_obfuscated_ROM_START;
```
ethteck, line 10: these should be generated and placed in ld_addrs if the segment is named properly in the yaml. I think that'd be better than explicitly putting them here

src/state_file_select.c
```diff
@@ -6,6 +6,18 @@
 #include "sprite.h"
 #include "model.h"
 
+#ifdef VERSION_JP
+extern Addr filemenu_ROM_START;
```
ethteck, line 10: same here as above

src/state_logos.c
```diff
@@ -100,13 +109,25 @@ void state_init_logos(void) {
...
     if (gGameStatusPtr->bSkipIntro) {
         if (intro_logos_fade_out(10)) {
             set_curtain_scale(1.0f);
             set_curtain_fade(0.0f);
             set_game_mode(GAME_MODE_TITLE_SCREEN);
         }
     } else {
+#ifdef VERSION_JP
+        // TODO: fix fake match
+        if (((u8) gGameStatusPtr->introState - 1 < 0xaU) && (a0 & 0xB000)) {
```
ethteck, line 125: this can be accomplished with an or `(introstate == x || introstate == x + 1 ...)`

ethteck: 0xB000 should be (0x8000 | 0x2000 | 0x1000) and the values for this are in enums.h (BUTTON_x)

src/state_pause.c
```diff
@@ -6,6 +6,18 @@
 #include "sprite.h"
 #include "model.h"
 
+#ifdef VERSION_JP
+extern Addr pause_ROM_START;
```
ethteck, line 10: these should be removable too

ver/jp/asm/os/bzero.s
```diff
@@ -0,0 +1,59 @@
+.include "macro.inc"
```
ethteck, line 1: are any of these handwritten asm files (not in nonmatchings) the same as us? I think it might be worth having a version agnostic "asm" dir for situations like this and referring to it from both yamls

ethteck (changes requested):
Great work! I'm really happy we have someone else looking at other versions, and this is hugely helpful in mapping things out.

marijnvdwerf:
Should be mergable/re-reviewable.

I wasn't able to address all the requested changes, but did talk about it with Ethan. Short writeup on that:
- Handwritten ASM files are most likely the same. I didn't have to make any changes in libultra stuff (aside from moving the vimode data, which originally would have been in separate files). Making those shared is probably going to require some changes in `./configure`, so I think that's better left as a TODO. It's a bit of a nuisance to diff because of the comments, labels and different splat versions. Compiling them and comparing them using `objdump` probably would make most sense.
- I added the `extern Addr <name>_ROM_START;` entries because those sections aren't in the linker script yet. I only did the main segment, and the rest is all done using `[0xoffset, bin]` entries. Which wouldn't work for the pause and filemenu sections, since they need the `_TEXT_END` and `_BSS_END` to be valid as well. I don't expect these sections to be _that_ horrible to match, but it's more effort than I'm willing to put in this first PR.

PR #942, opened 2023-02-14
Modern gcc + shiftability work

ethteck:


src/a5dd0_len_114e0.c
```diff
@@ -1883,10 +1887,10 @@ void clear_entity_data(s32 arg0) {
     }
 
     if (!gGameStatusPtr->isBattle) {
-        gEntityHeapBottom = BATTLE_ENTITY_HEAP_BOTTOM;
-        gEntityHeapBase = BATTLE_ENTITY_HEAP_BASE;
+        gEntityHeapBottom = WORLD_ENTITY_HEAP_BOTTOM;
+        gEntityHeapBase = WORLD_ENTITY_HEAP_BASE;
     } else {
-        gEntityHeapBottom = (s32)&D_801A7000;
+        gEntityHeapBottom = (s32)D_801A7000;
```
HailSanta, line 1893: Define D_801A7000 = BATTLE_ENTITY_HEAP_BOTTOM?

src/evt/evt.c
```diff
@@ -1341,15 +1354,15 @@ ApiStatus func_802C73B0(Evt* script) {
     return ApiStatus_DONE2;
 }
 
-s32 func_802C73B8(Evt* script) {
+ApiStatus func_802C73B8(Evt* script) {
```
HailSanta, line 1357: isInitiaCall arg

src/415D90.c
```diff
@@ -410,7 +410,7 @@ BSS s32 D_802AD66C;
 BSS s32 D_802AD670;
 
 //TODO remove it in func_802A45D8 and use D_802AD670 instead
```
pixel-stuck, line 412: seems like we can remove these comments now

src/a5dd0_len_114e0.c
```diff
@@ -3938,7 +3948,7 @@ void _load_model_textures(ModelNode* model, s32 romOffset, s32 size) {
 }
 
 void load_model_textures(ModelNode* model, s32 romOffset, s32 size) {
-    s32 battleOffset = ((gGameStatusPtr->isBattle != 0) << 17);
+    s32 battleOffset = ((gGameStatusPtr->isBattle != 0) << 17); // TODO FIX
```
pixel-stuck, line 3951: perhaps out of scope for this PR, but I imagine this will still match as:
```C
s32 battleOffset = (gGameStatusPtr->isBattle) ? 0x20000 : 0;
```

src/audio/2e230_len_2190.c
```diff
@@ -21,7 +21,9 @@ void func_80052E30(u8 index) {
     voice->priority = AU_PRIORITY_FREE;
 }
 
-#ifndef SHIFT
+#ifdef SHIFT
```
pixel-stuck, line 24: I think we can probably get rid of these shifting asm. It was good for proving the ROM segments were shiftable, but now that we've finished that, its probably fine to get rid of these sometime soon and just use the ifdef on the SBN offset.

pixel-stuck (commented):
Great work :)

src/evt/evt.c
ethteck: these (this and the two above) don't have those args

PR #957, opened 2023-02-26
two geometric functions

Unnunu:


BowserSlug:
🚀🚀 This PR matches 2 functions (+0.04%) on `us`.
🚀🚀 This PR matches 2 functions (+0.04%) on `jp`.
🚀🚀 This PR matches 2 functions (+0.04%) on `cn`.

ethteck (approved):

src/camera.c
```diff
@@ -243,30 +437,30 @@ void func_80032C64(Camera* camera) {
                 deltaPosX = settings2->posB.x - settings2->posA.x;
                 deltaPosZ = settings2->posB.z - settings2->posA.z;
 
-                if (func_800325E4(settings2->posA.x, settings2->posA.z, settings2->posA.x - deltaPosZ, settings2->posA.z + deltaPosX,
+                if (calculate_line_segment_intersection(settings2->posA.x, settings2->posA.z, settings2->posA.x - deltaPosZ, settings2->posA.z + deltaPosX,
                                   camera->targetPos.x, camera->targetPos.z, newPosX, newPosZ, &sp44, &sp48, &sp4C) && sp4C < 1000000.0f) {
```
HailSanta, line 441: express the constant as SQ(1000.0f)

src/camera.c
```diff
@@ -19,11 +19,205 @@ CameraControlSettings* test_ray_zone(f32 posX, f32 posY, f32 posZ, Collider** zo
...
+    if (fabsf(dx12) > fabsf(dx34)) {
+        /*
+        We represent intersection point P as P = P3 + alpha * (P4 - P3)
+        and solve the equation (P - P1) x (P2 - P1) = 0, where 'x' is cross product, P1 and P2 are points on the line
+        and P3 and P4 are the ends of the segment.
+        So, (P3 - P1 + alpha * (P4 - P3)) x (P2 - P1) = 0,
+        alpha = [-(P3 - P1) x (P2 - P1)] / [(P4 - P3) x (P2 - P1)]
+        */
+        // same as -(P3 - P1) x (P2 - P1)
+        alpha = x1 * dz12 - z1 * dx12 + dx12 * z3 - dz12 * x3;
```
HailSanta, line 95: I think it would be better to use the same notation for the variables as you do for the comments.
i.e., P1x, P1z instead of x1, z1.
Further, consider calling the points A1, A2, B1, B2, instead of P1/2/3/4 to indicate which line segment each belongs to.

PR #958, opened 2023-02-26
camera control points union

HailSanta:


include/common_structs.h
```diff
@@ -741,8 +741,24 @@ typedef struct CameraControlSettings {
     /* 0x00 */ s32 type;
     /* 0x04 */ f32 boomLength;
     /* 0x08 */ f32 boomPitch;
-    /* 0x0C */ Vec3f posA;
-    /* 0x18 */ Vec3f posB;
+    union {
+        struct {
+            f32 Ax;
```
ethteck, line 746: Can this not be 2 vec3fs still? I think I'd prefer that

HailSanta: Depending on the camera controller type, this set of six floats can be used in two different ways. It would be nice if we could union 2x Vec3f and 3x Vec2f, but in the 3x Vec2f configuration they store the components for C in the now-unused y-components for A and B. I feel this way of writing it allows the camera code to be clearer, rather than just having to know that in some cases posA.y means Cx and posB.y means Cz.

BowserSlug:
🚀 This PR matches 1 function (+0.11%) on `us`.

PR #959, opened 2023-03-01
update_camera_mode_unused

ethteck:
find_similar_areas saved the day here - it found that basically the bulk of the function (where all the issues remained) was identical to the bulk of update_camera_mode_6. I was able to copy/paste the code and change a couple minor things and got a match.

I also fixed an issue where the chinese version wasn't checked in coverage.py, and I updated the hard-coded size in progress.py since it seems to have changed after matching this function.


BowserSlug:
🚀 This PR matches 1 function (+0.02%) on `us`.
🚀 This PR matches 1 function (+0.02%) on `jp`.
🚀 This PR matches 1 function (+0.02%) on `cn`.

src/B0E0.c
```diff
@@ -40,54 +35,51 @@ void update_camera_mode_unused(Camera* camera) {
         camera->lookAt_obj.z = camera->lookAt_obj_target.z;
     }
 
-    if (!(playerStatus->flags & (PS_FLAG_FALLING | PS_FLAG_JUMPING))) {
+    if (!(playerStatus->flags & (4 | 2))) {
```
HailSanta, line 38: restore flags

HailSanta (approved):

PR #961, opened 2023-03-02
lightning_bolt_appendGfx

Unnunu:


BowserSlug:

include/effects.h
```diff
@@ -1775,11 +1775,11 @@ typedef struct LightningBoltFXData {
     /* 0x02C */ s32 lifetime;
     /* 0x030 */ Color4i outerColor;
     /* 0x040 */ Color3i innerColor;
-    /* 0x04C */ f32 unk_4C;
-    /* 0x050 */ f32 unk_50[12];
-    /* 0x080 */ f32 unk_80[12];
-    /* 0x0B0 */ f32 unk_B0[12];
-    /* 0x0E0 */ f32 unk_E0[12];
+    /* 0x04C */ f32 widthScale;
+    /* 0x050 */ f32 sparkleX[12];
```
HailSanta, line 1779: I don't think "sparkle" makes sense here, as these are the points used to build the triangle strips for the bolt "beam", interpolated points between start and end with some "random" offset from the preset. Perhaps boltVertexPos? Or something less clunky if you have a better idea.

HailSanta (approved):

PR #965, opened 2023-03-06
Misc Labeling

Macrepeh:
- Partially labeled Magikoopa miniboss file
- Labeled most of Buzzar's actions
- Labeled UnkBattleFunc2 as EnemyDrainGFX as both functions are for the effects that spawn when enemies drain HP from Mario (fuzzies etc)


src/battle/area/iwa/actor/buzzar.c
```diff
@@ -8,19 +8,19 @@
...
-extern EvtScript N(nextTurn_8021B918);
-extern EvtScript N(8021BEBC);
-extern EvtScript N(8021CCE8);
-extern EvtScript N(8021DF2C);
-extern EvtScript N(8021F030);
-extern EvtScript N(8021F7EC);
-extern EvtScript N(8021F808);
-extern EvtScript N(8021F824);
-extern EvtScript N(8021FA2C);
+extern EvtScript N(init);
```
HailSanta, line 11: All EvtScript should be EVS_PascalCaseName

src/battle/area/iwa/actor/buzzar.c
```diff
@@ -41,12 +41,12 @@ s32 N(idleAnimations_8021B644)[] = {
     STATUS_END,
 };
 
-s32 N(defenseTable_8021B650)[] = {
+s32 N(defenseTable)[] = {
```
HailSanta, line 44: PascalCase for data names

Macrepeh: I can't find a single enemy that has it as "DefenseTable"

HailSanta: It's okay, they haven't been cleaned up yet. There wasn't really a way to know.

src/battle/area/iwa/actor/buzzar.c
```diff
@@ -8,19 +8,19 @@
...
-extern EvtScript N(8021F7EC);
-extern EvtScript N(8021F808);
-extern EvtScript N(8021F824);
-extern EvtScript N(8021FA2C);
+extern EvtScript N(init);
+extern EvtScript N(taketurn);
+extern EvtScript N(idle);
+extern EvtScript N(handleEvent);
+extern EvtScript N(nextTurn);
+extern EvtScript N(attackWindBlast);
```
HailSanta, line 16: I think we're going to use EVS_Attack_AttackName for these, so you're on the right track here.

src/battle/area/iwa/actor/buzzar.c
```diff
@@ -8,19 +8,19 @@
...
+extern EvtScript N(attackFeatherFling);
+extern EvtScript N(attackGrappleDrop);
+extern EvtScript N(attackClawSwipe);
+extern EvtScript N(beginFlyToHome);
+extern EvtScript N(beginFlyToHomeGrappleFail);
```
HailSanta, line 21: This can be EVS_BeginFlyToHome_GrappleFail

src/battle/area/iwa/actor/buzzar.c
```diff
@@ -1126,19 +1126,19 @@ EvtScript N(8021F030) = {
     EVT_END
 };
 
-EvtScript N(8021F7EC) = {
-    EVT_EXEC_WAIT(N(8021F824))
+EvtScript N(beginFlyToHome) = {
```
HailSanta, line 1129: Generally when we have these wrapper scripts that just execute another script, the outer one is names EVS_Name and the inner one is EVS_Name_Impl. So this would be EVS_FlyToHome and the script its executing would be EVS_FlyToHome_Impl.

src/battle/area/mim/actor/forest_fuzzy.c
```diff
@@ -426,11 +426,11 @@ EvtScript N(80219054) = {
         EVT_CALL(PlaySoundAtActor, ACTOR_SELF, SOUND_214)
         EVT_CALL(GetActorPos, ACTOR_SELF, LVar0, LVar1, LVar2)
         EVT_ADD(LVar1, 10)
-        EVT_CALL(N(UnkBattleFunc2), LVar0, LVar1, LVar2, LVar3)
+        EVT_CALL(N(StartEnemyDrainGFX), LVar0, LVar1, LVar2, LVar3)
```
HailSanta, line 429: GFX should be avoided when not referring to Gfx (display list) data. Elsewhere in the codebase, we would name these functions SpawnXxxEffect or SpawnXxxFX. So something like SpawnDrainHealthStartFX and SpawnDrainHealthContinueFX would be consistent, albeit unwieldy.

HailSanta (approved):

PR #967, opened 2023-03-07
Build world/area_mac/machi for PAL

marijnvdwerf:
This is mostly meant as a place for discussion; how would we go about integrating the different event timings for the PAL release. Alternatively: is it worth bothering matching the PAL release?

(I have a manhandled US splatfile that generates a similar-ish rom to the PAL version, pushed to <https://github.com/marijnvdwerf/papermario/tree/pal>. The ROM addresses should match up fairly well, VRAM is likely a bit off)


HailSanta:
I think the best way to support this is to add a delta-time scaling factor to each of these constants.
```
#ifdef VERSION_PAL
#define FPS_FOR_VERSION 50
#else
#define FPS_FOR_VERSION 60
#endif

#define DT (FPS_FOR_VERSION/60)
```
Then we can use the US/JP times as a baseline. That is, EVT_WAIT(10) becomes EVT_WAIT(10 * DT), which resolves to EVT_WAIT(8) for VERSION_PAL. Likewise for speeds, EVT_FLOAT(2.1) becomes EVT_FLOAT(2.1 / DT), which should resolve to EVT_FLOAT(2.52). Of course, theres tens of thousands of places where this would need to be applied... and the EVT_FLOAT argument can be tricky, so I hope this works in all cases.

BowserSlug:
🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀 This PR matches 97 functions (+0.45%) on `pal`.

ethteck (approved):

src/world/area_mac/mac_01/npc/flower_gate.inc.c
```diff
@@ -331,10 +339,14 @@ s32 N(ItemList_MagicalSeeds)[] = {
 EvtScript N(EVS_MinhT_PlantSeed) = {
     EVT_CALL(SetNpcFlagBits, NPC_SELF, NPC_FLAG_IGNORE_PLAYER_COLLISION, TRUE)
     EVT_CALL(SpeakToPlayer, NPC_MinhT, ANIM_MinhT_Talk, ANIM_MinhT_Idle, 0, LVar4)
-    EVT_WAIT(10)
+    EVT_WAIT(10 * DT)
     EVT_CALL(UseSettingsFrom, CAM_DEFAULT, 266, 27, 373)
     EVT_CALL(SetPanTarget, CAM_DEFAULT, 266, 27, 373)
+#if VERSION_PAL
```
marijnvdwerf, line 345: This pattern is appearing quite a lot, even for this single area. Would be nicer if it could be a single macro `FOO(1, / DT)` (though starting with an operator looks a bit weird)

nanaian: maybe EVT_TIMESCALED(1) that is an int on ntsc but a float divided by DT on pal

unsure about the name

marijnvdwerf: There's also cases where it's multiplied by DT instead of divided (https://github.com/pmret/papermario/pull/967/files#diff-6fe3ae3603c232cbade9a79373aad40a5a940e12d115bcc449e5c932e6b5ed11R316).

src/world/common/entity/SuperBlock.inc.c
```diff
@@ -305,7 +305,7 @@ API_CALLABLE(N(SuperBlock_AnimateEnergyOrbs)) {
             }
 
             userData->scatterStateTime++;
-            if (userData->scatterStateTime >= 21) {
+            if (userData->scatterStateTime >= (int)(21 * DT)) {
```
HailSanta, line 308: s32

marijnvdwerf:
mac, tik, kgr and kmr are built for the PAL release now. I think this is a good spot for a merge, with the PR still being a controllable length. Most changes in events involved adding ` * DT`, so that's not very exciting. There's a couple of functions impacted, which are more interesting:
- [func_802405F0_8AC580](https://github.com/pmret/papermario/pull/967/files#diff-7f2b316071331f0272c7af1db4395c8299a58ed77a8a77f34de1ba1dd50b55c3R106): Variable storage changed here, achieved by using `#define`s.
- [ManageEffects](https://github.com/pmret/papermario/pull/967/files#diff-64245365cc75b555eacd91dae1c385ab0f45fd57688d9a702c223683eceb260dR28): The intervals changed from 10/20/30/40 to 8/16/24/32
- [appendGfx_records_impl](https://github.com/pmret/papermario/pull/967/files#diff-95b49304517f9a3c72a1572f3d8818bda2e59fd8b3cebee810c0e21848a7d453R60): The charset passed changed from 0 to 1.

ethteck (approved):

src/world/area_kmr/kmr_00/kmr_00_3_revival.c
```diff
@@ -103,16 +103,30 @@ API_CALLABLE(N(func_802405F0_8AC580)) {
         FUNC_STATE_3        = 3
     };
 
+#if VERSION_PAL
+#define tempVar script->varTable[11]
```
HailSanta, line 107: Going forward, please ALL_CAPS for these to distinguish from local variables.

marijnvdwerf: I went looking through the codebase for precedent, and I was wondering if there's a preference for including the `script->` part or not. In `world/partner/goombario.c`, there's `#define USE_STATE functionTemp[0]`, used as `script->USE_STATE++;`

HailSanta: We've had discussions about it, but not widely implemented them. Leaving off the script part is preferred.

nanaian:
Will take a look at this later today

nanaian (approved):

src/world/area_kmr/kmr_00/kmr_00_3_revival.c
```diff
@@ -104,11 +104,11 @@ API_CALLABLE(N(func_802405F0_8AC580)) {
     };
 
 #if VERSION_PAL
-#define tempVar script->varTable[11]
-#define tempVar2 script->varTable[12]
+#define VAR_1 varTable[11]
```
HailSanta, line 107: Interesting that PAL exchanges a function temp for a script var here.

PR #990, opened 2023-03-22
matched func_802428C8_A2CB08

Unnunu:


BowserSlug:
🚀 This PR matches 1 function (+0.05%) on `us`.


🚀 This PR matches 1 function (+0.05%) on `pal`.

ethteck:
like the other recent match, I can't see anything to really take issue here with other than possibly the names of things, which we can always revise later. so in the interest of nyoom-culture, gonna merge this.

PR #994, opened 2023-03-23
disaptch_damage_event_actor

ethteck:
Solved by copying the beginning of dispatch_damange_event_partner, running the permuter, and then removing `partner->currentHP = 127;` since it's not in this version.


BowserSlug:

PR #998, opened 2023-03-25
Update Jenkinsfile to include parallel steps 

Ampthere:
The purpose of this PR is to add the parallel directive to some of the steps in the build process. This could potentially add some speedup (not sure how much). But thought it might be of benefit.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,10 +6,13 @@ pipeline {
     stages {
         stage('Setup') {
             steps {
-                sh 'cp /usr/local/etc/roms/papermario.us.z64 ver/us/baserom.z64'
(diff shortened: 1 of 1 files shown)


ethteck:
Thanks for your interest and work on improving our build @Ampthere. However, I think to properly make things more performant, we should come up with a plan in chat first. The key slowdown involves building the code and assets, which this doesn't parallelize. 

My advice is that we 
1. parallelize the build for each version
2. migrate progress reporting to the new progress API to simplify the stash/unstash stuff
3. abstract common assets so they're only extracted once for all the versions

Idea 3 here renders 1 less efficient, so we can probably switch back to a single build with all versions once the asset extraction/building is completely de-duplicated.

I'd be happy to chat about this in discord sometime to implement some of these things. As for this PR, I think let's close it and start anew once we have a plan.

PR #1004, opened 2023-03-26
Introduce clang-format

nanaian:


ethteck:
@nanaian have you given this more thought? I still think the benefits of having auto-formatting don't outweigh the cons atm, without lots of engineering work to make it fit nicely with EVTs. And considering I think where we last ended that convo was that we would put EVTs in separate files with their own DSL, that seems like it blocks this anyway

nanaian:
As @marijnvdwerf said perhaps it would be beneficial to have maintainers look at the diff clang-format spits out as a suggestion but not fail the PR for it. That is, we'd use it to find and suggest fixes to nits. Does that make sense?

ethteck:
Kinda, but I find the output kinda blah, like "look at these lines"

nanaian:
Mm, that's true. Perhaps we close this PR for now but keep the branch and revisit once we have a better solution for evts

marijnvdwerf:
> Kinda, but I find the output kinda blah, like "look at these lines"

That's why I added the `clang-format` command to the action. It's output does show the actual diff:
image

ethteck:
i sheep

ethteck:
I don't think we want this whole PR - just the clang format file by itself

nanaian:
> I don't think we want this whole PR - just the clang format file by itself

I've moved the job to the `lint` workflow which can continue to be disabled until we solve the EVT issue

.clang-format
```diff
@@ -0,0 +1,61 @@
...
+    Priority:        10
+  # everything else
+  - Regex:           '.*'
+    Priority:        1
+    SortPriority:    0
+
+AllowShortFunctionsOnASingleLine: None
+---
+
+...
```
marijnvdwerf, line 61: What are these dots?

ethteck: image

PR #1006, opened 2023-03-27
add git hook to make sure assets/$version has been gitignored

nanaian:
This will hopefully dissuade any modders from trying to commit vanilla assets to their repos

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,7 +26,6 @@ build.ninja
 *.z64
 *.Yay0
 *.msg.h
-*.bin
 *.i
 *.o
 *.zip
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -26,7 +26,6 @@ build.ninja
 *.z64
 *.Yay0
 *.msg.h
-*.bin
```
nanaian, line 29: This is to enable people to override `.bin` files with an asset stack. We already ignore the dump asset directories

.gitignore
```diff
@@ -41,7 +40,12 @@ build/
 /tools/permuter_settings.toml
 
 # Assets
-/assets/*
```
nanaian, line 44: This is to make it so people hopefully won't have to go modifying gitignore when adding new asset directories

nanaian:
Overall it would be better if we checked to see if the commit _contains_ any assets, since force-adding is still possible, but this is probably good enough & isn't an expensive check to make on commit

Mastergatto:
Hi, may I suggest to also add *.n64 and *.v64 to .gitignore file? Just to be 100% sure, although this may be a thing that is unlikely to occur, if it ever happens.

PR #1007, opened 2023-03-28
Add run script to launch emulator

nanaian:
Introduces `./run`, which runs ninja for the current version and tries to run it in an emulator, trying a bunch of common emulator executable paths.

This comes out of @Wrymouth asking for an easy way to compile & run the ROM, as vscode-star-rod lets him do.

I'm not sure whether we want this in pmret or just in papermario-dx. What do we think?

### DIFF
--- a/run
+++ b/run
@@ -1,0 +1,27 @@
+#!/usr/bin/env python3
+
+import os
+
+ROM = "ver/current/build/papermario.z64"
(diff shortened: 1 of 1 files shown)


nanaian:
Closing per @ethteck on Discord:

> I think this is better for dx. The decomp repo doesn't need "play the game" features imo

PR #1009, opened 2023-04-07
fold doc 1

HailSanta:


include/enums.h
```diff
@@ -5983,50 +5984,73 @@ enum FoldStateFlags {
...
     FOLD_TYPE_F                       = 0xF,
-    FOLD_TYPE_10                      = 0x10,
-    FOLD_TYPE_11                      = 0x11,
+    FOLD_TYPE_10                      = 0x10,   // unused?
+    FOLD_UPD_ALLOC_COLOR_BUF          = 0x11,   // args: count
 } FoldType;
 
+typedef enum FoldAnim {
+    FOLD_ANIM_SHOCK                   = 0x00, // used for Goombaria and Goompapa when Kammy drops the hammer block
+    FOLD_ANIM_SHIVER                  = 0x01, // used when Goombaria gives Mario a kiss
```
nanaian, line 6009: I would think this is also used when hitting a hammer block that is stronger than the current hammer level

include/enums.h
```diff
@@ -5983,50 +5984,73 @@ enum FoldStateFlags {
...
-    FOLD_TYPE_1                       = 0x1,
-    FOLD_TYPE_2                       = 0x2,
-    FOLD_TYPE_3                       = 0x3,
-    FOLD_TYPE_4                       = 0x4,
-    FOLD_TYPE_5                       = 0x5,
-    FOLD_TYPE_6                       = 0x6,
-    FOLD_TYPE_7                       = 0x7,
-    FOLD_TYPE_8                       = 0x8,    // color overlay?
+    FOLD_UPD_CLEAR                    = 0x0,    // or FOLD_UPD_INIT?
+    FOLD_TYPE_1                       = 0x1,    // unused?
```
nanaian, line 5988: Why UPD (update?) over TYPE now?

nanaian (approved):

ethteck (approved):

include/enums.h
HailSanta: Yes, you can see this on line 1554 of 77480.c:
`func_802DDFF8(playerStatus->anim, FOLD_UPD_SET_ANIM, FOLD_ANIM_SHIVER, 1, 1, 0, 0);`

include/enums.h
HailSanta: These are commands consumed by `fold_update`. Originally, I believed each FOLD_TYPE was associated with a different 'paper' animation, but now its clear that only FOLD_UPD_SET_ANIM (and to some extent FOLD_UPD_WAVY) does this. Instead, most of the "FOLD_UPD" commands are used to adjust color and alpha blending parameters for the objects they're applied to, with fairly subtle distinctions between them.

PR #1010, opened 2023-04-08
appendGfx_shading_palette

ethteck:
And some odds and ends:
* Updated diff.py, fixed diff_settings.py so our expected dir setting actually works
* Fixed offset in `HeartBlockContentData` struct (thanks IceBound777 !)
* Fixed a texture that was split in half due to splat limitations
* Implemented proper use of `"gcc/string.h"` header to avoid warnings in modern GCC
* Reduced other warnings for modern GCC
* Improved update_symbol_addrs.py a bit
* WIP on nonmatchings
* Various other code cleanup


BowserSlug:

🚀 This PR matches 1 function (+0.06%) on `ique`.

src/325AD0.c
```diff
@@ -55,10 +55,18 @@ s32 func_E0200044(s32 max, s32 idx) {
 #ifdef NON_EQUIVALENT
 void func_E02000AC(s32 arg0, s32 arg1) {
     s32 i, j;
+    s32 newvar;
+    s32 width = 80;
+    s32 tmp = 0;
+    s32 x, y;
+    int new_var;
+    int new_var2;
```
nanaian, line 63: IK this is nonequivalent but probably best to be using `s32` over `int` anyway

nanaian (approved):
j'approuve, other than typing nit that doesn't really matter

PR #1015, opened 2023-04-11
image fx

HailSanta:


nanaian (commented):
Regarding capitalisation, you've gone for "Img**fx**" as if its one word but we have the following precedent

- AuFX
- AuFxBus

...looks like we need to pick something consistent; I'd prefer capital F and lowercase x personally (ImgFx) but up for discussion

ethteck:
it depends on the context I think, unfortunately.

I think I prefer imgFX / ImgFX to imgFx / ImgFx

BowserSlug:

🚀 This PR matches 1 function (+0.06%) on `ique`.

nanaian:
@ethteck fine with me. I think im just against 'imgfx' as one word as it is in this PR

PR #1020, opened 2023-04-20
Add shell.nix

NotNite:
Congrats on 100%!

Adds a `shell.nix` for easy `nix-shell` usage. Effectively the same steps as Docker - skips the install process but requires the user to run `./configure` and `ninja`.

I'm not super happy with the quality of this - having tar extracts in a shell hook is terrible. I was originally not gonna PR this, but I was told it's welcomed, so I'm doing it anyways.

Because of the "interesting" quality of this, I'll leave it as a draft for now in case any actually intelligent Nix users are around this block. But hey, it's the first time I've PR'd Nix support to something, so that's cool!

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -6,6 +6,7 @@ This repository supports:
 - [macOS](#unix)
 - [Windows Subsystem for Linux 2](#wsl-2)
 - [Docker](#docker) (any host OS)
+- [Nix](#nix)
 
 If you encounter any issues setting up the repo, please feel free to [reach out to us on Discord](https://discord.gg/urUm3VG).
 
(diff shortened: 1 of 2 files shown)


ethteck:
I don't see this being *harmful*, so I'm willing to accept it, personally. but I also am not that interested in testing it and I'm not sure who would want to...any takers?

I guess the only thing about it that's slightly non-ideal is that I wouldn't be able to maintain it if we added dependencies to the project, like we may do with Rust/Java as we add some more tooling for asset support.

nanaian:
I like this, but neither @ethteck nor I would be able to maintain it. @NotNite is there a way we can contact you for assistance when we are adding new dependencies to the project?

NotNite:
Yeah - I'm in the Discord server and I can be reached over email.

nanaian:
I approve merging this then without one of us testing @ethteck 😛

ethteck:
@NotNite did you wanna tidy up anything? Or shall we just take it out of draft and wahoo

NotNite:
You can wahoo away!

nanaian (approved):

Issue #392, opened 2021-08-25
Merge version-specific asset directories


nanaian:
Related to #383.

It's been suggested that we merge assets/us and assets/jp into an assets/core directory that, similar to src, only uses different asset files for each version where necessary. Doing this will probably require a change to the custom `mapfs` and `npc_sprites` segments such that we can define both a name and a folder to use (e.g. `toad` loading the `toad_jp` folder on jp).

PR #1021, opened 2023-04-21
[WIP] ui_images

marijnvdwerf:
This needs some thinking about file structuring (German/Spanish uses a folder, French uses a locale suffix). Personally I'm partial to `{{ lang }}/ui/pause/label_stats.png`. But we need to think about how that would work with different regional releases. There is #392, but that seems to be mostly about version-based asset grouping.


nanaian:
I agree on the file structure. It mimics the way both the asset stack and [LayeredFS](https://yuzu-emu.org/help/feature/game-modding/) work.

nanaian:
#392 is for versions, and is primarily about making it so we have a canonical base version that the other assets build off of. This might be an abstract "core" directory that is version-indepdendent, and then the version-specific assets live in override directories for each. (So JP's asset stack would be `jp, core`)
I don't think there's any need to develop a similar system here since the languages are mutually exclusive, i.e., they are switched between rather than override eachother.

nanaian (approved):
Why's this marked WIP? LGTM!

marijnvdwerf:
Because French and German still use a different naming approach (for the sake of discussion), so that needs to be made the same.

HailSanta:
I prefer the way French does it. The PAL assets will already be separated from the core assets on a version-specific portion of asset stack. Keeping language variations of textures together and suffixed in the same directory makes all the version-specific variations clearly visible and comparable while browsing. I think separate asset directories should only exist per version (JP, PAL, etc).

nanaian:
I wouldn't be strongly against the suffix method if thats preferred by the majority

marijnvdwerf:
Hard disagree on suffixes making items more visible/comparable. The `de` folder clearly shows what files are "overridden" for a single language, and for the French assets you need to scour the entire `assets` folder.

image

image

ethteck:
yeah I think I'm in the "asset dir first" camp. I get clover's point, but I prefer to just have all the language-specific stuff in one top-level differing dir

nanaian (approved):

ethteck (approved):
Approving, but we should wait to merge because @HailSanta doesn't agree with this approach and we haven't really hashed it out yet

HailSanta (approved):

PR #1025, opened 2023-04-24
--debug fixes

nanaian:
Allows `./configure --debug us` to work.
`create_elf_section_list_auto` was either removed or renamed - worth looking into this @ethteck @LightningTH?

### DIFF
--- a/include/effects_internal.h
+++ b/include/effects_internal.h
@@ -3,6 +3,12 @@
 
 #include "effects.h"
 
+#ifdef PRINT_ON_PANIC
+// TODO: provide shim for func_80025F44 instead
+#define ASSERT(condition) if (!(condition)) { while (TRUE) {} }
+#define PANIC() while (TRUE) {}
+#endif
+
 s32 func_E0200000(s32);
 s32 func_E0200044(s32, s32);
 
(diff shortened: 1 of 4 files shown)


include/effects_internal.h
```diff
@@ -3,6 +3,12 @@
 
 #include "effects.h"
 
+#ifdef PRINT_ON_PANIC
+// TODO: provide shim for func_80025F44 instead
```
ethteck, line 7: what's that function do? what does this change have to do with fixing --debug

nanaian: See macros.h

ethteck:
I think we should hold off then until we know that "fixing" it is going to produce something usable. The debug build was basically lightning's experiment that never ended up being 'usable' for its intended purpose, so if anything I think we should remove the --debug flag until we have something that we know works

LightningTH:
I'm fine with the flag being removed. I had debug fully building and
running but ran into issues with gdb and my goal of being compatible with
normal gdb. I didn't want to maintain a special gdb branch but that went
into deep territory and is outside this bug.

LightningTH:
Poke me in discord and we can talk more about it. I wrote a doc on my
findings and situation.

ethteck:
Given what we talked about in discord, can we close this until we have something more concrete to work with?

nanaian:
Yep

PR #1027, opened 2023-04-24
Replace python-githooks with manual command

### LINKED ISSUE #1026: install.sh git hook feature complains when used as submodule
image

https://github.com/nanaian/merlon/blob/310679be162f0f053e9cce9e0e0bea721c327804/src/new.rs#L91

Gorialis:
This removes `python-githooks` and replaces it with a manual addition of the hook, fixing two known issues:
- Previously, the hook would not install properly in a submodule and now it does (fixes #1026)
- `python-githooks` would wrap the hook in a Python script that would sometimes fail to run if the system didn't have a `python` -> `python3` symlink. This now no longer has a wrapper script, so that problem doesn't occur.

### DIFF
--- a/.githooks.ini
+++ b/.githooks.ini
@@ -1,7 +1,0 @@
-# To refresh this file:
-#   python3 -m python_githooks
-
-# Make sure dumped assets have not been unignored
-# https://github.com/pmret/papermario/wiki/Asset-overrides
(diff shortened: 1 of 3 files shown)


nanaian (approved):
LGTM, thanks for the swift fix!

ethteck:
Thanksh!

PR #1044, opened 2023-05-01
title_screen_draw_copyright (iQue)

ethteck:
Merging these was pretty yucky. We might want to just split this into 4-5 different versions instead at some point. Not sure


src/state_title_screen.c
```diff
@@ -73,7 +73,7 @@ typedef struct TitleDataStruct {
 extern s16 D_800A0970;
 extern TitleDataStruct* TitleScreen_ImgList;
 extern s32* TitleScreen_ImgList_Logo;
-extern s32* TitleScreen_ImgList_Copyright;
+extern u8 (*TitleScreen_ImgList_Copyright)[144]; // 144 being the width of the image
```
marijnvdwerf, line 76: 144 should be a constant. The JP version uses a width of 128
https://github.com/pmret/papermario/blob/main/tools/splat_ext/pm_map_data.py#LL141C1-L147C62

nanaian (approved):
I don't think the merged function reads awfully, I like what's been done with the macros 👍

src/state_title_screen.c
```diff
@@ -450,15 +450,29 @@ void title_screen_draw_press_start(void) {
...
     gDPPipeSync(gMainGfxPos++);
 }
 
 #if VERSION_IQUE
-INCLUDE_ASM(void, "state_title_screen", title_screen_draw_copyright);
+#define RECT_SIZE 0x28
+#define YL_BASE 724
+#define YH_BASE 764
+#define COPYRIGHT_TEX_CHUNKS 4
+#define COPYRIGHT_IMG &TitleScreen_ImgList_Copyright[k]
```
marijnvdwerf, line 465: Maybe at least make these a function-like macro, so it's more obvious it depends on `k` /`i`?
```suggestion
#define COPYRIGHT_IMG(k, i) &TitleScreen_ImgList_Copyright[k]
```

marijnvdwerf (changes requested):
Some numbers should be changed to use `x << 2` format, so all the coordinates are in the screen size range. I don't think the combined function reads awfully, but the macros could be a bit clearer in what other state they depend on.

PR #1047, opened 2023-05-03
Clean executable files

marijnvdwerf:
- Remove the executable bit from any files without a shebang
- Introduce some missing shebangs (make_expected.sh)
- Make all shebangs use env, so venv python is properly used

To talk about:
- I removed leading spaces from all shebangs to make them uniform. No personal preference, just wanted consistency
- Do all the files that currently have shebangs need to have them/be executable? Asking because earlier I found one in `diff_settings.py`, which doesn't need to be executable by itself.


nanaian (approved):
> I removed leading spaces from all shebangs to make them uniform. No personal preference, just wanted consistency

No preference from me either, this looks good :+1:

> Do all the files that currently have shebangs need to have them/be executable? Asking because earlier I found one in diff_settings.py, which doesn't need to be executable by itself.

Not necessary; sounds like `diff_settings.py` doesn't need it. Feel free to remove it if you want, but it doesn't really matter.

PR #1048, opened 2023-05-04
fix configure.py to find correct split.py

Rainchus:


src/190B20.c
```diff
@@ -2342,7 +2342,7 @@ void func_802666E4(Actor* actor, f32 x, f32 y, f32 z, s32 damage) {
 }
 
 // grossness
-void func_802667F0(s32 arg0, Actor* actor, f32 x, f32 y, f32 z) {
+void DisplayActionRating(s32 arg0, Actor* actor, f32 x, f32 y, f32 z) {
```
HailSanta, line 2345: show_action_rating (convention for non-callable function names, also we use show_ a lot more than display_)

HailSanta (changes requested):

PR #1051, opened 2023-05-07
Support Manjaro

### LINKED ISSUE #1050: install.sh doesn't understand what distro Manjaro arm64 is
Hey I just wanted to report that the install script doesn't seem to know what Manjaro arm64 is which is based on Arch.

```
The following distros (and their derivatives) are supported by install.sh:
- Debian/Ubuntu (apt)
- Arch Linux (pacman)
- openSUSE (zypper)
- Alpine Linux (apk)
Please consider contributing and adding an installation script for your distro.
```

nanaian:
Should fix #1050.

@coreybruce can you test?

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -60,7 +60,7 @@ if cat /etc/os-release | grep -E 'ID=debian|ID_LIKE=(.*)debian' &> /dev/null; th
 fi
 
 # Arch Linux and derivatives (pacman)
-if cat /etc/os-release | grep -E 'ID=arch|ID_LIKE=arch' &> /dev/null; then
+if cat /etc/os-release | grep -E 'ID=arch|ID_LIKE=(arch|manjaro-arch)' &> /dev/null; then
     supported=true
 
     echo "Installing packages for Arch Linux or derivative (pacman)"
--- a/install.sh
+++ b/install.sh
@@ -60,7 +60,7 @@ if cat /etc/os-release | grep -E 'ID=debian|ID_LIKE=(.*)debian' &> /dev/null; th
 fi
 
 # Arch Linux and derivatives (pacman)
-if cat /etc/os-release | grep -E 'ID=arch|ID_LIKE=(arch|manjaro-arch)' &> /dev/null; then
+if cat /etc/os-release | grep -E 'ID=arch|ID_LIKE=(arch|manjaro arch)' &> /dev/null; then
     supported=true
 
     echo "Installing packages for Arch Linux or derivative (pacman)"


coreybruce:
> Should fix #1050.
> 
> @coreybruce can you test?



> Should fix #1050.
> 
> @coreybruce can you test?

Sure, do I just need to clone his git repo and try it out?

ethteck:
You clone this repo and check out the install-sh-manjaro branch

alternatively, you can run this gh command after cloning the repo if you have the github cli tools set up: `gh pr checkout 1051`

ethteck:
I think we should just merge this

nanaian:
> I think we should just merge this

I'd rather wait for confirmation that it works. There's no need to rush to merge it and @Gorialis mentioned something in Discord about arch and manajaro packages not necessarily being 1:1; we don't consider that at all here so it might be a problem

ethteck:
okay, I'll tell the part of me that likes seeing a light PR list to chill out for a while 😛

nanaian:
@coreybruce Awesome! If you run into any issues, feel free to join us in #decomp-chat in https://discord.gg/paper-mario-modding

ethteck:
sounds like you don't have the github command line tools installed. no problem - yeah, do that

ethteck:
yep!

ethteck:
Yeah I personally am not so keen on us maintaining an install.sh for a ton of different arches and would prefer to just have a document as well. it'd allow us to explain the caveats of different platforms and yada yada

coreybruce:
Yeah I mean the script shouldn't to specify on architectures and be one universal script for Linux. So where do we go next with this?

nanaian:
"be one universal script for Linux" - this would work if the dependencies were under flatpak or something, but all the distros have different opinions about package management.

nanaian:
> Would make it easier to maintain this repo and be less work if you gave the commands to build instead of using a custom script you will have to keep updating for every distro

Moving the commands from a script to docs would do nothing more than make it slightly less convenient for users. The commands differ per distro - for example, Arch and friends use `pacman` and the package names across Arch or Manjaro for example will probably be different.

We could just add a clause to the docs that says "hey if install.sh doesnt work then we recommend you find out how to install these things: x y z h" / fall back in the script

ethteck:
I'll try to explain my reasoning for why a human-readable doc is better than a bash script in this case. 

For starters, we end up installing things without really informing the user, running `sudo`. I think this is pretty bad practice, and I know it's been raised before as something to consider. I also think it's just generally better to have users install their own dependencies instead of running a bunch of commands for them. We of course can provide said commands, but running the commands automatically in their native environment to install system packages has always felt a bit intrusive to me.

Also, as we see in cases like these, if the feedback we expect doesn't exactly match what the bash script checks, the script just totally gives up or fails. Whereas a human-readable doc might lead someone to the correct solution.

In both cases, insufficiencies in the script/doc would lead people to ask for help and then we'd fix said script or doc. What's nice about the doc, though, is that it allows us to easier add notes and caveats that someone might care about instead of littering the shell script with logic and/or print-outs that will be drowned out by the other output from commands.

I guess you could ask "why not both?" and to that I'd say that maintaining two things is even worse than one. So I'd really prefer not to do that, but I acknowledge it'd be the best of both worlds...just at the expense of 2x the maintenance work.

nanaian:
That sounds good. Perhaps we could have configure.py print out a nice error it if notices that you are missing dependencies, and link to this doc?

ethteck:
Shall we close this given #1059 ?

coreybruce:
Sure, if I run into any problems building with build commands on Arm than I'll make a new issue post.

Issue #949, opened 2023-02-20
Python scripts aren't formatted using Black


marijnvdwerf:
When making a change in configure.py, I wanted to make sure that I properly formatted the code. I assumed `black` would do the trick, but it introduced a lot of changes. It would be nice if automatic formatting would be supported.

nanaian:
- black all the Python files except splat
- Run black in CI
- Enable black as vscode default formatter

Issue #1053, opened 2023-05-10
configure: sprite building should be able to merge sprite dirs


nanaian:
Currently, the following directory structure

```
assets/us/sprite/npc/
    SpriteSheet.xml
    Raster0.png
assets/mymod/sprite/npc/
    SpriteSheet.xml
```

will be interpreted by configure/sprite.py as

```
assets/*/sprite/npc/
    SpriteSheet.xml          <---- assets/mymod/sprite/npc/SpriteSheet.xml
    [no Raster0.png]
```

which will then error because SpriteSheet.xml will refer to Raster0.png which it thinks doesn't exist anymore. The directories should be seen as merged, which is how all the other assets work, i.e.

```
assets/*/sprite/npc/
    SpriteSheet.xml          <---- assets/mymod/sprite/npc/SpriteSheet.xml
    Raster0.png              <---- assets/us/sprite/npc/Raster0.png
```

PR #1055, opened 2023-05-14
Player sprites & more

ethteck:
Adds player sprite building, fixes various things, separates the installation of compilers (in-repo tools) into a separate script

Todo:
- [x] Add player sprite support to all versions (except JP, since it's not identical like the other ones)
- [x] Generate player sprite header
- [x] Add max thing clover wants to sprite header
- [x] Incorporate player sprite header into code wherever possible
- [x] Migrate NPC sprite animations to the new human-readable format
- [x] Combine player and npc sprites into one segment
- [x] Un-hardcode npc sprite filesize offset from main sprite header
- [x] Make build dependencies 
- [x] Add build date to main output xml
- [x] Implement building with an asset stack
- [x] ~~Figure out differences with JP sprite bundle~~ (there were no differences)
- [x] Implement marking the proper objects as ninja dependencies in the asset stack


tools/build/configure.py
```diff
@@ -23,11 +23,21 @@
...
+
+def write_ninja_rules(
+    ninja: ninja_syntax.Writer,
+    cpp: str,
+    extra_cppflags: str,
+    extra_cflags: str,
+    use_ccache: bool,
+    shift: bool,
+    debug: bool,
+):
```
nanaian, line 40: Looks like you ran `black` - can we close #949 perhaps?

tools/build/sprite/player_sprites.py
```diff
@@ -0,0 +1,466 @@
...
+        if (
+            os.path.getsize(player_sprite_dir / "rasters" / f"{raster_name}.png")
+            == 0x10
+        ):
+            special_raster_bytes = (
+                b"\x80\x30\x02\x10\x00\x00\x02\x00\x00\x00\x00\x01\x00\x10\x00\x00"
+            )
+            RASTER_CACHE[raster_name] = PNGInfo(0, 0, cur_offset, special_raster_bytes)
+            cur_offset += 0x10
+            continue
```
nanaian, line 297: Have we tried looking at the loading code to figure out what this data is supposed to be? `80300210` looks like it might be an address...

tools/build/sprite/player_sprites.py
```diff
@@ -0,0 +1,466 @@
...
+        f.write(final_data)
+
+
+if __name__ == "__main__":
+    if len(sys.argv) != 3:
+        print("usage: player_sprites.py [OUTBIN] [DIR]")
+        exit(1)
+
+    _, outfile, sprite_dir = sys.argv
+    build(Path(sprite_dir), Path(outfile))
```
nanaian, line 466: Instead of passing a single directory, can we pass all source files or a set of directories that will be layered? See #1053

tools/build/configure.py
```diff
@@ -540,29 +707,49 @@ def build(object_paths: Union[Path, List[Path]], src_paths: List[Path], task: st
...
+                    "sprite_combine",
+                )
+                build(entry.object_path, [entry.object_path.with_suffix(".bin")], "bin")
+            elif seg.type == "pm_player_sprites":
+                build(
+                    entry.object_path.with_suffix(".bin"),
+                    entry.src_paths,
+                    "player_sprites",
+                    glob_deps=False,
+                )
```
nanaian, line 728: I think this needs implicit dependencies on all the files inside `entry.src_paths[0]` so that the player sprites are rebuilt when any of the source files change. Currently, I presume, this flow can happen:
1. Configure
2. Ninja
3. Modify `assets/*/sprite/player/SpriteSheet.xml`
4. "ninja: no work to do" 💀

Doing #1053 as discussed in my other comment should resolve this though, since we can make `entry.src_paths` be all the files instead of just the parent directory.

nanaian (changes requested):

tools/build/configure.py
ethteck: That issue is to have it run consistently, so my cleanup doesn't address that

tools/build/sprite/player_sprites.py
ethteck: 8030 doesn't seem like an address to me, but I could be wrong. I think only clover can answer that question and it may have to wait until the code is figured out further. Or he may, already know now. Lol @HailSanta do you?

tools/build/sprite/player_sprites.py
ethteck: How do we influence the order properly? I don't really understand how this would work.

tools/build/configure.py
ethteck: I haven't set up the dependencies for the build tasks yet

tools/build/sprite/player_sprites.py
nanaian: Specifically it would look a bit like
```
player_sprites.py [OUTBIN] assets/us/sprite/player/SpriteSheet.xml assets/us/sprite/player/MyRaster.png assets/mymod/sprite/player/NewRaster.png
```
where are file stems are unique but if a file exists higher up the asset stack we prefer it. there's a function in configure.py for this

tools/splat_ext/pm_player_sprites.py
```diff
@@ -0,0 +1,643 @@
...
+        )
+
+        #########
+        # Writing
+        #########
+
+        self.out_path().mkdir(parents=True, exist_ok=True)
+        # Write build info (date)
+        with open(self.out_path() / "build_info.txt", "w") as f:
+            f.write(build_date)
```
nanaian, line 602: Not sure what's better, but we hardcode this string depending on verson when for building mapfs, and in shiftable mode we use the current date

ethteck: Yeah I don't think it matters, but I like having it extracted

nanaian: Actually, not with the current namescheme. But #1055 makes me want to ask for a more generic name like `sprites.xml` given that its already in the `player/` directory

nanaian: basically what im thinking is like
assets/*/sprite/{npc,player}/sprites.xml
```
<xml>
   <BuildDate>10 AM dog time</BuildDate>
   <Names>
     ...
   </Names>
</xml>
```

ethteck: shounds good

tools/splat_ext/pm_player_sprites.py
```diff
@@ -0,0 +1,643 @@
...
+from n64img.image import CI4
+from segtypes.n64.segment import N64Segment
+from util import options
+from util.n64.Yay0decompress import Yay0Decompressor
+
+sys.path.insert(0, str(Path(__file__).parent))
+from sprite_common import AnimComponent, read_offset_list
+
+PAL_TO_RASTER: Dict[str, int] = {
+    "8bit": 0x57C90,
```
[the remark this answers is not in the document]
ethteck: We considered it, but this is really only a convenience feature for better visualizing palettes. That being said, I'm happy to add it to the yaml so it's easier to change if people want. I'll add a todo for that ✅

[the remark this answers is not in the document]
nanaian: Fine to leave it TBH until we start thinking about other versions

nanaian (approved):

PR #1057, opened 2023-05-17
various cleanup

HailSanta:
tutankoopa, battle state suspension, item entites, actor palette swapping, many unk fields


include/common_structs.h
```diff
@@ -1506,13 +1506,13 @@ typedef struct PushBlockGrid {
...
     /* 0x04 */ f32 gravity; /* 2 = normal, 1 = low gravity, higher values never 'settle' */
-    /* 0x08 */ f32 unk_08;
+    /* 0x08 */ f32 collisionRadius;
     /* 0x0C */ f32 constVelocity;
     /* 0x10 */ f32 velx;
     /* 0x14 */ f32 velz;
     /* 0x18 */ f32 moveAngle;
     /* 0x1C */ s32 timeLeft;
-    /* 0x20 */ s32 unk_20;
+    /* 0x20 */ s32 useSimplePhysics;
```
ethteck, line 1515: The way this is named, it seems like a bool

HailSanta: Good catch

src/181810.c
```diff
@@ -554,32 +562,32 @@ void save_tattle_flags(s32 actorType) {
...
-            if (actorType == list[j]) {
-                for (k = 0; list[k] != 0xFF; k++) {
-                    actorType = list[k];
+        for (j = 0; types[j] != 0xFF; j++) {
+            if (actorType == types[j]) {
+                for (k = 0; types[k] != 0xFF; k++) {
+                    actorType = types[k];
 
-                    gb = get_global_byte((actorType / 8) + 0x16D); // GameByte(0x16D) is GB_Tattles_00 (first tattle)
+                    gb = get_global_byte((actorType / 8) + EVT_INDEX_OF_GAME_BYTE(GB_Tattles_00));
```
marijnvdwerf, line 590: Maybe also swap these around, so it becomes `EVT_INDEX_OF_GAME_BYTE(GB_Tattles_00) + actorType / 8`? Or add a macro for getting this byte, considering how often this is used?

src/182B30.c
```diff
@@ -2,13 +2,96 @@
...
+};
+
+s16 FearPaletteAnimXOffsets[] = { -2, 2, 0, 0, -2, 2, 0, 0, 0, 0, -2, 2, 0, 0, 0, 0, 0, 0, 255 };
+
+s16 ParalyzePaletteAnimXOffsets[] = { -2, 2, 0, 0, -2, 2, 0, 0, 0, 0, -2, 2, 0, 0, 0, 0, 0, 0, 255 };
+
+// animated palettes for Watt
+// each pair gives { mode, duration }
+u8 WattPalettesAnim[] = {
+    1,  2,
```
marijnvdwerf, line 67: These numbers are right-aligned, while the other arrays in this file are left-aligned.

Also, not sure what `clang-format` is gonna do to this. Maybe something like this could work?
```
#define TUPLE(foo, bar) (foo), (bar)

u8 StaticPalettesAnim[] = {
    TUPLE(0, 32),
    TUPLE(1, 4),
    TUPLE(2, 2),
    255
};
```

src/182B30.c
```diff
@@ -2,13 +2,96 @@
...
+    0, 54,
+    1,  2,
+    0, 28,
+    1,  2,
+    0,  6,
+    1,  2,
+    0, 44,
+    1,  2,
+    0, 44,
+    255
```
marijnvdwerf, line 80: Is there no constant for end-of-list markers?

src/182B30.c
```diff
@@ -2150,106 +2236,110 @@ void func_80259AAC(s32 isNpcSprite, ActorPart* part, s32 yaw, Matrix4f mtx, s32
...
+        decorationTable->nextPalTime--;
     } else {
-        temp = D_80284080[decorationTable->unk_6C2];
+        paletteType = StaticPalettesAnim[decorationTable->palAnimState];
     }
-    switch (temp) {
-        case 0:
+
+    switch (paletteType) {
+        case 0: // no change
```
marijnvdwerf, line 2295: Should this not just be an enum?

src/17D6A0.c
```diff
@@ -2246,11 +2246,11 @@ s32 bMarioDefendAnims[] = {
     STATUS_FROZEN,          ANIM_Mario1_Frozen,
     STATUS_TRANSPARENT,     ANIM_Mario1_Idle,
     STATUS_TURN_DONE,       ANIM_Mario1_Still,
-    STATUS_13,              ANIM_MarioB2_BerserkStill,
+    STATUS_TURN_DONE_BERSERK,   ANIM_MarioB2_BerserkStill,
```
marijnvdwerf, line 2249: This breaks the table alignment :/

nanaian: Need more input, looks like both marjin and I would prefer that we stop doing alignment in these files, as it makes for bad diffs ([relevant Discord discussion](https://discord.com/channels/279322074412089344/768585278331289622/1108316885503721524)). Therefore, if @ethteck and @HailSanta agree, we can just leave this line as-is and remove the clang-format rule that enforces alignment (bearing in mind use of clang-format is optional for contributions right now)

src/181810.c
nanaian: mhm, like `EVT_TATTLE_GAME_BYTE(actorType)`

src/17D6A0.c
HailSanta: I think we should continue doing alignment in the base repo, but not enforce it for mods.

src/182B30.c
HailSanta: I think the formatter should be disregarded in these cases. There are many instances of tables like this throughout the code.

src/182B30.c
nanaian: I reckon we should make an effort define an `END_` value for all types of sentinel lists

src/181810.c
HailSanta: Ah, I missed this one. Good catch.
Thought about making a macro, but decided against it because this actorType -> flag calculation is done in a couple different ways in this file and it doesn't make much sense to macro-ify some of them but not others. 
Would be nice if we could match with something like
`s32 flagByte = actorType / 8;`
`s32 flagIdx = actorType % 8;`
but unfortunately, it does not match that way.

src/effects/damage_indicator.c
```diff
@@ -69,58 +69,62 @@ f32 D_E003CD1C[] = {
...
+    f32 angle;
     DamageIndicatorFXData* part;
-    s32 origNumParts = numParts;
+    s32 trueDamageAmt = damageAmt;
     s32 i;
 
-    if (numParts > 20) {
-        numParts = 20;
+    
+//    fx_damage_indicator(0, posX, posY, posZ, 10.0f, a, damageAmt, &gDamageCountEffects[i]);
```
ethteck, line 85: is this comment necessary? seems leftover

ethteck (approved):

ethteck:
@nanaian @marijnvdwerf do ye aprov

nanaian (approved):

PR #1058, opened 2023-05-18
install.sh gcc-12

hydrationMan:
brew install's gcc-13 without specification. This results in configure.py failing due to the hard coded specification of gcc-12

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -31,7 +31,7 @@ if [[ "$uname" == "Darwin" ]]; then
     fi
 
     # Install packages
-    brew install md5sha1sum ninja gcc nanaian/brew/mips-linux-gnu-binutils || exit 1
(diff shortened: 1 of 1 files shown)


ethteck:
Thanks for this PR! I think what we should do isn't lock a specific version of gcc in the install script (which will be going away soon anyway) but just have configure look for multiple gcc versions

hydrationMan:
That makes sense, I suppose I could get something running with glob

nanaian:
Can confirm I’ve been having this issue too. Solution should be to have configure.py try cpp-{14,13,12,11}

nanaian:
@hydrationMan are you interested in implementing the wildcard behaviour into `configure.py`? (If not, I'd be happy to do it!)

ethteck:
I've implemented this in #1055 since I'm making configure.py changes there anyway. I'll close this for now

PR #1060, opened 2023-05-20
Document PRNG Related Functions

Lincoln-LM:
This is a pretty simple PR that gives some functions and constants more fitting names, restructures the ``effect_rand_int`` (formerly ``func_E0200000``) and ``advance_rng`` functions to match their intended implementations, and adds a few comments explaining them.


include/effects_internal.h
```diff
@@ -3,7 +3,7 @@
 
 #include "effects.h"
 
-s32 func_E0200000(s32);
+s32 gfx_rand_int(s32);
```
pixel-stuck, line 6: Not quite sure I'd call this "gfx" as that's often what we use to refer to things that are directly interfacing with the n64 graphics hardware. Its a silly differentiation, but these graphical effects are just referred to as "fx" or "effect" I think "effect_rand_int" is probably a good name.

src/43F0.c
```diff
@@ -487,53 +487,62 @@ void dma_write_block(Addr dramAddr, u32 devAddr, s32 size) {
...
-    return temp_v0 / 32768.0;
+    return (advance_rng() & 0x7FFF) / 32768.0;
 }
 
-s32 func_80029994(s32 arg0) {
-    u32 div = -1;
-    s32 plusOne = arg0 + 1;
+// return a random integer [0,max]
+s32 rand_int_internal(u32 max) {
+    u32 partition_size = 0xFFFFFFFF;
```
HailSanta, line 506: partitionSize and maxPlusOne case for local cariables

HailSanta (changes requested):

ethteck (approved):
Thanks for the cool PR! :)

PR #1063, opened 2023-05-24
Document Probability Distribution of Calls to rand_int

Lincoln-LM:
This is another very simple documentation PR which prefaces the majority of calls to ``rand_int`` used in conditions with a fractional representation of the probability of that condition being true as well as an approximate representation of the probability as a percentage in the cases where it is constant. I believe this is helpful as it more directly points out when the off-by-one nature of rand_int likely unintentionally differs from what would otherwise be assumed (ex. a rand_int(100) < 50 would likely read as "50/100, 50% chance", when it actually represents "50/101, 49.5% chance"). It may still be a little unclear what the comments are exactly referring to, so if a different structure, placement, or system of labeling in general seems more fitting I'd be happy to correct it.


src/16F740.c
```diff
@@ -104,28 +115,39 @@ void btl_merlee_on_first_strike(void) {
...
                             playerData->merleeSpellType = MERLEE_SPELL_1;
+                        // 45/101 ≈ 44.6%
                         } else if (temp <= 90) {
                             playerData->merleeSpellType = MERLEE_SPELL_2;
+                        // 10/101 ≈ 9.9%
                         } else {
                             playerData->merleeSpellType = MERLEE_SPELL_EXP_BOOST;
                         }
+                    // 31/101 ≈ 30.7%
                     } else if (temp <= 30) {
```
nanaian, line 129: The formatting here is putting me off slightly since technically this in the `if` block above, so it should be indented, but then that wouldn't look like it applied to the `else if`. I think `clang-format` would complain too if we ran it.

Perhaps we could do something like this:

```
 } else if (temp <= 30) { // 31/101 ≈ 30.7%
```
or
```
 } else if (temp <= 30 /* 31/101 ≈ 30.7% */) {
```

Does anyone else have any thoughts? I like both ways I suggested.

nanaian: This need only apply to the `else if` cases in this function, nowhere else in the PR IMO.

nanaian (changes requested):
Awesome idea and work! I have a style comment on 16F740.c that needs some discussion, but other than that this is great :)

nanaian:
It would be good to add a comment above the `rand_int(int)` declaration, wherever it is in the header files, to describe this behaviour. You could put "See #1063 for more information" too if it seems appropriate.

src/16F740.c
ethteck: I much prefer the first way. I have a strong aversion to comments in between code

ethteck:
I agree with adding a comment to describe the off-by-one-ness with an example maybe. but I'd prefer we don't link to github stuff and would rather just have all relevant info in-repo

src/16F740.c
marijnvdwerf: What about just having it as the first line in the codeblock itself? so 
``` c
 } else if (temp <= 30) {
     // 31/101 ≈ 30.7%
     ...
 }
 ```

nanaian: I think I prefer the aside `} else if (temp <= 30) { // 31/101 ≈ 30.7%`, but failing that, your suggestion marijn

Lincoln-LM:
> It would be good to add a comment above the `rand_int(int)` declaration, wherever it is in the header files, to describe this behaviour. You could put "See #1063 for more information" too if it seems appropriate.

I assume by this something other than the already present:

https://github.com/Lincoln-LM/papermario/blob/ee3e93c84e817bdd50e5162f70e6a96ca4d3bfc0/src/43F0.c#L527-L528

is being suggested? Or is it just to move it to its declaration in functions.h?

nanaian:
Ah, looks fine then. I think we will move where doc comments live (e.g. to headers) at a later date; no need to worry about that for now!

HailSanta (changes requested):
I like the comments breaking down the percentages on the else/if ladders where it isn't clear how much probability each case spans. I think those would be best written as end of line comments:
``} else if (n < 100) { // 100/300 ~33%``
I like these as end of line comments because they're essentially supplementary footnotes. Comments on their own line are best used to explain the logic or motivation for a following block of code.

I think the other rand_int comments should be dropped, especially where they are only emphasizing the off-by-one bug. This is an interesting oversight to note, but it should be documented for the function rand_int instead of attending every instance of its use.

ethteck:
I agree - I'm a fan of the ones in the if/else chains, but I'd prefer not decorating every single call and instead have the behavior described in the rand func's documentation, itself.

ethteck:
@Lincoln-LM are you cool with making the above changes? If not, someone else can take over

Lincoln-LM:
Apologies for the delay, focus has been elsewhere. Should be fine now.

ethteck:
No worries! Thank you :)

PR #1069, opened 2023-06-18
Let Jenkins do parallel builds

marijnvdwerf:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,10 +6,6 @@ pipeline {
     stages {
         stage('Setup') {
             steps {
-                sh 'cp /usr/local/etc/roms/papermario.us.z64 ver/us/baserom.z64'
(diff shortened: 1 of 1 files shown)


marijnvdwerf:
This is turning out to be more complicated than I expected. @ethteck I suspect that the matrix workspaces aren't being cleaned. Checking out https://jenkins.deco.mp/job/papermario/view/change-requests/job/PR-1069/20/flowGraphTable/, ique and jp both say there's nothing to be done.

```
+ bash -o pipefail -c ninja 2>&1 | tee build_log_ique.txt
ninja: no work to do.
```

ethteck:
that's disappointing. I know in the pipeline stuff, there's a function to clean the workspace, which I thought we were using...but we might need to be more liberal with where we use it. or perhaps there's a different approach you're supposed to take with matrix builds

ethteck:
I agree. this was a great effort, but probably it's better to focus on improving the efficiency of other aspects of the build before we revisit parallelizing it

PR #1072, opened 2023-06-27
tex archives

HailSanta:


ethteck (approved):

tools/build/configure.py
```diff
@@ -887,7 +894,15 @@ def build(
...
                     elif name.endswith("_tex"):
                         compress = False
-                        bin_path = path
+                        tex_dir = path.parent / name
+                        build(
+                            bin_path,
+                            [tex_dir],
+                            "tex",
+                            variables={
+                                "tex_dir": str(tex_dir)
```
nanaian, line 903: Is `str()` required? Elsewhere it doesn't seem to be

nanaian (changes requested):
tex_archives.py is susceptible to the same problem as #1053, happy to merge it and fix it later though if we open an issue

tools/build/configure.py
ethteck: maybe you're thinking of something else? I see lots of calls to str() to convert Paths to strings here, since the variables dict is typed as str,str

PR #1076, opened 2023-07-01
whoooooweee that is some good ol' fashioned decomp (pal funcs)

ethteck:


BowserSlug:
🚀🚀🚀🚀🚀 This PR matches 5 functions (+0.05%) on `pal`.

src/pause/pause_partners.c
```diff
@@ -605,6 +611,14 @@ void pause_partners_init(MenuPanel* panel) {
     }
     setup_pause_menu_tab(gPausePartnersWindowBPs, ARRAY_COUNT(gPausePartnersWindowBPs));
 
+#if VERSION_PAL
+    gWindows[37].width = D_PAL_80271B50[gCurrentLanguage];
```
HailSanta, line 615: WINDOW_ID_etc

HailSanta (approved):

BowserSlug:
🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀 This PR matches 10 functions (+0.35%) on `pal`.

include/enums.h
```diff
@@ -6584,13 +6584,12 @@ enum WindowStyles {
     WINDOW_STYLE_MAX    = 22,
 };
 
-#if VERSION_PAL
+#define LANGUAGE_DEFAULT 0
```
nanaian, line 6587: Maybe LANGUAGE_EN is more appropriate
```suggestion
#define LANGUAGE_DEFAULT LANGUAGE_EN
```

include/variables.h
```diff
@@ -265,6 +265,8 @@ extern PlayerData gPlayerData;
 
 #if VERSION_PAL
 extern s32 gCurrentLanguage;
+#else
+#define gCurrentLanguage 0
```
nanaian, line 269: Same here
```suggestion
// Only English is supported
#define gCurrentLanguage LANGUAGE_EN
```

src/effects/chapter_change.c
```diff
@@ -295,6 +264,13 @@ void chapter_change_appendGfx(void* effect) {
         case 16:
         case 17:
         case 18:
+#if VERSION_PAL
+            if (gCurrentLanguage == LANGUAGE_DE) {
+                ptr0 = D_E010E838_b;
+                ptr1 = D_E010E838_c;
```
nanaian, line 270: Maybe we can rename D_E010E838_b and D_E010E838_c to D_E010E838_de D_E010E838_de? Or is it better to wait until we fully know what it is

src/filemenu/filemenu_common.c
```diff
@@ -1028,10 +1027,51 @@ void filemenu_init(s32 arg0) {
...
+
+        if (menu->page == 0) {
+            fio_has_valid_backup();
+            if (D_800D95E8.saveCount >= 4) {
+                D_800D95E8.saveCount = 0;
+            }
+            gGameStatusPtr->saveSlot = D_800D95E8.saveCount;
+        }
+
+
```
nanaian, line 1050:
```suggestion
```

tools/m2ctx.py
```diff
@@ -25,6 +25,7 @@
     "-D__asm__(test...)=",
     "-ffreestanding",
     "-DM2CTX",
+    "-DVERSION_PAL",
```
nanaian, line 28: Revert this probably

nanaian (changes requested):

include/enums.h
ethteck: The idea here is that the first index of the arrays this often is used to index is the default language for that version (Chinese for iQue, since it only has one language. Japanese for JP, for the same reason)

src/effects/chapter_change.c
ethteck: I'd prefer to wait

tools/m2ctx.py
ethteck: It's intentional. Without it, m2ctx is useless currently, since all of US is finished. Parameterizing it would be nice, but this is a quick fix in the meantime

include/variables.h
ethteck: this is also 0 rather than english for the reason above

PR #1080, opened 2023-07-08
rename D_800A0948 to gLoadedFromFileSelect

Rainchus:


src/111f0_len_860.c
```diff
@@ -37,7 +37,7 @@ void state_step_enter_world(void) {
                 gGameStatusPtr->isBattle = FALSE;
                 gGameStatusPtr->disableScripts = FALSE;
 
-                if (D_800A0948 == 0) {
+                if (gMapLoadFrom == LOAD_FROM_MAP) {
```
HailSanta, line 40: I suspect D_800A0948 is actually a boolean "gLoadingDemoMap" (instead of sharing the values from this enum), considering the usage here and where its set.

ethteck (changes requested):
Jenkins failed with something on the JP build. Make sure you're renaming across all versions

src/111f0_len_860.c
```diff
@@ -37,7 +37,7 @@ void state_step_enter_world(void) {
                 gGameStatusPtr->isBattle = FALSE;
                 gGameStatusPtr->disableScripts = FALSE;
 
-                if (D_800A0948 == 0) {
+                if (gLoadedFromFileSelect == FALSE) {
```
ethteck, line 40:
```suggestion
                if (!gLoadedFromFileSelect) {
```

PR #1082, opened 2023-07-11
Snug bugs unrelated to and never been inside of a rug

ethteck:


tools/splat_ext/pm_map_data.py
```diff
@@ -1,9 +1,8 @@
+from math import ceil
 import os, sys
 from pathlib import Path
-from typing import List
 from segtypes.n64.segment import N64Segment
 from util.n64.Yay0decompress import Yay0Decompressor
-from util.color import unpack_color
```
nanaian, line 6: Why this change?

nanaian (changes requested):
Question about a change related to `unpack_color` because now we are duplicating code

tools/splat_ext/pm_map_data.py
ethteck: it's not in splat anymore. Eventually pigment/n64img will export these functions for others to use. 

For now I'd rather just copy the code than do more and more path hacking, which has already gotten really ugly. It makes it easy to identify all the clones, anyway

nanaian: Fair enough, once pigment gets it we should import it from there yes

nanaian (approved):

PR #1085, opened 2023-07-15
SBN encoding

Hurricane996:


ethteck:
image
from Jenkins - I assume maybe a rename didn't go through?

tools/splat_ext/pm_sbn.py
```diff
@@ -0,0 +1,884 @@
...
+    # the "size" field of the SBN File entry usually, but not always, matches the actual file size.
+    # since this field is unused (I think), it doesn't actually matter other than for reproducing the original cart
+    # I couldn't figure out the pattern behind these fake sizes (if there even is one), so I'm just hardcoding them.
+    fakesize: int
+    # The file ID of the SBN file - needed to guarantee unique filenames. the filename in the header can't be used because it's not unique
+    ident: int
+    def decode(self, data: bytes, ident: int) -> int:
+        self.signature, = struct.unpack(">4s", data[0x00:0x04])
+        self.size, = struct.unpack(">i", data[0x04:0x08])
+        self.name, = struct.unpack(">4s", data[0x08:0x0C])
```
ethteck, line 489:
```suggestion
        self.signature, self.size, self.name = struct.unpack(">4si4s", data[0x00:0x0C])
```

ethteck (changes requested):

tools/splat_ext/pm_sbn.py
Hurricane996: we should probably even use unpack from for this

ethteck (approved):

tools/splat_ext/pm_sbn.py
```diff
@@ -0,0 +1,886 @@
...
+
+        with open(path / "sbn.yaml", "w") as f:
+            # Filename->ID map
+            f.write("# Mapping of filenames to entry IDs. Use 'id: auto' to automatically assign a unique ID.\n")
+            f.write("""
+# 'fakesize is an interesting case. In the final ROM, the size of a file is stored in the file header and the entry table. 
+# For bank files, however, the entry table has seemingly random data stored for file size. This data appears to be unused.
+# Rather than figure out how this data is actually generated, I hardcoded it. What does this mean for modders?
+# Well, for non bank files and bankfiles less than 65536 bytes it means nothing. Simply don't set the fakesize field and it will be fine. 
+# As for bank files that are more than 65536 bytes long, you have to set fakesize to something but it doesn't appear to matter what.\n""")
```
nanaian, line 224: > What does this mean for modders?

Generally, as long as it works, modders aren't bothered about details such as this.

I figure we should treat `fakesize` as a boolean: 
- when size < 65536 bytes, fakesize defaults to 0
- when size >= 65536 bytes, fakesize defailts to 1

tools/splat_ext/pm_sbn.py
```diff
@@ -0,0 +1,886 @@
...
+                f.write(f"    file: {bgm_file_name}\n")
+                f.write(f"    bk_files: [{', '.join(bk_file_names)}]\n")
+            f.write("\n")
+
+            # INIT mseqs
+            f.write("# AuGlobals::mseqFileList. Not sure why there's non-MSEQ files here!\n")
+            f.write("mseqs:\n")
+            for id, entry in enumerate(self.init.mseq_entries):
+                f.write(f"  - id: 0x{id:02x}\n")
+                f.write(f"    file: {self.files[entry].file_name()}\n")
```
nanaian, line 260: lol we should figure this out

tools/build/configure.py
```diff
@@ -1015,6 +1018,12 @@ def build(
                     },
                 )
                 build(entry.object_path, [entry.object_path.with_suffix("")], "bin")
+            elif seg.type == "pm_sbn":
+                sbn_path = entry.object_path.with_suffix("")
+                build(sbn_path, entry.src_paths, "pm_sbn") # could have non-yaml inputs be implicit
```
nanaian, line 1023: The non-yaml inputs _should_ be marked as implicit

tools/splat_ext/pm_sbn.py
```diff
@@ -0,0 +1,886 @@
...
+
+    def __str__(self):
+        return "SBNFile(" + self.file_name() + ")"
+
+    def __hash__(self):
+        return hash((self.signature, self.size, self.name, self.data))
+
+
+# Could extend SBNFile as BGM etc in future if we want higher-level formats
+
```
nanaian, line 525: This comment appears to be lost, maybe turn it into a doc comment on SBNFile

tools/splat_ext/pm_sbn.py
```diff
@@ -0,0 +1,886 @@
...
+        header = INITHeader(*struct.unpack_from(INITHeader.fstring, data))
+
+        data = data[:header.size]
+        # Decode song list
+        song_addr = header.tblOffset
+        song_number = 0
+        while True:
+            song = InitSongEntry(*struct.unpack_from(InitSongEntry.fstring, data, song_addr))
+
+            if(song.bgmFileIndex == 0xffff):
```
nanaian, line 549: No parens needed here - `black` should fix this

nanaian (changes requested):
Note to self: also need to confirm that the asset stack works

tools/splat_ext/pm_sbn.py
ethteck: also this should be bgm_file_index

tools/splat_ext/pm_sbn.py
nanaian: @HailSanta I believe it was you who named `AuGlobals::mseqFileList` - do you recall why you named it that? It seems there are non-MSEQ files in that list

ethteck:
@nanaian the author said they're willing to look at this soon and address the feedback - is there a reason you self assigned this?

nanaian:
Nope, will unassign if chair/Hurricane is happy to update it!

tools/splat_ext/pm_sbn.py
Hurricane996: This directly corresponds to the InitSongEntry::bgmFileIndex field defined in audio.h. As such I think it makes sense to match the name. Same with all the other fields that match to C struct fields

ethteck: ahh I see. okay, that's a good argument 👍

tools/build/configure.py
Hurricane996: I haven'tr eally looked into the build system all that much - in fact I've never worked with ninja before- so I don't quite understand what this means.

ethteck: if I understand alex's message correctly, I think he's referring to one of the optional arguments of build(), which is a list of implicit dependencies...or outputs? can't remember. meowp

nanaian: Yeah, so implicit is a kwarg to this function that takes a list of paths that are to be considered dependencies of this build step but aren't actually included in the build command (in which case they'd just be normal dependencies). Applied to this case, the idea is that the files that are used to produce the SBN besides the yaml file (BGM files, MSEQ files, etc) should be given as `implicit`. See https://ninja-build.org/manual.html#ref_dependencies for more info

ethteck:
super bunny network encoding

tools/build/configure.py
nanaian: Decision on Discord was to implement this using depfiles for all asset types in a future PR. See #1098

nanaian (approved):

PR #1088, opened 2023-07-17
Require pigment64, use it to build most standalone images

ethteck:


nanaian (approved):
https://github.com/pmret/papermario/assets/9429556/6664c490-ef36-45ae-be92-b7a6087f5dde

PR #1090, opened 2023-07-19
Mopping

ethteck:
Removed many hard-coded rom addresses from the US yaml and updated splat to accommodate the change. I hope to eventually remove more and more hard-coded rom addresses, making it easier for us to make yamls for other versions or augment the US one.

I also changed the effect c files to literally refer to the shim functions they call (without the shim_ prefix). This is accomplished the same way as dead.h is - they're renamed when cpp runs over the file.


tools/build/configure.py
```diff
@@ -596,11 +596,7 @@ def build(
                     encoding = "EUC-JP"
 
                 # Dead cod
-                if (
-                    isinstance(seg, segtypes.common.c.CommonSegC)
-                    and isinstance(seg.rom_start, int)
-                    and seg.rom_start >= 0xEA0900
-                ):
+                if "dead" in entry.src_paths[0].parts:
```
nanaian, line 599: Maybe better to use a custom flag on the segment?
like
```
dead_code: true
```

nanaian (changes requested):
We have a lot of segments - not subsegments - with `start: 0xDEADBEEF`... could we change any of these to `start: auto` too or is there something blocking that?

PR #1092, opened 2023-07-20
Shiftable BSS

ethteck:
Although the project has been "shiftable" for a while now, the BSS for the first 3 main overlays was still relatively hard-coded to be certain sizes and positions. This PR makes use of the SHIFT_BSS macro, which uses an attribute to send symbols to the .bss section for shiftable builds and fall back to `extern` on matching builds.

I also was able to clean some things up and remove some things from variables.h.

I also turned off the "redundant declarations" warning for matching builds, as the new externed variables are redundant with various headers at times. I think this is fine at this stage in the project, but I just wanted to call attention to it anyway.


ethteck:
Damn, this actually breaks the shiftable build... @Rainchus do you think you could look at this with me sometime soon?

tools/build/configure.py
```diff
@@ -75,7 +75,7 @@ def write_ninja_rules(
 
     CPPFLAGS = "-w " + CPPFLAGS_COMMON + " -nostdinc"
 
-    cflags = f"-c -G0 -O2 -gdwarf-2 -x c -B {BUILD_TOOLS}/cc/gcc/ {extra_cflags}"
+    cflags = f"-c -G0 -O2 -gdwarf-2 -x c -B {BUILD_TOOLS}/cc/gcc/ {extra_cflags} -Wno-redundant-decls"
```
nanaian, line 78: Move to L1290

tools/build/configure.py
```diff
@@ -195,7 +195,7 @@ def write_ninja_rules(
     ninja.rule(
         "as",
         description="as $in",
-        command=f"{cross}as -EB -march=vr4300 -mtune=vr4300 -Iinclude $in -o $out",
+        command=f"{cpp} {CPPFLAGS} {extra_cppflags} $cppflags $in -o  - | {cross}as -EB -march=vr4300 -mtune=vr4300 -Iinclude -o $out",
```
nanaian, line 198: Why?

ethteck: because we have to remove the contents of the files if it's a shiftable build. See the changes to them

Issue #275, opened 2021-04-22
Documentation site generation (Doxygen / Docco)


nanaian:
Also make available at https://docs.papermar.io

nanaian:
Should also probably have a CI test that checks doxygen isn't broken on PRs

nanaian:
Doxygen is currently dying on the long multiline comment in flo_08/CAED40.c:
```
/github/workspace/src/world/area_flo/flo_08/CAED40.c:203: error: Reached end of file while still inside a (nested) comment. Nesting level 1 (probable line reference: 69) (warning treated as error, aborting now)
```

nanaian:
Need to decide what we're doing w.r.t this. Do we like Doxygen? Do we prefer something like Docco? etc etc

PR #1093, opened 2023-07-21
Remove outdated / unused tools and such from the repo

ethteck:
Removes
* Docker stuff (not used by anyone, not maintained)
* Nix script (not used by anyone, not maintained)
* Doxygen stuff (not working and hasn't been for over 2 years, we'll fix it properly in #275 maybe someday)


nanaian:
IIRC this was asked in Discord already but if @NotNite or @mkst want to maintaining Nix and Docker respectively, that would be appreciated, else we will remove support for them as we can't maintain them. If it's feasible, you could automate testing of them in a similar way we do with Ubuntu on Jenkins.

nanaian (changes requested):
Can you remove https://github.com/pmret/papermario/blob/main/SETUP.md#docker and https://github.com/pmret/papermario/blob/main/SETUP.md#nix from SETUP.md

mkst:
I have no real motivation to support this, so no issues seeing it removed - you guys use your own Jenkins instance with custom image (iirc) so this is superfluous to your needs

PR #1096, opened 2023-07-24
DMA_COPY_SEGMENT macro for straightforward DMAs of segments

ethteck:


include/macros.h
```diff
@@ -330,4 +330,6 @@
 #define DT (1.0f)
 #endif
 
+#define DMA_SEGMENT(segment) dma_copy(segment##_ROM_START, segment##_ROM_END, segment##_VRAM)
```
nanaian, line 333: Can we make the name include "dma_copy"? I initially assumed that this would expand to the three values not wrapped in a dma_copy call

```suggestion
#define DMA_COPY_SEGMENT(segment) dma_copy(segment##_ROM_START, segment##_ROM_END, segment##_VRAM)
```

PR #1100, opened 2023-07-29
Add Python linter to github actions

lshamis:
Adding black to the CI as requested here: https://github.com/pmret/papermario/issues/949
This PR only adds the check, it does not format. That would make this change larger than I'd like.

Note, that it finds unusual code in [`tools/star_rod_idx_to_c.py:422`](https://github.com/pmret/papermario/blob/main/tools/star_rod_idx_to_c.py#L422)
```
error: cannot format tools/star_rod_idx_to_c.py: cannot use --safe with this file; failed to parse source file AST: f-string: invalid syntax (<unknown>, line 422)
```


.github/workflows/lint.yaml
```diff
@@ -4,7 +4,7 @@ on:
 
 jobs:
   cpp_lint:
-    name: Format and lint
+    name: C++ format and lint
```
ethteck, line 7: Why C++?

.github/workflows/lint.yaml
```diff
@@ -33,3 +33,22 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.11
```
ethteck, line 47: We should probably set up 3.8 since that's the minimum version we support

.github/workflows/lint.yaml
```diff
@@ -33,3 +33,22 @@ jobs:
...
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install black
+      - name: Run Black
+        run: |
+          black . --check --exclude tools/splat/ --exclude tools/splat_ext/ --exclude diff.py
```
ethteck, line 54: Is there a reason splat_ext is excluded? Those are our custom segments, though getting path stuff set up correctly might be tricky for that directory

ethteck (changes requested):
Thanks for the PR! A few pieces of feedback

Also, does the new check pass currently? If not, this pr should probably introduce the changes required for it to do so

ethteck:
Also, black is a formatter rather than a linter. For linting, we use mypy. Though I imagine getting that set up and fixing all issues is a lot more work, so we can tackle that later. I'd just rename the pr / step in the yaml to say format rather than lint

.github/workflows/lint.yaml
lshamis: In https://github.com/pmret/papermario/issues/949 [nanaian](https://github.com/nanaian) calls out that "splat/diff.py/friends" should be excluded

.github/workflows/lint.yaml
lshamis: I want to differentiate the name from the new Python format/lint.
The job name is cpp_lint. Would you prefer `Cpp format and lint`?

.github/workflows/lint.yaml
lshamis: Done

lshamis:
> Also, black is a formatter rather than a linter. For linting, we use mypy. Though I imagine getting that set up and fixing all issues is a lot more work, so we can tackle that later. I'd just rename the pr / step in the yaml to say format rather than lint

Renamed `py_lint` -> `py_format`
Changed name `Python format and lint` -> `Python format`

.github/workflows/lint.yaml
ethteck: Yeah, splat ext is distinct from splat though

lshamis:
> Also, does the new check pass currently? If not, this pr should probably introduce the changes required for it to do so

If you like, I can include the ~60 files that would be reformatted.

It would still not pass, though, due to [tools/star_rod_idx_to_c.py:422](https://github.com/pmret/papermario/blob/main/tools/star_rod_idx_to_c.py#L422)
That line doesn't look like valid Python to me.

.github/workflows/lint.yaml
lshamis: Removed the exclude on splat_ext

ethteck:
Could you add the files, and then I'll fix the issue with that one script?

.github/workflows/lint.yaml
ethteck: It should be "C format and lint" since this game doesn't have any c++

lshamis: And yet the job name is cpp_lint :)
Done

lshamis:
> Could you add the files, and then I'll fix the issue with that one script?

Done

.github/workflows/lint.yaml
ethteck: I think that means c preprocessor in this case, but yeah it's a weird name. thanks!

ethteck:
Since we have the "Lint" workflow off (since c linting is a bag of worms we haven't really figured out yet for this project), this won't run after we merge it. I think it'd be fine if we made a Python Lint workflow (a new yaml) and put this there, which we could enable. Would you like me to do that, or do you want to?

ethteck:
I'm also observing that only the most recent instance `--exclude` takes effect. So if I exclude diff.py last, it'll exclude that, and if I exclude another thing last, diff.py says it'll be reformatted. I think we want --extend-exclude for subsequent uses of that flag

marijnvdwerf:
Wouldn't excludes usually be defined in `pyproject.toml`? That way you can also run `black` locally and have it ignore the same files as CI.

ethteck:
I'm not sure, but this isn't really a python project so it seems weird to have such a file for it. I also wouldn't know how to set it up to work with black, but I guess if someone wants to do that, that's cool

ethteck (approved):

lshamis:
I'm not going to be able to make any changes for another day or so. Feel free to make any further changes and commit without me.

ethteck:
shounds good. Thanks for this effort! Definitely long overdue

nanaian:
> Wouldn't excludes usually be defined in `pyproject.toml`? That way you can also run `black` locally and have it ignore the same files as CI.

This would be worth doing before merge I think, to support other editors

nanaian:
good pr though, thanks!

PR #1104, opened 2023-07-29
use enum for EnableActorBlur, EnablePlayerBlur, EnableIdleScript args

Rainchus:


src/18F340.c
```diff
@@ -1299,7 +1299,7 @@ EvtScript EVS_PlayerDies = {
         EVT_CALL(SetActorYaw, ACTOR_PLAYER, LVar0)
         EVT_WAIT(1)
     EVT_END_LOOP
-    EVT_CALL(EnablePlayerBlur, 0)
+    EVT_CALL(EnablePlayerBlur, IDLE_SCRIPT_ENABLE)
```
HailSanta, line 1302: This should be a separate enum specific to this function: BLUR_DISABLE|BLUR_ENABLE|BLUR_RESET

Rainchus: should `IDLE_SCRIPT_RESTART` become `IDLE_SCRIPT_RESET`?

HailSanta: Its not necessary for the enums to be consistent, as they're not really related. The handler for `IDLE_SCRIPT_RESTART` calls `restart_script`, better to be consistent with that.

include/enums.h
```diff
@@ -6472,4 +6472,16 @@ enum Language {
     LANGUAGE_ES = 3,
 };
 
+enum IdleScriptState {
```
ethteck, line 6475: Per the original issue, I
 think we wanted this one in script_api.h

ethteck (changes requested):
Not sure if we also wanna rename the function(s) as mentioned in the issue too. @HailSanta thoughts?

include/enums.h
HailSanta: I've considered renaming them in the past, but decided against it because the vast, vast majority of calls do not use the RESET/RESTART options (3/117) for EnableActorBlur and (67/1060) for EnableIdleScript. I haven't found any alternative names to be a sufficiently compelling improvement over EnableX for a function which is nearly always used to enable and disable.

PR #1105, opened 2023-07-30
item data generation

HailSanta:
All item data is now generated from three yaml files, including the item table, enum, entity scripts, hud scripts, and icon lists. Ranges for item type checks are also auto-generated. The pause map data is also generated from an xml file.


include/enums.h
```diff
@@ -1756,396 +1756,7 @@ enum AuxCameraParams {
...
-#define ITEM_FIRST_CONSUMABLE       ITEM_FIRE_FLOWER
-#define ITEM_LAST_CONSUMABLE        ITEM_JELLY_SHROOM1
-#define ITEM_FIRST_BADGE            ITEM_SPIN_SMASH
-#define ITEM_LAST_BADGE             ITEM_ATTACK_FX_F5
-
-#define ITEM_NUM_KEYS (ITEM_LAST_KEY - ITEM_FIRST_KEY + 1)
-#define ITEM_NUM_CONSUMABLES (ITEM_LAST_CONSUMABLE - ITEM_FIRST_CONSUMABLE + 1)
-#define IS_ITEM(itemID) (itemID >= ITEM_FIRST_KEY && itemID <= ITEM_LAST_CONSUMABLE)
-#define IS_BADGE(itemID) (itemID >= ITEM_FIRST_BADGE && itemID <= ITEM_LAST_BADGE)
+#include "gen/item_enum.h"
```
ethteck, line 1759: I'm not sure we need to put generated includes/headers in a "gen" directory. For the effect stuff I recently did, I didn't do that. so if we do want to, we should probably be consistent..but I'd vote to just put this not in one

src/pause/pause_map.c
```diff
@@ -21,94 +21,24 @@ void pause_map_handle_input(MenuPanel* tab);
...
-    { .pos = { .x = 158, .y = 110 }, .parent =  2, .pathLength =  6, .path = gPauseMapPaths[29], .afterRequirement = STORY_61,                          .id = LOCATION_STAR_WAY },
-    { .pos = { .x = 163, .y =  78 }, .parent = 29, .pathLength =  6, .path = gPauseMapPaths[30], .afterRequirement = STORY_61,                          .id = LOCATION_STAR_HAVEN },
-    { .pos = { .x =  96, .y =  49 }, .parent = 30, .pathLength = 13, .path = gPauseMapPaths[31], .afterRequirement = STORY_EPILOGUE,                    .id = LOCATION_BOWSERS_CASTLE },
-    { .pos = { .x =  88, .y =  22 }, .parent = 31, .pathLength =  3, .path = gPauseMapPaths[32], .afterRequirement = STORY_EPILOGUE,                    .id = LOCATION_PEACHS_CASTLE },
-    { .pos = { .x =  98, .y = 147 }, .parent =  0, .pathLength =  0, .path = gPauseMapPaths[33], .afterRequirement = STORY_61,                          .id = LOCATION_MARIOS_HOUSE }
-};
 
-s32 gPauseMapArrowWobble[] = { 0, 2, 3, 3, 4, 4, 4, 4, 3, 2, 1, 0 };
-MenuWindowBP gPauseMapWindowBPs[] = {
+#include "gen/world_map.h"
```
ethteck, line 38: Header files shouldn't contain declarations, so this would be an .inc.c rather than a .h imo

tools/build/configure.py
```diff
@@ -548,6 +563,34 @@ def build(
...
 
+        build(
+            self.build_path() / "include/gen/world_map.h",
+            [Path("src/gen/world_map.xml")],
+            "world_map",
+        )
+
+        build(
+            [
+                self.build_path() / "include/gen/move_data.h",
```
ethteck, line 574: data should be an inc.c and enums .h

tools/build/pm_icons.py
```diff
@@ -48,10 +48,10 @@ def build(out_bin: Path, in_xml: Path, out_header: Path, asset_stack: Tuple[Path
         file = Icon.attrib["name"]
 
         if file is None:
-            raise Exception("Icon os missing attribute: 'name'")
+            raise Exception("Icon is missing attribute: 'name'")
```
ethteck, line 51: lol good catch

ethteck (changes requested):
Lgtm! just some renaming / moving suggestions. I also would prefer the build scripts not be prefixed with `pm_` - the reason we do that for the splat extensions is to more obviously distinguish them from the base segment types.

marijnvdwerf:
Are we sure the other releases don't have any changes in these tables?

src/gen/item_table.yaml
```diff
@@ -0,0 +1,5459 @@
...
+    sellValue: -1
+    sortValue: 0
+    targetFlags: [ TARGET_FLAG_2 ]
+    typeFlags: [ ITEM_TYPE_FLAG_GEAR, ITEM_TYPE_FLAG_ENTITY_FULLSIZE ]
+    moveID: MOVE_NONE
+    potencyA: 0
+    potencyB: 0
+- FirstDegreeCard:
+    category: QUEST
+    nameMsg: MSG_ItemName_FirstDegreeCard
```
marijnvdwerf, line 153: Can't most of these be generated using `MSG_ItemName_{{ name }}`?

ethteck: good point! If there are exceptions, you can just leave those in and make the script generate the name according to that pattern if the nameMsg field is omitted

tools/build/configure.py
HailSanta: Agreed, I couldn't get configure to build with inc.c before, but its working now.

src/gen/item_table.yaml
HailSanta: In many cases, yes, but I don't think they should necessarily be so, especially for modders. Many of them are missing/left blank and use MSG_NONE, which is also the default value atm if the field is left out.

include/enums.h
```diff
@@ -232,7 +232,7 @@ enum StoryProgress {
     STORY_CH8_REACHED_BOWSERS_CASTLE            = 94,
     STORY_CH8_REACHED_PEACHS_CASTLE             = 95,
     STORY_EPILOGUE                              = 96,
-    STORY_61                                    = 97,
+    STORY_NEVER                                 = 97,
```
ethteck, line 235: maybe STORY_END is better

src/world/area_mac/mac_02/npc/tayce_t.inc.c
```diff
@@ -4,340 +4,26 @@
...
-    ITEM_HONEY_SHROOM,
-    ITEM_BLAND_MEAL,
-    ITEM_THUNDER_RAGE,
-    ITEM_EGG_MISSILE,
-    ITEM_APPLE_PIE,
-    ITEM_KOOKY_COOKIE,
-    ITEM_BIG_COOKIE,
-    ITEM_SPECIAL_SHAKE,
-};
+#include "recipes.h"
```
ethteck, line 26: should be an .inc.c

include/enums.h
HailSanta: Its not for the end of the story, its intended as a value which is never reached.

src/world/area_mac/mac_02/npc/tayce_t.inc.c
HailSanta: Same problem as map table

PR #1111, opened 2023-08-13
Match libultra iQue ver

Alto1772:
Matches **most** of the iQue libultra functions...

Only the C files located in `src/os/` that are affected have been replaced using the sources listed below. Associated header files corresponding to these C files have also been occasionally updated.

There are some differences with the code in the final release of iQue `libultra_rom.a` and in the iQue version of Paper Mario, which I have marked them with `VERSION_IQUE`. These files are modified to match with the aformentioned rom, including the compiler flags.

Here are the sources I've utilized to ensure where all of these files gathered from:
- [ultralib ique branch](https://github.com/decompals/ultralib/tree/ique) (used for the most part)
- [decomp.me scratch](https://decomp.me/scratch/JsOS1) (used for strcmp.c, common.c; good thing this one just mysteriously appeared in their database)

Please notify me of any potential issues that some of you may doubt, and I would appreciate any assistance in resolving them.


BowserSlug:
src/os/xldtob.c:158: warning: initialization from incompatible pointer type
src/os/xldtob.c:158: warning: initialization from incompatible pointer type
src/os/sirawdma.c:22: warning: implicit declaration of function `skKeepAlive'
src/os/devmgr.c:86: warning: passing arg 3 makes pointer from integer without a cast
src/os/devmgr.c:101: warning: passing arg 4 makes pointer from integer without a cast
src/os/xldtob.c:158: warning: initialization from incompatible pointer type

ver/ique/splat.yaml
```diff
@@ -377,7 +403,13 @@ segments:
     - [0x729B0, .rodata, audio/reverb]
     - [0x72A10, .rodata, 5B320]
     - [0x72A80, .rodata, world/world]
-    - [0x73B10, rodata, libultra]
+    - [0x73B10, .rodata, os/lookathil]
```
ethteck, line 406: The entire .rodata list can be removed and you can just end change the end segment marker to the first .rodata section. so like after the last .data, just do - [0x71DB0]

ethteck:
welcome back to the project :) happy to see you around 

this is really exciting stuff - thanks for all your hard work! I'm not sure why Jenkins isn't reporting the change in match%, and that concerns me a bit. I'll look into it before I do my more thorough review within the next day or so

src/os/xlitob.c
```diff
@@ -50,9 +46,7 @@ void _Litob(_Pft *args, char type) {
     }
 
     if (args->prec < 0 && (args->flags & (FLAGS_ZERO | FLAGS_MINUS)) == FLAGS_ZERO) {
-        i = args->width - args->n0 - args->nz0 - args->n1;
-
-        if (i > 0) {
+        if ((i = args->width - args->n0 - args->nz0 - args->n1) > 0) {
```
ethteck, line 49: was this kind of change necessary to match the file for iQue? I'd prefer to keep unnecessary changes out. though if this is how ultralib did it, I guess that's fine

src/os/xldtob.c
```diff
@@ -177,18 +179,17 @@ void _Genld(_Pft* px, char code, u8* p, s16 nsig, s16 xexp) {
...
-        xexp += 1;
+        xexp++;
         if (code != 'f') {
             if (((px->flags & 8) == 0) && nsig < px->prec) {
                 px->prec = nsig;
             }
 
-            px->prec -= xexp;
-            if (px->prec < 0) {
+            if ((px->prec -= xexp) < 0) {
```
ethteck, line 192: same here - not a fan of the assignments inside conditions if we can avoid it

ethteck (changes requested):
My main feedback here is that I'd prefer to not change the formatting / styling of things unless it was necessary to match ique or there's some other reason I'm unaware of.

I see lots of extra added newlines, assignments in if conditions, etc. If those can be reverted (or you have a specific reason for doing that), that'd be ideal.

BowserSlug:
This PR matches -54 functions (+1.19%) on `ique`.

ethteck (approved):

HailSanta (approved):

PR #1113, opened 2023-08-21
Build shift US on Jenkins

nanaian:


nanaian:
> Exception: SBN directory assets/us/audio already exists

a bug!

PR #1118, opened 2023-09-28
last actors

HailSanta:


src/actor_api.c
```diff
@@ -2,8 +2,8 @@
 #include "effects.h"
 #include "battle/battle.h"
 
-extern s32 D_8029FBD0;
-extern s8 D_8029FBD4;
+extern s32 isGroupHeal;
```
ethteck, line 5: maybe IsGroupHeal since this is a global?

ethteck (changes requested):
kyool

src/actor_api.c
HailSanta: Good catch. Annoyingly, its only used in a single function and has no business being a global. Alas.

ethteck: this and everything after it in the file can possibly be declared in-function static if we set it up correctly. it'd be good to do a pass of that on the whole game at some point (for symbols only used in one function)

HailSanta: Agreed, I'd love to see that. Might help us find some "hidden" file splits.

src/battle/area/omo2/actor/boss_common.h
```diff
@@ -0,0 +1,49 @@
+#ifndef _GENERAL_GUY_COMMON_H_
+#define _GENERAL_GUY_COMMON_H_
+
+#include "mapfs/omo_bt07_shape.h"
+
+enum N(BattleCommon) {
```
nanaian, line 6: `=` alignment is inconsistent here

nanaian (changes requested):

src/battle/area/omo2/actor/boss_common.h
HailSanta: The intent is to add an additional indent level for the enumerated values of each actor var.

nanaian: ah, I see! 👍

nanaian (approved):

PR #1122, opened 2023-10-24
Allow SBN files to be reencoded as little endian

Hurricane996:


ethteck:
I'd prefer that this be done the way I've done it in the other scripts like sprite_shading_profiles.py. `END = ">" if endian == "big" else "<"` and then prepend END to the string

ethteck:
this should also be hooked up to sbn.py and have an arg there so the little endian code path can be reached.

ethteck:
@Hurricane996 are you planning on working on this still? No pressure, just wondering since I haven't heard anything since my review

PR #1135, opened 2023-12-25
JP pm_msg

Alto1772:
There were some things that I've encountered along the way:
- `parse_compile.py` is messy and needs some refactoring
- Contents of `msg_jp.yaml` has not been fully completed yet


BowserSlug:
✅ This PR fixes 1 warnings!

PR #1140, opened 2024-01-07
Assetify crash font

nanaian:
Depends on https://github.com/decompals/pigment64/pull/16.

Have to do a cast because there's no way to ask splat to emit images as u32 arrays right now (although pigment64 has C array format support?).


ver/us/splat.yaml
```diff
@@ -369,7 +369,12 @@ segments:
       - [0x511E0, vtx, theater/wall_shadows_vtx]
       - [0x51260]
       - [0x51540, ia8, ui/no_controller, 128, 32]
-    - [auto, .data, crash_screen]
+    - start: auto
+      type: .data
+      name: crash_screen
+      subsegments:
+      - [0x52690, i1, crash_screen/font, 32, 63]
```
ethteck, line 376: the start addr here can be auto I'd think

ethteck: nm

ethteck:
this doesn't match unfortunately. the new rom is larger than the baserom, and it starts I think around the crash screen font area

ver/us/splat.yaml
nanaian: This gives `Error: crash_screen/font should end at 0x4FADC, but it ends at 0x5278C`

PR #1141, opened 2024-01-07
Image asset rework

ethteck:
This PR accomplishes 2 things:
* Removes all *png.inc.c and *pal.inc.c files from the build, replacing this with simply `.incbin`ing the *.png.bin and *pal.bin files.
* Moves the symbol name for image assets into the c source files, reducing the burden of symbol_addrs.txt needing to store this information across multiple versions. This should make other version support easier as we go.

I plan to do step 2 for vtx and gfx assets soon as well.


.vscode/settings.json
```diff
@@ -69,4 +69,5 @@
     "black-formatter.args": [
         "-l 120"
     ],
+    "search.useIgnoreFiles": false,
```
nanaian, line 72: Can we justify enabling this for all users? I feel like build files should not show up in search by default as they're only relevant if you're debugging the build system

nanaian (approved):

PR #1145, opened 2024-01-09
yaml-ify map metadata & iQue pm_map

Alto1772:
Replaces hardcoded offsets and sizes for each version in the script for this reason.

In the continuation of the tasks listed on #1126 this also resolves by dumping the raw compressed title data section from the baserom, then at the map segment building step we first write the mentioned raw dump to make out with the unusable data at the end, then overwrite in place with our newly formed title data using the new size as the result. It's likely possible to get rid of this dump method entirely in which we could find out how to manage those unused garbage data at the end.

and yeah be wary that the file `sbk_bg.alt.png` has been renamed to `sbk_bg.1.png`. lmk what any other suggestions or changes I've missed on or if any sections of this wasn't in ur taste just tell me.


ethteck:
lgtm! I'll wait for another review or two though

tools/splat_ext/mapfs.yaml
```diff
@@ -0,0 +1,1040 @@
+- [dro_01_shape, geom_shape]
```
nanaian, line 1: What's the reason for adding the type (e.g. `geom_shape`) to this list? The type can always(?) be inferred from the name suffix and the engine assumes particular suffixes

nanaian (changes requested):
LGTM besides comment

tools/splat_ext/mapfs.yaml
Alto1772: Because I wanted to make it like the same functionality from splat and I didn't know the engine could take care of identifying types by names so I added it for this reason. I'll just remove the type metadata completely...

nanaian: Fair reasoning, but yeah, no need :+1:

nanaian (approved):
woo

PR #1146, opened 2024-01-09
support modded NPC sprites

nanaian:
- improve compatibility with Star Rod SpriteSheet.xml files
  - SR's intention is to move to the decomp xml but the current release of SR emitted incompatible xml
- use npc.xml instead of npc_sprite_names.yaml to generate linker entries. this allows mods to add new sprites


tools/splat_ext/pm_sprites.py
```diff
@@ -832,11 +835,17 @@ def split(self, rom_bytes) -> None:
...
         from splat.segtypes.linker_entry import LinkerEntry
+        import splat.scripts.split as split
 
         src_paths = [options.opts.asset_path / "sprite"]
 
-        # for NPC
-        src_paths += [options.opts.asset_path / "sprite" / "npc" / sprite_name for sprite_name in self.npc_cfg]
+        # read npc.xml - we can't use self.npc_cfg because nonvanilla asset packs can change it
+        # for each sprite, add to src_paths
+        asset_stack = tuple(Path(p) for p in split.config["asset_stack"])
```
ethteck, line 844: In the future, I think we should have a "custom options" mechanism that exposes project-defined options to segments, so we don't have to do stuff like this

ethteck (approved):
looks grood

tools/splat_ext/pm_sprites.py
nanaian: agreed

PR #1158, opened 2024-01-22
Support Nix package manager

nanaian:

### DIFF
--- a/tools/build/configure.py
+++ b/tools/build/configure.py
@@ -365,7 +365,7 @@ def does_iconv_work() -> bool:
         sub = subprocess.run(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, input=stdin, cwd=ROOT)
         return sub.stdout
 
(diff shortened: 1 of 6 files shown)


tools/build/configure.py
```diff
@@ -362,10 +362,10 @@ def does_iconv_work() -> bool:
     stdin = "エリア ＯＭＯ２＿１".encode("utf-8")
 
     def run(command, stdin):
-        sub = subprocess.run(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, input=stdin)
+        sub = subprocess.run(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, input=stdin, cwd=ROOT)
         return sub.stdout
 
-    expected_stdout = run(["tools/build/iconv.py", "UTF-8", "CP932"], stdin)
+    expected_stdout = run(["python3", "tools/build/iconv.py", "UTF-8", "CP932"], stdin)
```
ethteck, line 368: is there a reason for this? the shbang should use the correct python executable. not all python 3s are named "python3"

PR #1160, opened 2024-01-26
Fix Icons.xml not respecting asset stack

bates64:

### DIFF
--- a/tools/build/configure.py
+++ b/tools/build/configure.py
@@ -253,7 +253,7 @@ def write_ninja_rules(
 
     ninja.rule(
         "icons",
-        command=f"$python {BUILD_TOOLS}/icons.py $out $list_path $header_path $asset_stack",
+        command=f"$python {BUILD_TOOLS}/icons.py $out $header_path $asset_stack",
     )
 
     ninja.rule(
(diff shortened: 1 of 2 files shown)


tools/build/configure.py
```diff
@@ -941,7 +941,7 @@ def build(
                     entry.src_paths,
                     "icons",
                     variables={
-                        "list_path": entry.src_paths[0],
+                        "list_path": self.resolve_asset_path(entry.src_paths[0]),
```
ethteck, line 944: We pass the unresolved path to the script because it's supposed to resolve it. So if it's not doing that, I think it's a bug in the build script

nanaian: Ah, that makes sense. After spending so much time tracking this one down I didn't consider that!
Perhaps unresolved asset paths should be `assets/__not_resolved__/...` instead of defaulting to the final path in the stack, so that we can catch bugs like this in future

ethteck: I guesh we could do that, but it's the path of least resistance to not mutate the path before giving it to that script

nanaian: I think there's a misunderstanding here. I'll update the script to do the resolving

ethteck:
my memory might be wrong, but I believe what needs to be passed to the script is now just `icon/Icons.xml` rather than `assets/us/icon/Icon.xml` as it is currently

PR #1164, opened 2024-02-03
iQue pm_msg

Alto1772:
Due to usage of OCR there might be inconsistencies of some of each characters. If there is one, tell me about it.


bates64 (approved):

ethteck:
this is so cool to have in the project. thanks as always for your contributions :)

PR #1173, opened 2024-02-17
`SHIFT_BSS` be gone

Alto1772:


BowserSlug:
🚀🚀🚀🚀🚀🚀🚀🚀 This PR matches 8 functions (+0.00%) on `jp`.
🚀 This PR matches 1 function (+0.00%) on `ique`.

tools/build/configure.py
```diff
@@ -69,7 +69,7 @@ def write_ninja_rules(
 
     CPPFLAGS_272 = CPPFLAGS_COMMON + " -nostdinc"
 
-    CPPFLAGS_EGCS = CPPFLAGS_COMMON + " -D__USE_ISOC99 -DBBPLAYER -nostdinc"
+    CPPFLAGS_EGCS = CPPFLAGS_COMMON + " -D__USE_ISOC99 -nostdinc"
```
ethteck, line 72: is there a reason we can't just always define BBPLAYER for this compiler like we used to?

ethteck (commented):
Echoing what I said in discord:

I am all for the asm moving into src/ changes

I'll need some more time with it regarding all the BSS stuff

my main concern with this approach is the fact that making things static unfortunately makes them less visible. However, it seems like SHIFT_BSS was defined as just `static` in some cases anyway...hmm

src/os/nusys/nusched.c
```diff
@@ -3,21 +3,23 @@
...
 NOP_FIX
 
+NUSched nusched;
 NUScPreNMIFunc nuScPreNMIFunc = NULL;
+u8 nuScPreNMIFlag;
 
 char nusys_version[] = "NuSystem2.05";
 
-u32 nuScRetraceCounter = (u32) nusys_version; // wtf?
+u32 nuScRetraceCounter = (u32) nusys_version;
```
ethteck, line 12: does this make sense to you? I think I'd prefer we keep the comment here unless it has some reasonable explanation

tools/build/configure.py
Alto1772: I've moved that define to use for both the C code and HASM code inside the `os/` folder, and into making it easier to create a bootable iQue version for N64 emulators and into real hardware.

src/os/nusys/nusched.c
Alto1772: Yeah, but It's in the original NuSYS 2.05 code that it was actually forced intact and not a coincidence, as far my understanding goes. It's conflicting whether or not I should remove that comment intentionally, as I wanted to make it match the original code.

But you've actually noticed this so I will take action to remove this change now...

Alto1772:
> my main concern with this approach is the fact that making things static unfortunately makes them less visible. However, it seems like SHIFT_BSS was defined as just `static` in some cases anyway...hmm

Yeah good suggestion. The reason I did is to just be made as intended by the original developers. The BSS addresses located between the 2 common sections were originally meant to be static so I made it one too...

As a suggestion and my personal understanding, I'll just make the static ones visible by using BSS macro instead.

src/os/nusys/nusched.c
ethteck: Ah, I didn't realize it was like this in the actual code. In that case the comment can definitely be removed... Sorry to make you undo your change again

ethteck (approved):

PR #1175, opened 2024-02-21
Clean up map filenames

bates64:
- renames map files to not include the map name in the file name, because they're in that directory
- removes the 0_ 1_ 2_ etc prefix for map filenames because
  - this convention is not used anywhere else
  - it's not interesting or useful to know the order in which they are linked
  - if you really want to know, you can check splat
- renames \*header.c to \*settings.c
  - MapHeader was renamed to MapSettings
  - don't want to overload "header"


bates64:
I'd also like to change the header file names to be consistent but that's another PR

bates64:
Need help with iQue

PR #1176, opened 2024-02-27
document D_802946E0

bates64:

### DIFF
--- a/src/battle/1A5830.c
+++ b/src/battle/1A5830.c
@@ -2701,7 +2701,7 @@ API_CALLABLE(RemoveActor) {
 
 API_CALLABLE(DropStarPoints) {
     /// Star Point multiplier, indexed by actor count.
-    /// +10% multiplier for two actors
+    /// +10% multiplier for four actors
     /// +30% multiplier for three or more actors
     static s32 multiplier[] = {
         100, 100, 100,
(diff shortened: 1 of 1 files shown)


src/battle/1A5830.c
```diff
@@ -2702,6 +2700,15 @@ API_CALLABLE(RemoveActor) {
 }
 
 API_CALLABLE(DropStarPoints) {
+    /// Star Point multiplier, indexed by actor count.
+    /// +10% multiplier for two actors
```
HailSanta, line 2704: Isn't this 10% for 3, 30% for 4+?

bates64: Don't know how I managed to mess this one up, lol

bates64 (commented):
silly me

src/battle/1A5830.c
```diff
@@ -2702,6 +2700,15 @@ API_CALLABLE(RemoveActor) {
 }
 
 API_CALLABLE(DropStarPoints) {
+    /// Star Point multiplier, indexed by actor count.
+    /// +10% multiplier for four actors
+    /// +30% multiplier for three or more actors
+    static s32 multiplier[] = {
```
ethteck, line 2706: I'd rather this not be static because it hides it from tools and other files. We could call it star point multiplier

PR #1179, opened 2024-03-05
Unhardcode many for loop conditions

Rainchus:


src/draw_box.c
```diff
@@ -474,8 +474,8 @@ s32 draw_box(s32 flags, WindowStyle windowStyle, s32 posX, s32 posY, s32 posZ, s
                 guMtxCatF(mtx2, rotScaleMtx, mtx1);
             }
             if (outMtx) {
-                for (i = 0; i < 4; i++) {
-                    for (j = 0; j < 4; j++) {
+                for (i = 0; i < COUNT_ROWS(mtx1); i++) {
```
ethteck, line 477: I think these would be better as ROW_COUNT & COL_COUNT , but also not so sure how necessary they are. it's not like you'd want to change these values

src/effects/motion_blur_flame.c
```diff
@@ -157,7 +157,7 @@ void motion_blur_flame_appendGfx(void* effect) {
     gSPSegment(gMainGfxPos++, 0x09, VIRTUAL_TO_PHYSICAL(((EffectInstance*)effect)->graphics->data));
     gSPDisplayList(gMainGfxPos++, D_E00A29D0[temp_a2]);
 
-    for (i = 0; i < 4; i++) {
+    for (i = 0; i < Unk_Array_Size1; i++) {
```
ethteck, line 160: I'd prefer all caps for this

ethteck (changes requested):

src/draw_box.c
bates64: I disagree

bates64 (approved):

PR #1185, opened 2024-04-25
revert texture renaming

z64a:
Removes stripping area name prefix and file extension suffix from textures before dumping and reapplying them during build. This behavior created "nicer" texture names, but has been a headache for modders and custom map/tex assets.


bates64:
Note to users: `./configure --clean` is required and you might need to fix custom textures too

Issue #1059, opened 2023-05-20
Transform install.sh into a markdown guide


ethteck:
To quote me in #1051 :

> I'll try to explain my reasoning for why a human-readable doc is better than a bash script in this case.
> 
> For starters, we end up installing things without really informing the user, running sudo. I think this is pretty bad practice, and I know it's been raised before as something to consider. I also think it's just generally better to have users install their own dependencies instead of running a bunch of commands for them. We of course can provide said commands, but running the commands automatically in their native environment to install system packages has always felt a bit intrusive to me.
> 
> Also, as we see in cases like these, if the feedback we expect doesn't exactly match what the bash script checks, the script just totally gives up or fails. Whereas a human-readable doc might lead someone to the correct solution.
> 
> In both cases, insufficiencies in the script/doc would lead people to ask for help and then we'd fix said script or doc. What's nice about the doc, though, is that it allows us to easier add notes and caveats that someone might care about instead of littering the shell script with logic and/or print-outs that will be drowned out by the other output from commands.
> 
> I guess you could ask "why not both?" and to that I'd say that maintaining two things is even worse than one. So I'd really prefer not to do that, but I acknowledge it'd be the best of both worlds...just at the expense of 2x the maintenance work.

After #1055 is merged, install.sh will only contain external (system) dependencies, at which point it can be replaced with a guide that can help people obtain the required dependencies for the project.

PR #1187, opened 2024-05-28
Update install_deps.sh

matejsmycka:

### DIFF
--- a/install_deps.sh
+++ b/install_deps.sh
@@ -21,7 +21,7 @@ if [[ "$uname" == "Darwin" ]]; then
 
     # Install packages
     brew install md5sha1sum ninja gcc nanaian/brew/mips-linux-gnu-binutils nanaian/brew/mips-linux-gnu-gcc || exit 1
-    python3 -m pip install -U -r requirements.txt || exit 1
+    python3 -m pip install -U -r requirements.txt || python3 -m pip install -U -r requirements.txt --break-system-packages || exit 1
     cp tools/precommit_check_no_assets.sh "$(git rev-parse --git-path hooks)/pre-commit" || exit 1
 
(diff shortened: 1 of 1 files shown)


ethteck:
I appreciate the intent behind this pr, but I don't think we should be automatically running pip install with the break system packages flag.

We have an open issue to rewrite this into a document that can be followed and doesn't have people mindlessly running `sudo some script.sh`. I think it'd be great to turn this into that document so we can explain things like why you might need to run pip install with this flag and what that means.

matejsmycka:
I understand. While this initiative is great, users should be able to follow the guide or run a random script; taking responsibility for the script, the user should decide which approach he prefers.

About pip.

 You are inherently breaking system dependencies using pip. Either you are using a proper virtual environment or want it to "just work".  The current approach is halfway. My approach is to "just work". Maybe a better way would be to warn the user if command `x` doesn't work, use `--break-dependencies`. Or add poetry or pyenv....

Darxoon:
If we are forcing everyone who wants to use the decomp to run a random shell script then we should really make sure it doesn't break people's systems. There's a reason `break-system-packages` is called the way it is

matejsmycka:
Yes, but the current state still breaks them.

Darxoon:
No? You will need to setup a venv before running install_deps.sh but it won't be able to harm the stability of your linux installation in general like `break-system-packages` has the potential to.

Remember that install_deps.sh is ran by a lot of people locally on their machines and using this flag carelessly like in your PR will inevitably make someone have to reinstall their OS because this flag *has* the potential to brick your entire linux installation if you're unfortunate.

matejsmycka:
This is implicit info afaik

bates64:
See #1059

Issue #1186, opened 2024-05-04
Enforce up to date dependencies


Darxoon:
Currently, when papermario's dependencies (which are installed with install_deps or install_copilers) are out of date and not on the correct version, splat and ninja will just fail with obscure error messages, e.g.:

<details>

<summary>Error message with splat</summary>

    configure: configuring version us
    splat 0.21.5 (powered by spimdisasm 1.20.0)
    Loading symbols (symbol_addrs): 100%|████████████████████████████████████████████████████████████████████████████████████| 26306/26306 [00:01<00:00, 15187.18it/s]
    Scanning charset:   1%|█                                                                                                        | 11/1037 [00:00<00:46, 22.14it/s]
    Traceback (most recent call last):
      File "/home/laura/projects/papermario/tools/build/configure.py", line 1440, in <module>
        configure.split(not args.no_split_assets, args.split_code, args.shift, args.debug)
      File "/home/laura/projects/papermario/tools/build/configure.py", line 424, in split
        split.main(
      File "/home/laura/projects/papermario-venv/lib/python3.11/site-packages/splat/scripts/split.py", line 528, in main
        do_scan(all_segments, rom_bytes, stats, cache)
      File "/home/laura/projects/papermario-venv/lib/python3.11/site-packages/splat/scripts/split.py", line 280, in do_scan
        segment.scan(rom_bytes)
      File "/home/laura/projects/papermario-venv/lib/python3.11/site-packages/splat/segtypes/common/group.py", line 131, in scan
        sub.scan(rom_bytes)
      File "/home/laura/projects/papermario/tools/splat_ext/pm_charset_palettes.py", line 27, in scan
        palette = N64SegPalette.parse_palette_bytes(data[i : i + 0x10])
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    AttributeError: type object 'N64SegPalette' has no attribute 'parse_palette_bytes'

</details>
<details>

<summary>Error message with ninja</summary>

    [1/9486] cc_272 src/os/epirawdma.c
    FAILED: ver/us/build/src/os/epirawdma.c.o 
[...]

PR #1189, opened 2024-06-02
Use crunch64 for compression

ethteck:


bates64:
LGTM, but can we also have configure check if crunch64 is installed (https://github.com/pmret/papermario/issues/1186)?

bates64 (approved):

PR #1198, opened 2024-09-05
[JP] Battle Code and Battle Areas

mrshigure:


tools/build/configure.py
```diff
@@ -620,16 +620,28 @@ def build(
...
+            build(
+                [
+                    self.build_path() / "include/battle/actor_types.inc.c",
+                    self.build_path() / "include/battle/actor_types.h",
+                ],
+                [
+                    Path("src/battle/actors.yaml"),
+                ],
+                "actor_types",
+            )
```
mrshigure, line 644: Had to create a new actors yaml file for JP. Not sure if this is the way we want to handle that or if it's better to somehow support all versions in the same file.

ethteck: I think for now this is fine and we can handle it later when we handle other assets unification between versions

PR #1202, opened 2024-09-12
Fix charset craziness

ethteck:
This is a loooong time coming, and although I was reluctant to do it, I'm so glad I did. Fixes several shiftability issues with charsets.


bates64 (approved):
gr8
3 alignment on INCLUDEd data is crazy

PR #1205, opened 2024-09-12
fix LoadPartyImage for shift builds

Rainchus:


src/world/common/todo/LoadPartyImage.inc.c
```diff
@@ -5,6 +5,35 @@
 #error "Define PARTY_IMAGE to the asset name to use LoadPartyImage."
 #endif
 
+#ifdef SHIFT
+//this is required on modern compilers to guarantee the raster comes directly after the palette
+typedef struct Img {
+    u16 palette[256];
```
z64a, line 11: use PAL_BIN and IMG_BIN types here

PR #1212, opened 2024-10-03
npc-enemy-flags

z64a:
Named the remaining NPC, ENEMY, and EVT_GROUP flags.


include/evt.h
```diff
@@ -129,7 +129,7 @@ enum EventGroupFlags {
     EVT_GROUP_00    = 0x00,
     EVT_GROUP_0A    = 0x0A, // 8 | 2
     EVT_GROUP_0B    = 0x0B, // 8 | 4 | 1
-    EVT_GROUP_1B    = 0x1B, // 10 | 8 | 4 | 1
+    EVT_GROUP_1B    = 0x1B, // EVT_GOUP_EXIT_MAP? 10 | 8 | 4 | 1
```
bates64, line 132: Goup
green soup

include/functions.h
```diff
@@ -73,7 +73,11 @@ void player_input_to_move_vector(f32* angle, f32* magnitude);
 void game_input_to_move_vector(f32* x, f32* y);
 void exec_ShakeCamX(s32 arg0, s32 arg1, s32 arg2, f32 arg3);
 void exec_ShakeCam1(s32 arg0, s32 arg1, s32 arg2);
-f32 func_800E5348(void);
+
+// returns the angle that the player *sprite* is facing in world-space, this is NOT the direction the player character
```
bates64, line 77: Probably worth specifying that this returns the target direction of the sprite, i.e. whilst turning around it interpolate

bates64 (commented):
Would be great to add the renamed enum fields to [renames](https://github.com/pmret/papermario/tree/main/tools/renames)

include/functions.h
z64a: It's not. The target sprite direction is into or out of the screen, while this is to the left or right of the player depending on which direction they are facing. This function is used for hammer and parasol collision checks, not to find the target for interpolation. Attached is a picture demonstrating this. I am walking toward the camera (yaw = 180), but facing left (yaw 270).
forward

bates64: My comment was meant to be phrased as a question, but I failed at that somehow.. Thanks

bates64 (approved):

PR #1226, opened 2024-12-26
Use SPRITE_HEAP_SIZE in sprite.c _heap_create

ethteck:


bates64 (approved):
This is exactly what we do in papermario-dx and it works great :shipit:

PR #1230, opened 2025-03-09
sprite fixes

z64a:
- fix several bugs in the fields of generated sprite xml files
- add support for SetRaster/SetPalette/SetParent to use names instead of IDs
- add support for Keyframe commands (only when building, not when generating)


ethteck:
awesome :)

PR #1241, opened 2025-07-24
chore: replace true and false

### LINKED ISSUE #1169: Lowercase TRUE and FALSE
Yay C23!!

a2937:
I feel like I did something not quite right here in this PR given all the changed files. 

Closes #1169


ethteck:
The changes here look good to me. This project just has an enormous number of source files. Unfortunately the build is failing for some reason due to "bool" being defined twice. I'm trying to figure out where the other place is.,,

a2937:
I don't mind at all. The only thing I care about is the code working.

bates64 (approved):

PR #1244, opened 2025-08-13
Nix flake

bates64:
- `shell.nix` is now `flake.nix`
  - use `nix develop` instead of `nix-shell`
- ELFs can be built with `nix build .#papermario_us` (or any other verison)
- Default development shell requires all baseroms
- Uses dream2nix to handle Python dependencies, instead of every `nix develop` doing `pip install` which was awful


ethteck:
It'd be nice to have the issues not related to nix in a different PR because there's already a lot going on with the nix-related changes

bates64:
The configure changes are required, e.g. passing CLFAGS as an environment variable is needed for the libpapermario package. We should be following this anyway.

I don't know how else you solve the dependency problem without making configure into a pyproject. Are there any actual downsides?

tools/configure/src/configure/main.py
```diff
@@ -664,12 +667,23 @@ def build(
 
             assert entry.object_path is not None
 
+            vram_class = seg.get_most_parent().vram_class
```
ethteck, line 670: this can be removed and the overlays dict above

tools/configure/src/configure/main.py
```diff
@@ -330,6 +327,11 @@ def write_ninja_rules(
 
     ninja.rule("pm_sbn", command=f"$python {BUILD_TOOLS}/audio/sbn.py $out $asset_stack")
 
+    ninja.rule("archive",
```
ethteck, line 330: can be removed

ethteck (changes requested):
from our discord convo, theez

ethteck (changes requested):
looks grood, but the configure script seems to be failing the black check

ethteck (approved):

PR #1262, opened 2026-03-05
Fix update_symbol_addrs.py compatibility

scarletcafe:
The old version of the script does a lot of manual parsing with hardcoded offsets into lines and so on, which no longer works due to both modifications made to the relevant files since the script's creation, and the fact these offsets can vary between versions.

This PR changes a few things:
- Allows you to specify an argument e.g. `python tools/update_symbol_addrs.py jp` to specify a version to work with instead of `current`.
- Reworks most of the manual parsing to be regex-based, so it is a bit more robust
- Uses dataclasses with named fields instead of various tuple and list based anonymous types (making it easier to read the code that uses them).
- Adds type hints to make the script pass with pyright in strict mode (makes it easier to examine the script in IDEs)

I have not committed the actual update `symbol_addrs.txt` files because the script, being automatic, removes all of the manually authored comments, but the updated script runs for all supported versions and I can add them in if desired.

### DIFF
--- a/tools/configure/requirements.txt
+++ b/tools/configure/requirements.txt
@@ -15,3 +15,4 @@ crunch64>=0.5.1
 spimdisasm==1.31.0
 splat64==0.31.0
 requests
+mapfile_parser>=2.12.1,<3.0.0
--- a/tools/update_symbol_addrs.py
+++ b/tools/update_symbol_addrs.py
@@ -2,6 +2,7 @@
 
 import argparse
 import dataclasses
+import logging
 import pathlib
 import re
 import subprocess
@@ -11,6 +12,8 @@ import typing
 import tqdm
 
 # Always the same
+LOGGER = logging.getLogger('update_symbol_addrs')
+
 SCRIPT_DIRECTORY = pathlib.Path(__file__).parent
 ROOT_DIRECTORY = SCRIPT_DIRECTORY.parent
 VERSION_DIRECTORY = ROOT_DIRECTORY / "ver"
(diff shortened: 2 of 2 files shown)


ethteck:
Thanks for working on this. my preference would be that we use [mapfile_parser](https://pypi.org/project/mapfile_parser/) instead, as it has pretty much already solved the problem being attempted in this script. I was meaning to update it at some point, but I never got around to it..would you be willing to modify the script to use it instead?

scarletcafe:
Sure thing. I'll take a look through it and see how I can rework the script.

scarletcafe:
I've changed the mapfile parsing to use `mapfile_parser`. From what I can see, its calculations for ROM and RAM addresses match the existing script, although `mapfile_parser` is much better at figuring out the file and section related to a symbol.

We don't output any of the additional info `mapfile_parser` gets us within `symbol_addrs.txt` itself, so I've put in an `--output-json` and `--output-csv` flag in as a bonus, and also added a `-v`/`--verbose` flag to enable the existing verbose logging without needing to modify the script itself.

BowserSlug:
🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀 This PR matches 71 functions (+0.77%) on `jp`.

